PR #43, opened 2020-05-21
Fix approval files line endings for Windows

haraldreingruber:
It's a weird issue, but aparently the LineEndings on Windows are not CRLF. This improvement might fix it.

Please test if the fix works on macOS as well?


haraldreingruber:
Hmm, seem to fail in Linux like enviornments... will have a look.

PR #115, opened 2021-11-22
Glass 2.0

srothst1:
The purpose of this pull request is to improve the appearance of glass in our target application. This will make our 3d streets more visually appealing, providing users with a more immersive experience. 

Here is an overview of my current progress: 
- Added the function `glassreflection` to the file `src/textures/textures.js`. This function updates the texture of glass objects in the scene based on a resource URL. When this function is applied to `a-mixin` object in `assets.js`, we see a noticeable difference in the texture of the glass.
  3dstreet_window_1 3dstreet_window_0
- Updated `index.html` to include `./src/textures/textures.js` as a script.
- Updated the version of `aframe-gltf-helpers` to `0.0.4`.
- Updated various assets in `assets.js` to apply the `glassreflection` script when being rendered in the scene.
A few questions, comments, and concerns that need to be addressed before this pull request can be merged:
- Is there a better way to ensure that the `a-mixin` objects in `assets.js` have access to the function `glassreflection` in `./src/textures/textures.js`? I am currently including `./src/textures/textures.js` as a script in `index.html`.
- When running this code, I noticed various errors in my console. The errors are as follows: `[gltf-part] Character_15 not found in model.`
- What assets in `assets.js` should be updated using `glassreflection`?
- What source image should we use for our texture? From @kfarr "For the image ideally we can use the equirectangular sky used in github.3d.st for both the sky and the reflection"

Overall, we are still largely in the prototyping stage. There are a lot of details that must be resolved before this pull request is ready to be merged.


srothst1:
Reference image from Babylon.js
image

srothst1:
https://glitch.com/edit/#!/fan-working-soap?path=index.html%3A16%3A39

Issue #116, opened 2021-12-04
Add cyclist and scooter 3D models to 3dstreet.


kfarr:
glb should be ready to go https://github.com/3DStreet/3dstreet-assets/blob/main/sets/micro-mobility-devices/micro-mobility-devices_v01.glb

segments to support:
scooter
bike-lane

possible or future segments that could be supported? (might need scooter and bicycle versions without people)
scooter-drop-zone
sidewalk-bike-rack

open questions:
- Place just 1 at center like vehicle? Or dispersed along length of segment like pedestrians? Or something else?

srothst1:
I will take a look at this shortly and open up a PR in [3dstreet](https://github.com/3DStreet/3dstreet).

PR #118, opened 2021-12-06
Micro Mobility Support 2.0 (Electric Scooters and Bicycles) 

srothst1:
This pull request addresses #116 

This pull request adds support for the `scooter` and `bike-lane` segments.

Regarding the open questions outlined in #116, I believe that we should just place one entity (scooter or bicycle) in the center of the lane. In streetmix, we are given the option to add more than one pedestrian. Streetmix does not support adding more than one scooter or bicycle in a lane.
  image

Here is a screenshot showcasing the new 3D models in 3dstreet:

image


srothst1:
Another fun screenshot :grinning: 

image

kfarr (approved):
Looks great to me!

PR #129, opened 2022-01-04
Modern Street Lamp Support

srothst1:
This pull request adds support for a more modern street lamp model.

image

A big thanks to @gnansai for the model. cc: https://github.com/3DStreet/3dstreet-assets/issues/14

@kfarr the pride flags look a little bit low to me, should I position them slightly higher?

I tested this pull request using the following street: https://streetmix.net/sam.rothstein/3/3dstreet-demo-street


kfarr:
Agreed on moving the pride flags up higher; my brief google street view research on castro street shows that the flags are poking up at least halfway above the tree line.

So perhaps moving it up one length -- the current top of the flag should be the new bottom of the flag position.

srothst1:
@kfarr I just updated the height of the pride flag. This looks much better to me. I could maybe go even a little bit higher. What do you think?

image

For context, here is the streetmix view

image

kfarr:
This looks great to me! I will merge this after a final check

kfarr:
Beautiful! 
image

PR #130, opened 2022-01-04
Migrating Models to Street Props Set

srothst1:
Updated the following models:
- bike_rack
- bike_share
- lamp_post_traditional
- palmtree
- park_bench
- sea_wall
- track
- tree-01

cc: https://github.com/3DStreet/3dstreet-assets/issues/15

@kfarr


kfarr:
Test street that includes all the objects listed:
https://streetmix.net/kfarr/69/3dstreet-pr130-test-street

2 issues:
- scale of bike rack
- rotation of sea wall

Here is current repo (github.3d.st)
image

Here is PR (localhost)
image

srothst1:
@kfarr I just added two commits that address both issues outlined above.

The only other slight issue that I just noticed relates to the `lamp-traditional` a-mixin. It looks correct if the scale is set to `scale="2 2 2"`

image

Otherwise, it looks comically small.

image

What are your thoughts? Should I leave the scale as `scale="2 2 2"`?

kfarr:
Agreed the new 2x scale you chose for the traditional lamps look much better!

The whole thing looks great! Had a little bit of trouble merging together but it seems to be working
image

PR #131, opened 2022-01-10
Magic Carpet

srothst1:
Adds support for a magic carpet in 3DStreet :magic_wand: 

Here was my initial implementation:
image

As you can tell, the carpet looked a little bit too low, so I decided to adjust the height slightly:

image

Thanks to @gnansai for the 3D Model.

@kfarr let me know if there is anything else I should tweak before this can be merged.

cc https://github.com/3DStreet/3dstreet-assets/issues/19


srothst1:
Going to fix lint issues next.

srothst1:
Lint issues have been resolved :+1:

PR #135, opened 2022-01-19
Suburbs 2.0

srothst1:
This pull request addresses the following issue by updating the `glTF` models used for the suburban buildings.

https://github.com/3DStreet/3dstreet-assets/issues/18

Let me know if there is anything else that needs to be done. 

cc @kfarr


srothst1:
From @kfarr:
I think you may have replaced the suburban houses with the same glb file
This is the one to update https://github.com/gnansai/3dstreet-assets/blob/main/sets/suburban-houses/gltf-exports/suburban-houses.glb
actually this link: https://github.com/3dstreet/3dstreet-assets/blob/main/sets/suburban-houses/gltf-exports/suburban-houses.glb

srothst1:
@kfarr we should be good to go 👍 

image

image

kfarr (commented):
looks good! clever to keep mixin id the same and then not have to update the js

kfarr:
looks good!

PR #170, opened 2022-05-18
driveways for suburban house models

srothst1:
The goal of this pull request is to add driveways to the suburban house models.

From my testing, these two housing models are best suited for driveways.

Screen Shot 2022-05-18 at 1 22 55 PM

However, I am willing to experiment with other driveway types.

### DIFF
--- a/src/tested/aframe-streetmix-parsers-tested.js
+++ b/src/tested/aframe-streetmix-parsers-tested.js
@@ -41,6 +41,28 @@ function createBuildingsArray (maxLength = 150, buildingType = 'narrow') { // es
       mixin: building.id,
       position: '' + (length + building.width / 2) + ' 0 0'
     };
+
+    if (buildingEntity.mixin === 'SM_Bld_House_Preset_08_1809') {
+      buildingEntity.child = {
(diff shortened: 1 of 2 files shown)


srothst1:
Here are some other screenshots: 

Screen Shot 2022-05-18 at 1 27 14 PM
Screen Shot 2022-05-18 at 1 27 46 PM

kfarr:
this is great, a few ideas
- might be nice to have class name that says "driveway" (i can make this change and test locally real quick)
- i could imagine similar treatment but perhaps with brick texture for walkway to door (future)

srothst1:
```
- i could imagine similar treatment but perhaps with brick texture for walkway to door (future)
```
I really like this idea! In general, there is a lot that we can do to make the suburban buildings more visually appealing and inspiring (gardens, community book cubbies, etc.).

```
- might be nice to have class name that says "driveway" (i can make this change and test locally real quick)
```
I am not sure exactly what you mean here but I am looking forward to seeing what you put together. Are you talking about some sort of variant string?

kfarr (approved):

PR #181, opened 2022-06-13
Sidewalk Animation 1.0

srothst1:
The goal of this pull request is to add character animations to the sidewalks in 3DStreet.

Character animations can be added to the `sidewalk` segment by adding the variant string `animated`. For instance, we have:
```
  {
    "width": 9,
    "variantString": "normal|animated",
    "type": "sidewalk"
  }
```

Here are some relevant resources that I have used:
- https://aframe.io/docs/1.3.0/components/gltf-model.html
- https://github.com/n5ro/aframe-extras/blob/master/src/loaders/animation-mixer.js
- https://aframe.io/docs/1.3.0/components/animation.html#api_to
- https://glitch.com/edit/#!/deep-cliff-bay

PR #182, opened 2022-06-17
Vehicle Animation 1.0

srothst1:
The goal of this pull request is to add vehicle animations to the street segments in 3DStreet.

Character animations can be added to the `drive-lane` segment by adding the variant string `animated`. For instance, we have:
```
  {
    "width": 9,
    "variantString": "outbound|car|animated",
    "type": "drive-lane"
  }
```

Here are some relevant resources that I have used:
- https://github.com/3DStreet/3dstreet/pull/181
- https://docs.google.com/spreadsheets/d/1VGRxCEPN6w99BxHQoIeWVnZFDhAL4FB_JfhMOvQihxI/edit?usp=sharing


srothst1:
Screen Shot 2022-06-17 at 3 16 10 PM

Here is a video showcasing the vehicle animations: https://youtu.be/kUgoRIKC6nA

srothst1:
Hi @kfarr 👋 Please take a look at this when you get the chance.

src/aframe-streetmix-parsers.js
```diff
@@ -295,7 +295,8 @@ function createBusElement (isOutbound, positionX) {
 }
 
 function createDriveLaneElement (variantList, positionX, segmentWidthInMeters, length, animated = 'false') {
-  var totalStreetDuration = (length / 5) * 1000;
+  var speed = 10; // meters per second
```
kfarr, line 298: this increases the default speed to 10 m/s ~= 22mph

kfarr:
Also for future reference here is reference to calculate rotations of wheel relative given a velocity and wheel radius https://www.texasgateway.org/resource/61-angle-rotation-and-angular-velocity#64411

src/aframe-streetmix-parsers.js
```diff
@@ -294,7 +294,17 @@ function createBusElement (isOutbound, positionX) {
   return busParentEl;
 }
 
-function createDriveLaneElement (variantList, positionX, segmentWidthInMeters, length) {
+function createDriveLaneElement (variantList, positionX, segmentWidthInMeters, length, animated = 'false') {
```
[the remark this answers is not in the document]
srothst1: No good reason - not using a boolean actually makes the code less clean. Just added a commit that changes `animated` to a boolean value in `createSidewalkClonedVariants` and `createDriveLaneElement` 👍

src/aframe-streetmix-parsers.js
srothst1: very speedy 🐎

srothst1:
Here is another resource that might be helpful: https://www.omnicalculator.com/everyday-life/rpm

A 2014 Toyota Camry has tires with a [diameter of 32 inches](https://www.goodyear.com/en-US/tires/toyota/camry/2014). Thus, if we are traveling at 22mph, our rpm should be roughly 230 according to omnicalculator.

Screen Shot 2022-06-24 at 10 15 25 AM

This seems a little bit high to me, but I would be curious to see how the animation looks.

srothst1:
There should be one full wheel rotation every `02.60869` seconds. This rate is that of a car that is navigating at 1 m/s.

Issue #196, opened 2022-08-25
Option to enable animations by default


srothst1:
For loud bicycle game we need ability to enable all animations by default

Here is one implementation option:
flag on street component like enableAnimations similar to [showGround](https://github.com/3DStreet/3dstreet/blob/main/src/index.js#L15) or showStriping with default = true. When set to true all object that include an animation are animated

kfarr:
closed by https://github.com/3DStreet/3dstreet/pull/201

Issue #191, opened 2022-08-10
Option to disable rendering of vehicles


kfarr:
For loud bicycle game we need ability to render a street without placing vehicles in the z center of a street segment

Options:
1) flag on street component like `showVehicles` similar to [`showGround`](https://github.com/3DStreet/3dstreet/blob/main/src/index.js#L15) or `showStriping` with default = `true`. When set to false the vehicles are not placed in the scene.
2) streetmix json "variant" at the segment level that can control vehicle presence on a segment by segment basis

I suggest 1 as a starting point

kfarr:
closed by https://github.com/3DStreet/3dstreet/pull/201

PR #201, opened 2022-08-25
option to disable vehicle rendering, options to enable global animations

srothst1:
resolves:
- https://github.com/3DStreet/3dstreet/issues/196
- https://github.com/3DStreet/3dstreet/issues/191


src/aframe-streetmix-parsers.js
```diff
@@ -841,7 +859,7 @@ function processSegments (segments, showStriping, length) {
       let reusableObjectStencilsParentEl;
 
       groundMixinId = 'bright-lane';
-      segmentParentEl.append(createDriveLaneElement([variantList[0], 'car'], positionX, segmentWidthInMeters, length));
+      segmentParentEl.append(createDriveLaneElement([variantList[0], 'car'], positionX, segmentWidthInMeters, length, false, showVehicles));
```
kfarr, line 862: looks like the `false` should be removed here? was probably used in testing? should be this instead:
      segmentParentEl.append(createDriveLaneElement([variantList[0], 'car'], positionX, segmentWidthInMeters, length, showVehicles));

kfarr (changes requested):

PR #202, opened 2022-09-02
VR button toggle based on XR capabilities 

srothst1:
VR button toggle based on XR capabilities 
Screen Shot 2022-09-02 at 3 30 41 PM

### DIFF
--- a/index.html
+++ b/index.html
@@ -37,6 +37,16 @@
     </div>
 
     <script>
+      if (typeof navigator.xr === 'undefined') {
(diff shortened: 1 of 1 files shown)


srothst1:
@kfarr to test on ~real~ VR device.

kfarr:
In order to try in the headset without having to use a cable and localhost server, I adapted this code to a glitch scene with the "stock" a-frame vr button and it worked!

https://glitch.com/edit/#!/lapis-wide-cartoon?path=index.html%3A23%3A12

PR #208, opened 2022-10-11
fix building position exceeds length issue

Algorush:


kfarr:
@Algorush before I review can you please:
* update test
* check linting (npm run lint:fix) and fix manual linting issues if needed

PR #210, opened 2022-10-16
Wheel animation update

kfarr:


src/tested/vehicle-wheel-animation-test.js
```diff
@@ -28,36 +28,39 @@ AFRAME.registerComponent('wheel', {
...
-    const t = (distx / dist) * (2 * Math.PI);
+    // let distance = speed * t ;
+    // let angle = (distance/(Math.PI*wheelDiameter))*2*Math.PI;
+    // let rateOfRotation = angle/t;
+
+    let rateOfRotation = (2*(speed/wheelDiameter));
 
     if (this.main_bone) {
-      this.main_bone.position.z += distx;
+      this.main_bone.translateY(speed);
```
kfarr, line 43: moving the vehicle via bone translation causes rendering issues.
instead, only wheels should be animated with bones and use a-frame / three.js to move the vehicle using animation component

vehicle-wheel-animation.html
```diff
@@ -1,7 +1,8 @@
 <html>
 
 <head>
-  <script src="https://aframe.io/releases/1.3.0/aframe.min.js"></script>
+  <script src="https://aframe.io/releases/1.2.0/aframe.min.js"></script>
```
kfarr, line 4: why downgrade a-frame version?

src/tested/vehicle-wheel-animation-test.js
kfarr: the test can use this componet:
https://github.com/networked-aframe/networked-aframe/commits/master/examples/js/forward.component.js

kfarr: added

vehicle-wheel-animation.html
kfarr: fixed

PR #213, opened 2022-10-19
add createEntities function

Algorush:


src/json-utils.js
```diff
@@ -50,3 +50,52 @@ function getAttributes(entity) {
...
+  // load default attributes
+  for (let attr in entityData.components) {
+    entity.setAttribute(attr, entityData.components[attr]);
+  }
+
+  if (entityData.id) {
+    entity.setAttribute("id", entityData.id);
+  }
+  if (entityData.class) {
+    entity.setAttribute("className", entityData.class);
```
vincentfretin, line 86: This is not React here, it's vanilla js. You need to use `class` not `className`.
`entity.setAttribute("class", entityData.class);`

src/json-utils.js
```diff
@@ -50,3 +50,52 @@ function getAttributes(entity) {
   }
   return elemObj;
 }
+
+function createEntities(EntitiesJSON) {
+  const entitiesData = EntitiesJSON.data;
```
vincentfretin, line 55: You should call the parameter in the function `entitiesData`
and use the `createEntities` function with createEntities(EntitiesJSON.data)
so the `createEntities` is not aware of the `EntitiesJSON` structure with the data key in the json.
Later we could have other keys in the json, like `EntitiesJSON.permissions`, but this is backend implementation specific, no need for this createEntities function to be aware of it.

src/json-utils.js
```diff
@@ -50,3 +50,52 @@ function getAttributes(entity) {
...
+    entityElements.appendChild(createEntity(entityData))
+  }
+
+  return entityElements;
+}
+
+/*
+Add a new entity with a list of components and children (if exists)
+ * @param {object} entityData Entity definition to add:
+ *   {element: 'a-entity', components: {geometry: 'primitive:box'}}
```
vincentfretin, line 70: You can add children, id, class in this documentation.

src/json-utils.js
```diff
@@ -50,3 +50,52 @@ function getAttributes(entity) {
...
+
+  if (parentEl) {
+    parentEl.appendChild(entity);
+  }
+
+  if (entityData.children) {
+    let childrenEntities = entityData.children;
+    childrenEntities.forEach(childEntityData => {
+      createEntity(childEntityData, entity);
+    });
```
vincentfretin, line 97: for consistency, use a for-of loop

src/json-utils.js
vincentfretin: see my comment https://github.com/3DStreet/3dstreet/issues/212#issuecomment-1283639723

src/json-utils.js
vincentfretin: same issue in the saving part https://github.com/3DStreet/3dstreet/blob/fd568a7cd78d89c6087a365f55069d5303cea113/src/json-utils.js#L39-L41

vincentfretin: My bad, we indeed have `entity.className` in vanilla js, so that's ok for the saving part. But `entity.setAttribute('className', 'something')` is definitely not setting the class, `entity.className` or `entity.getAttribute('class')` don't change.

vincentfretin:
Also you will need to emit the `entitycreated` event for each created entity if you want the left panel to update. See the original `createEntity` function in `lib/entity.js`

PR #219, opened 2022-11-13
Basic UI for loading streetmix or json

kfarr:


kfarr:
Updates:
* new behavior: no street loaded by default when opening index.html
* instead use the buttons on the right-hand side to load files. For streetmix it opens an input dialog to paste in a streetmix URL, no work needed on this.
* For 3dstreet json file it opens an input dialog to paste in a string for testing.

Still to be done:
- [ ] ability to load from copy/paste string the basic test case 1 json (a-frame hello world shapes scene) 
- [ ] ability to load from a .json file
- [ ] create welcome dialog with 4 options: Play Edit Load Help

PR #222, opened 2022-11-23
Unify with editor index page

kfarr:
possibly breaking things:
- rename style url
- change to click to teleport instead of orbit wide perspective


kfarr:
still to finish:
- update viewer logo image to latest (both)

PR #224, opened 2022-12-03
add asset categories functionality

Algorush:
- add functionality to allow specifying of which categories of assets are loaded in buildAssetsHTML


kfarr:
was able to verify reduction from 27MB to 21MB from loading all assets to loading just the minimum required for basic streetmix imported street scene without intersection

PR #243, opened 2023-02-02
change fences glb path to the draco version

Algorush:


kfarr:
tested and looks good

PR #245, opened 2023-02-11
add random placement for static entities

Algorush:


kfarr:
this looks great, thank you @Algorush !

PR #246, opened 2023-03-01
create 3d curb at intersections

Algorush:
- add box geometry for sidewalks and curbs
- fix repeat property for sidewalk and curb textures

### DIFF
--- a/intersection.html
+++ b/intersection.html
@@ -83,9 +83,7 @@
         street="length: 55"
         streetmix-loader="streetmixStreetURL: https://streetmix.net/srothst1/7/lapara; showBuildings: false;"
       ></a-entity>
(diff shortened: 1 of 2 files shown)


kfarr:
hey @Algorush thanks for this pr. all looks good except one issue, for intersection with only 2 sides, there is a 0 width curb that shouldn't exist. 

to reproduce, go to http://localhost:7001/intersection.html

see screenshot items highlighted in red outline should not be there
image
and
image

Algorush:
Okay thanks, I didn't notice that issue. I also want to make sure that the sidewalk boxes don't overlap each other to fix the flickering of the textures (when viewed from outside the scene)

kfarr:
looks great!

PR #249, opened 2023-03-15
migrate anisotropy component

Algorush:
migrate anisotropy component into the 3dstreet repo


kfarr:
@Algorush i see that you have copied the component into the directory, but you did not update the imports in index.js:
https://github.com/3DStreet/3dstreet/blob/main/src/index.js#LL7

PR #255, opened 2023-03-29
add more randomly placed vehicles in parking lane

Algorush:
added the ability to specify the number of machines through the function parameter, as well as the placement step


kfarr:
test street: http://localhost:7001/#https://streetmix.net/kfarr/15

PR #266, opened 2023-04-12
updating 3dstreet to support a-frame 1.4

kfarr:


kfarr:
tested in VR mode on quest 2 and works as expected especially loading on network

PR #268, opened 2023-04-12
add brt-station segment support

Algorush:
Added for review scale of Brt-station. If it looks good, I'll scale brt-station glb with same size. I described the situation in this thread:
https://github.com/3DStreet/3dstreet-assets-source/issues/20#issuecomment-1506031741


kfarr:
Here's an example street for review: https://streetmix.net/kfarr/97/brt-example

Algorush:
So, I changed scale of object 'brt-station' in brt-station.glb to 0.75. Now I can replace it in 3dstreet-assets-dist repo, right? And also remove scaling to 0.75 value throught setAttribute in the code. It will be better for perfomance I think

kfarr:
@Algorush I agree it is best for the `3dstreet-assets-source` repo as the "source of truth" to have the updated scale and the app to import it without scale modification. Yes it would help to modify the original blender source, the dist glb and app logic.

PR #273, opened 2023-04-28
add perpendicular parking option

Algorush:
- refactor createDriveLaneElement function
- add perpendicular parking option


Algorush:
Looks like I need to update some tests

kfarr:
test street
http://localhost:7001/#https://streetmix.net/tomasborja21/1/university-blvd

kfarr:
it looks mostly ok

except it should use white stripes betweens paces instead of the white + sign

i think you can use this mixin:
https://github.com/3DStreet/3dstreet/blob/main/src/assets.js#LL156

example image:
image

and yes to fix the failing tests. it does look like you moved some things around.

Algorush:
added a temporary solution for correct texture rendering when using atlas-uvs

kfarr:
@Algorush I think I merged it successfully...

PR #276, opened 2023-05-03
Sidewalk segment variants

Algorush:
added all sidewalk variants and repaired animation. Also maked some optimization and refactor

### DIFF
--- a/src/aframe-streetmix-parsers.js
+++ b/src/aframe-streetmix-parsers.js
@@ -181,7 +181,7 @@ function getZPositions (start, end, step) {
   return arr.sort(() => 0.5 - Math.random());
 }
 
-function addLinearStreetAnimation(reusableObjectEl, speed, streetLength, xPos, yVal = 0, zPos, direction) {
+function addLinearStreetAnimation (reusableObjectEl, speed, streetLength, xPos, yVal = 0, zPos, direction) {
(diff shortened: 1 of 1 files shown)


kfarr:
tested against:
http://localhost:7001/#https://streetmix.net/kfarr/98/russelsheimer-str-q1

issues:
* animated doesn't seem to work (empty) -- change line 68 of index.html to:
`        <a-entity id="default-street" street="globalAnimated: true" streetmix-loader set-loader-from-hash></a-entity>`
* change from THREE object3D.set to AFRAME setAttribute

@Algorush

Algorush:
> tested against: http://localhost:7001/#https://streetmix.net/kfarr/98/russelsheimer-str-q1
> 
> issues:
> 
> * animated doesn't seem to work (empty) -- change line 68 of index.html to:
>   `        <a-entity id="default-street" street="globalAnimated: true" streetmix-loader set-loader-from-hash></a-entity>`
> * change from THREE object3D.set to AFRAME setAttribute
> 
> @Algorush

Sorry, I remembered the animation doesn't work because the assets models of the animated people are not loaded. That is, to make it work, add 'people-rigged' to the categories component for the street-assets element in index.html

kfarr:
@Algorush you're totally right. I just tested with the proper models loaded and it works great. I'll merge. I have a separate work in progress to simplify asset loading (i hope)

PR #283, opened 2023-05-24
fix NaN issue with pedestrians

Algorush:
I noticed that the pedestrians do not appear after the last merge due to the fact that NaN values were passed to their position. I made a typo in the penultimate commit when I changed object3D.set to setAttribute. I fixed a typo in this commit.


kfarr (approved):

kfarr:
@Algorush thanks for catching

PR #285, opened 2023-05-26
fix empty building variant showing

Algorush:


kfarr:
thanks for the quick fix @Algorush !

PR #287, opened 2023-06-03
Save load fixing

Algorush:
it remains to add environment and 2D-Layers nodes in JSON


kfarr:
@Algorush what is the best way to test this?

Should I copy this over to the editor repo to test saving? Or do you fire the save function via console JS on this repo to test?

Algorush:
> @Algorush what is the best way to test this?
> 
> Should I copy this over to the editor repo to test saving? Or do you fire the save function via console JS on this repo to test?
the editor uses json-utils.js, the path to which is specified in index.html. Therefore, I tested through the save / load buttons of the editor

kfarr:
Here is my test case:
* load a street from streetmix http://localhost:3333/#https://streetmix.net/kfarr/3/example-street where localhost:3333 is the editor repo with index.html pointing to updated json-utils.js
* select vehicle models and changing the mixins, duplicating some models, and changing position
* save as 3dstreet json
* load new blank scene http://localhost:3333/
* load from 3dstreet json and choose the same file
* select one of the modified vehicles, see that the mixin (model) displays correctly and is still user modifiable
* repeat similar modifications, save, and reload and confirm that the mixin (model) displays correctly are are still user modifiable

this appears to work correctly

are there more advanced test cases you recommend @Algorush ?

Algorush:
it might be worth checking save/load with different components. Because in order not to save the values ​​from the mixin, I have now manually specified the components that need to be skipped. This is probably not the best approach, but has not yet come up with a universal solution.

kfarr:
Agreed other components will need to be checked but for now this is an ok initial release. We can include in documentation the fact that at the moment the save / load only works for mixins, not custom components.

Here are times when this might be an issue, which we can resolve later in separate tickets but just brainstorming now:
* if a user changes a material / texture of a roadway or sidewalk surface -- this is not an officially supported feature at the moment anyway
* if a user adds a custom 2d layer that has src attribute to an image defined -- this would not use a mixin and might "just work" with this code but should be tested when we are ready to support this feature
* if a user changes the skybox src to a custom image -- same -- this would not be a mixin and might "just work"

kfarr:
I suppose we should test the behavior of loading an old saved file before this version, I will try that next

Algorush:
I haven't added the environment and 2d-layers nodes yet. There were some difficulties here, but I almost did

kfarr:
Ok can you do a new pull request for that? Not urgent

PR #300, opened 2023-06-15
add environment node support

Algorush:
I renamed branch but github added changes to the old one. I hope it's okay.
To support loading the envirinment and other nodes, I need to make changes to 3dstreet-editor. Here is the link to the PR with the changes: https://github.com/3DStreet/3dstreet-editor/pull/195


kfarr:
Note to self the testing process is similar to: https://github.com/3DStreet/3dstreet/pull/287#issuecomment-1577819495

kfarr:
Hey @Algorush I tested this and it doesn't quite work yet, but I think I made some mistake in not explaining the use case very well. I will try to explain in more detail, here are the desired steps:
* Open up a 3dstreet scene, such as http://localhost:3333/#https://streetmix.net/kfarr/3/example-street (using the testing method linked above)
* Selected environment layer (entity), and then scroll down to the `street-environment` component
* Change the `preset` value from `day` to `night`
* Save as JSON
* Inspect JSON file on local machine in text editor, expect to see environment entity with `street-environment` component `preset` attribute set to `night`
* Open JSON from 3dstreet and see a night scene
* Instead, what happens with the current version of this PR is that the scene gets brighter -- I think that is because it is adding the lights from the JSON file to the existing lights added by default in the scene

I think there are a few issues here:
* the `street-environment` component is not listed as a component value to be saved in the json save utils. That should be added, that should be part of this ticket.
* the `street-environment` component does not support dynamically changing from `night` into `day`. It should. This could be a separate ticket. (It is a common problem with a-frame components that they work when initialized once in a scene but need extra work to allow dynamic updating)
* we should not save any of the children of the `environment` id entity (the Environment layer). They should instead be controlled by the `street-environment` component of the `environment` id entity. That will probably also need your help to update in json save utils as part of this ticket.

Algorush:
Thanks for the detailed explanation of how saving the environment should work. I will change it.
In this PR I've made the environment save, taking into account the case if the user will change the light entities directly

PR #302, opened 2023-06-19
return parking-t mixin

Algorush:
return parking-t mixin for perpendicular parking option


kfarr:
Hi @Algorush this is excellent, thank you. It looks like this also resolves https://github.com/3DStreet/3dstreet-editor/issues/196 ?

kfarr:
tested URL: https://deploy-preview-302--3dstreet-core-builds.netlify.app/#https://streetmix.net/kfarr/10

screenshot
image

PR #304, opened 2023-06-20
Add first version to load JSON from URL

Algorush:
Now it works with files from localhost, but didnt tested with remote files yet


kfarr:
Hi @Algorush I'm curious how you went about testing this? Here is what I did:

Testing method for local:
http://localhost:7001/#/examples/json/crash-scene.json

Testing method for remote:
https://deploy-preview-304--3dstreet-core-builds.netlify.app/#/examples/json/crash-scene.json

Issue with both local and remote:
- look at dev console and see scrolling messages of street loader initializing 
- infinite loop loading a street scene pegs CPU and eventually browser runs out of memory resulting in crash
- It appears `set-loader-from-hash` is running init repeatedly. Might be related to using async with a-frame lifecycle method, i have not seen this before.
- Instead, suggest to try move away from async and instead go back to an old fashioned synchronous (blocking) call similar to the call to fetch streetmix street json ([here](https://github.com/3DStreet/3dstreet/blob/main/src/index.js#L78)). Yes this might be bad because it might freeze UI as compared to async, but this is loading the scene once and the user has to wait during this time anyway.

Architecture brainstorm for future:
- Create a save-load-json component intended to be added to the scene entity.
- In viewer mode detect url hash and load the scene from json via this component
- In editor mode use this component for save and load functions

Algorush:
> 

Hi @kfarr . Sorry, I forgot to say about this issue with repeatedly `set-loader-from-hash` init. This happens when the `set-loader-from-hash` component is in json. It calls the add scene function from JSON with the `set-loader-from-hash` component on the a-entity#default-street element. And when the component is initialized, the process is repeated and looped. It's not because of async/await.
In this PR I modified the json-utils.js code to save JSON without the `set-loader-from-hash` component. And if using the new version of JSON, then there is no looping. Probably it was necessary to change the version of json-utils. Also yesterday I was working on options to not initialize this component twice, but have not finished and uploaded this code yet

Algorush:
Maybe also such a test case?
http://localhost:7001/#https://deploy-preview-304--3dstreet-core-builds.netlify.app/examples/json/crash-scene.json

kfarr:
@Algorush thanks for the updates, it sounds like you're aware of these issues and I'll wait to test again until you do some more commits.

In the test case you provided yes it makes sense in theory, but the local page is loaded with `http` protocol so I don't think javascript can access `https` URLs and I think netlify forces https which means the json probably won't load.

Algorush:
I've added checks in two places to see if the component `set-loader-from-hash` is inside the JSON. In one place after downloading the file via XHR. And in json-utils I also added a check in case the component is loaded from the editor mode with the 'load JSON' button

Algorush:
> @Algorush thanks for the updates, it sounds like you're aware of these issues and I'll wait to test again until you do some more commits.
> 
> In the test case you provided yes it makes sense in theory, but the local page is loaded with `http` protocol so I don't think javascript can access `https` URLs and I think netlify forces https which means the json probably won't load.

that is, do not consider the case of downloading a file from a third-party source yet?

kfarr:
Thanks I've started testing:
- [*] test opening / changing / saving / re-opening local files
- [*] confirming that streetmix urls still work as expected
- [ ] test remote files

for later:
- detect hashchange event

PR #307, opened 2023-07-05
add update environment option

Algorush:


kfarr:
this works great, i'll merge this first and then test saving / loading of it

PR #319, opened 2023-07-22
Notification component

Algorush:
I made css a separate file and imported inside viewer-style.css. I think it's more convenient or no?


Algorush:
It looks like I created this branch not from the master by mistake... I hope everything will be fine

kfarr:
this is a great start, I tested using the following js console command:
`AFRAME.scenes[0].setAttribute('notify','successMsg','This is super awesome, thanks Alex.')`

some requests for updates, hopefully this isn't overthinking but makes it simpler for users:
- i do like the idea of importing the css for this, thanks!
- a few commented out lines index.html could be deleted
- suggest to add 1 additional custom type 'info', color blue with no icon
- therefore there are now 3 total types: 'error', 'success', 'info'; default should be 'info'
- But, instead of 3 methods for sending a message, have 1 `message` method, for example it would be called like this:
`AFRAME.scenes[0].components['notify'].message('yoyoyo')`
- and then there is optional second variable in message method that is type ('error', 'success', 'info') so this would be an error message: `AFRAME.scenes[0].components['notify'].message('something broke', 'error')`
- same feedback for the schema, there should be a new property 'type' that is default 'info' and can be one of 3 values 'error', 'success', 'info'; then there is only 1 property 'message' that triggers message sending; then remove the 'successMsg' and 'errorMsg' properties
- default x value should be center instead of right

Algorush:
It seems that I did everything as you suggested. It seems that all works good

PR #326, opened 2023-08-05
rename layers-2 entity

Algorush:
Also I'll do the same for 3DStreet-editor repo


kfarr (approved):
excellent -- was able to get a test scene from editor repo to save as json and then load successfully in this 3dstreet repo (sans traffic signals but that's a problem for another day)

PR #341, opened 2023-09-05
fix intersection saving issue

Algorush:
change intesection geometry primitive from plane to box and fix positions of intersection elements

### DIFF
--- a/index-intersection.html
+++ b/index-intersection.html
@@ -1,0 +1,227 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <!-- aframe -->
+    <script src="https://aframe.io/releases/1.4.2/aframe.min.js"></script>
+
+    <!-- 3dstreet -->
+    <script src="https://github.3dstreet.org/dist/aframe-street-component.js"></script>
+
+    <!-- user controls - vr teleport -->
+
+    <!-- mapbox -->
+    <script src="https://github.3dstreet.org/src/lib/aframe-mapbox-component.min.js"></script>
+
+    <!-- save / load -->
+    <script src="https://github.3dstreet.org/src/json-utils.js"></script>
+    
+    <!-- ocean -->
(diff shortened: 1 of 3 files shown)


kfarr:
@Algorush thanks, do you mind describing a bit more the logic behind changing positions as part of the solution?
for example https://github.com/3DStreet/3dstreet/pull/341/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R173 and https://github.com/3DStreet/3dstreet/pull/341/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R147

kfarr:
@Algorush the original issue appears resolved, but now there is a new regression: each of the 4 curbs (each corner of the intersection) is now higher than before.

You can see the difference in old and new screenshots below.

I also notice a lot of changes to z values in the diff of index.js:
https://github.com/3DStreet/3dstreet/pull/341/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R222

Why all the z value changes? Would reverting some of the z value changes fix the regression?

old:
image

new:
image

Algorush:
To avoid problems with color and to make the intersection element the same as the rest with asphalt, I changed the primitive from plane to box for the element to which the intersection component is applied. After that, all child elements of the intersection slightly changed their z-axis relative to the scene. Therefore, I had to correct each type of element separately. But strange, I have not seen such offsets as in the screenshots. I will check again

Algorush:
I'm still unable to reproduce the example of the z-position mismatch between curbs and footpath. Everything looks good on my local scene. Everything is good on this scene too: https://deploy-preview-341--3dstreet-core-builds.netlify.app/intersection.html

kfarr:
@Algorush yes you're right, I was able to reproduce and it was my fault because of not using the latest json-utils that a second intersection scene was being added upon reload from json that was slightly offset, but the original is correctly placed. Anyway it's good to go

PR #352, opened 2023-09-15
specify a URL to load scene from json file (v1)

Algorush:


kfarr:
test case:
- this glitch project: https://youthful-dear-iridium.glitch.me/

it returns a cors console error:
> - Access to XMLHttpRequest at 'https://dev-3dstreet.web.app/scenes/b020190e-29a0-453e-8ee1-bcf1ca2a6d69.json' from origin 'https://youthful-dear-iridium.glitch.me' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

working on cors issues here: 
https://github.com/3DStreet/3dstreet-editor/pull/292

Algorush:
I made these 3 points in the new commit. Now json-utils is called in index.js. But now inputStreetmix and inputJSON which are used in index.html for UI elements cannot be used. I'm thiking how to best fix it? Add this functions to one of compoents or add them to index.html? Or describe index.js as a module and export functions from module

kfarr:
> But now inputStreetmix and inputJSON which are used in index.html for UI elements cannot be used. I'm thiking how to best fix it? Add this functions to one of compoents or add them to index.html?

@Algorush yes I think it's a good idea to put the inputStreetmix and inputJSON into their appropriate component counterparts as functions to be called from front-end ui.

> Maybe I misunderstood you and I need to add all the relevant functions from json-utils inside the json-3dstreet component?

yes perhaps it's inputStreetmix function into streetmix-loader component and inputJSON function into json-3dstreet component for example.

Algorush:
Will index.html in the future have #street-container and #default-street elements or create them dynamically inside a-scene at the first level?

Algorush:
And I also ask because if we place the inputStreetmix function (with the prompt request) inside the component, then it will need to be called when assigning the streetmix-loader attribute, which is called for the #default-street element

kfarr:
@Algorush I think it's ok to require that certain elements exist on the page for v1

The primary use case is to allow people to bring their 3dstreet scenes to a custom html a-frame scene often hosted on glitch.com . We can create the basic glitch template that includes these elements as a starting point that people can "remix" from

Algorush:
I moved the inputStreetmix function inside the streetmix-loader component. But still part of the code remains in index.html. This is the eventListener code for a button with a custom input prompt

Algorush:
I moved the addEventListener for the change event for the inputSteetmix URL inside the `json-3dstreet` system. I don't know if this is a good idea, but it works and there is less code in index.html.
The same can be done for streetmix-loader, but is there a suggestion here to abandon streetmix-loader https://github.com/3DStreet/3dstreet/issues/398? Or did I misunderstand?

kfarr:
@Algorush  I'm reviewing this and struggling to understand all of the code. I wonder if there are too many changes going on at the same time and we should go 1 step at a time.

Here are the core user goals:
1) As a 3DStreet developer I want json-utils to exist only on 1 repo and be included in index.js
2) As an A-Frame developer I want to be able to specify a 3DStreet scene ID and/or scene URL in an A-Frame component so that I can include a 3DStreet scene in my A-Frame project.

Maybe we start with a very simple PR that does #1 only and then a separate PR that does #2 only?

Algorush:
> Maybe we start with a very simple PR that does #1 only and then a separate PR that does #2 only?

Yes, I agree that this will be better.

Algorush:
Ok, I will gradually transfer the changes from this PR to the new PRs for more convenient. This will take some time.

kfarr:
Closing this for now, we are splitting this into separate tasks

PR #355, opened 2023-09-22
dont save street propery in JSON

Algorush:


kfarr:
Is this removeProps list used in saving new files? if we ignore all of street component properties, that means all those props cannot be changed by the user such as turning on/off showGround or showStriping or adjusting length? At the moment those can't be edited in realtime anyway, they have to be saved and then the scene reloaded for the scene to adjust and display the changes correctly. (or the street entity removed and a new one loaded in its place)

Algorush:
Maybe I can add new boolean key 'loaded' to street property while saving JSON. And inside street component I can check it, if its equals true, then dont do nothing. And it  will also help not reload all scene in other cases

Algorush:
I added commits, where I adjusted the work of the function that filters components. Now removing JSON string in street component works well

kfarr:
@Algorush I am confused by this -- why do we remove street parent?

>     const streetParent = this.el.querySelector('.street-parent');
>     const streetEl = streetmixParsers.processSegments(streetmixSegments.streetmixSegmentsFeet, data.showStriping, data.length, data.globalAnimated, data.showVehicles);
>     if (streetParent) {
>       streetParent.remove();
>     }

kfarr:
* @Algorush can you remove the unrelated parent entity changes in index.js from this repo? 
* Can you also explain how to test before / after for the other changes? (ie do x, y, z to generate extra JSON in current branch, do x, y, z with branch and observe something different?) So that I can verify?

kfarr:
Works as expected, and both old and new json files generated are able to be opened
image

PR #387, opened 2023-10-19
up cyclist y position for bike-track

Algorush:


src/aframe-streetmix-parsers.js
```diff
@@ -528,7 +528,7 @@ function createMicroMobilityElement (variantList, positionX, segmentType, length
   }
   const bikeLength = 2.03;
   const positionZ = randomPosition(reusableObjectEl, 'z', length, bikeLength);
-  reusableObjectEl.setAttribute('position', positionX + ' 0 ' + positionZ);
+  reusableObjectEl.setAttribute('position', positionX + ' 0.2 ' + positionZ);
```
kfarr, line 531: Confused a bit about this, hard coding the Y position at 0.2 means that the cyclist will float if not on the sidewalk or higher elevation surface?

kfarr:
@Algorush I think this creates a new issue -- by hard coding 0.2 y position for cyclist it results in "floating" cyclist
image

Maybe instead we should focus on this ticket: https://github.com/3DStreet/3dstreet/issues/388

Algorush:
> Maybe instead we should focus on this ticket: https://github.com/3DStreet/3dstreet/issues/388

Yes, you're right, we need to support elevation level dynamically. 

>  I think this creates a new issue -- by hard coding 0.2 y position for cyclist it results in "floating" cyclist

That's my fault. In another scene the cyclist was lower than the sidewalk, so I thought he was lower everywhere...

kfarr:
All good, I'm going to close this one for now and will try another solution

PR #391, opened 2023-10-26
change roughness for all surfaces

Algorush:
change roughness for all surfaces to 0.8


kfarr:
test method:
- change to night
- add a light
- compare visual outcome different values

old
image

new
image

looks great!

Issue #396, opened 2023-11-02
Change structure of entities created by street component: BuildingsLeft, Segment1, 2, ... , Segment n, BuildingsRight


kfarr:
Issue:
- Today, each segment "parent" is an entity with 0 0 0 location. Child entities have x and y translation that corresponds to the specific segment key/value pairs from streetmix json
- Instead, each segment "parent" should be an entity that reflects the x and y location of the specific segment, and children should be 0 0 0 (or otherwise adjusted using the parent's 'correct' location as its origin)

This would also solve this problem https://github.com/3DStreet/3dstreet/pull/387 and https://github.com/3DStreet/3dstreet/issues/388

Checklist:
- [x] In aframe-streetmix-parsers.js, modify segment-parents to have correct X placement
- [x] add updated titles per tree structure proposal below
- [ ] update parent relationships per tree structure proposal below
- [x] Reconcile Y placement if needed (wait for AG work)
- [x] rename / restructure buildings per below

## street parent tree structure proposal
Today it is:
```
3DStreet Layers
├── default-street (entity with street and street-loader components)
│   ├── street-parent
│   │   ├── segment-parent-0
│   │   ├── segment-parent-1
│   │   ├── segment-parent-2
│   │   ├── segment-parent-n
│   ├── buildings-parent
│   │   ├── ground-left
│   │   ├── block-left
│   │   ├── ground-right
│   │   ├── block-right
```

Instead I propose:
```
3DStreet Layers
├── Street: Harrison Street SB at 16th (use string from streetmix street name)
│   ├── Segment: Sidewalk (use segment type string)
│   ├── Segment: Bench (use segment type string)
│   ├── Segment: Tree (use segment type string)
│   ├── Segment: Parking (use segment type string)
│   ├── Segment: Drive Lane (use segment type string)
│   ├── Segment: n
│   ├── Block: ResidentialBuildings (use building type string)
│   ├── Block: ResidentialBuildings (use building type string)
```

Algorush:
Yes, that's a good idea. 
But to support different elevations, `position.y` must be set to the element's parent (eg .pedestrian-parent) depending on the elevation. 
And by the way, it is enough to change `position.y` for .pedesrtian-parent and not for each pedestrian element, as now

kfarr:
@Algorush great, that makes sense. I'll wait on Y for now and then you can review

kfarr:
[...]

PR #397, opened 2023-11-02
New street parser structure

kfarr:
https://github.com/3DStreet/3dstreet/issues/396

### DIFF
--- a/src/aframe-streetmix-parsers.js
+++ b/src/aframe-streetmix-parsers.js
@@ -228,7 +228,7 @@ function getZPositions (start, end, step) {
   return arr.sort(() => 0.5 - Math.random());
 }
 
-function createSidewalkClonedVariants (segmentWidthInMeters, density, streetLength, direction = 'random', animated = false) {
+function createSidewalkClonedVariants (segmentWidthInMeters, density, elevationPosY = 0, streetLength, direction = 'random', animated = false) {
   const xValueRange = [-(0.37 * segmentWidthInMeters), (0.37 * segmentWidthInMeters)];
   const zValueRange = getZPositions((-0.5 * streetLength), (0.5 * streetLength), 1.5);
   const densityFactors = {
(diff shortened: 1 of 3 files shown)


Algorush (approved):
I deleted the last positionX and added support of elevation level for all segments

kfarr:
Hi @Algorush I just tested and bike lane works great however divider doesn't

here is example, the flower box is at correct elevation but not the grass (ground), however the streetmix json has elevation of 1
image
image

src/aframe-streetmix-parsers.js
```diff
@@ -692,16 +694,15 @@ function createSegmentElement (scaleX, positionY, rotationY, mixinId, length, re
 
   const scalePlane = scaleX + ' ' + scaleY + ' 1';
   const scaleBox = scaleX + ' 1 1';
+  let heightLevels = [0.2, 0.4, 0.8]
```
kfarr, line 697: confused why there are 3 heigh levels, aren't there only 2 potential values for elevation 0 and 1? @Algorush

kfarr:
also @Algorush separator markings are not showing in this update. here is new (left) and old (right) side-by-side comparison
image

src/aframe-streetmix-parsers.js
```diff
@@ -693,30 +688,29 @@ function createCenteredStreetElement (segments) {
...
+function createSegmentElement (scaleX, positionY, rotationY, mixinId, length, repeatCount, elevation = 0) {
   var segmentEl = document.createElement('a-entity');
   const scaleY = length / 150;
 
   const scalePlane = scaleX + ' ' + scaleY + ' 1';
   const scaleBox = scaleX + ' 1 1';
+  let heightLevels = [0.2, 0.4, 0.8]
+  let height = heightLevels[elevation];
+  if (elevation === 0) {
+    positionY = -0.1;
```
kfarr, line 700: This discards some positionY that may have been set earlier in processSegments such as for separator segments that set positionY = 0.01 (to prevent z fighting from 2 nearby planes) @Algorush

Issue #388, opened 2023-10-19
fix elevation level support


Algorush:
elevation level support for pedestrians, cyclists and other types of segments where the elevation level can be changed. Now this does not always work correctly

Algorush:
Yes, I agree with everything. Streetmix supports changing the elevation level for sidewalks segments, bike paths and other similar segments

kfarr:
Ok I'm going to do a try at this for next day or two and will ping you when ready to check it out

Algorush:
I was going to do it myself. Because not long ago I did support elevation, but only the default value.

kfarr:
I'm starting with this https://github.com/3DStreet/3dstreet/issues/396 first focusing on X axis, you're welcome to also start on this with the Y axis issue and then we can merge it together?

Algorush:
it took a little longer. Almost finished. I checked elevation for all possible segment types.

Algorush:
Now added commit with elevation support for all segments

kfarr:
should be closed with https://github.com/3DStreet/3dstreet/commit/d467f149b20a9b3f0335f79d920828f08662e3c9

Issue #383, opened 2023-10-19
fix the elevation level of the cyclist


Algorush:
The elevation level of the cyclist does not correspond to the level of the bike path. I'll fix it

Algorush:
https://github.com/3DStreet/3dstreet/pull/387

kfarr:
should be closed with https://github.com/3DStreet/3dstreet/commit/d467f149b20a9b3f0335f79d920828f08662e3c9

PR #400, opened 2023-11-06
Fix elevation level support

kfarr:
Related to:
https://github.com/3DStreet/3dstreet/issues/388 and
https://github.com/3DStreet/3dstreet/issues/383 and
https://github.com/3DStreet/3dstreet/issues/396


kfarr:
see https://github.com/3DStreet/3dstreet/pull/395/commits/d467f149b20a9b3f0335f79d920828f08662e3c9 instead

PR #405, opened 2023-11-08
add environment option color

kfarr:


kfarr:
will the environment preset setting (including color) save and / be reloaded? i vote to test upon deploy and assume it won't proven otherwise

PR #406, opened 2023-11-11
Screenshot without helpers

Algorush:
I changed the code of the screentock component a little. I moved the tock event code to the update event, I think it's more efficient. Now this code is not called at every next toсk of the entire scene. But now its not a screentock component...
Before saving the screenshot, I run renderer.render() so that the scene is rendered first.
@kfarr Let me know if its ok or no.
We can also leave the screentock code as it was.
And rename the resulting component to screenshot??
And also for the screenshot, perhaps it’s worth hiding the round teleport control pointer?

### DIFF
--- a/src/components/screentock.js
+++ b/src/components/screentock.js
@@ -14,32 +14,6 @@
 //   AFRAME.scenes[0].setAttribute('screentock', 'takeScreenshot', true);
 // }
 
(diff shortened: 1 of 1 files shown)


kfarr:
These changes are included in https://github.com/3DStreet/3dstreet/pull/409 so let's use that for discussion and further edits, I will close this PR

PR #409, opened 2023-11-12
Title and logo in screenshot

Algorush:
I've done everything so far as part of the `screentock` component. Maybe it's better to rename component to `screenshot` now? Since there is no need to use the `tock` method here, as I believe. But maybe there's something I don't know?

### DIFF
--- a/src/components/screentock.js
+++ b/src/components/screentock.js
@@ -104,7 +104,7 @@ AFRAME.registerComponent('screentock', {
     const screenshotCanvas = await createCanvasWithScreenshot(renderer.domElement);
 
     if (type == 'img') {
-      imgElement.src = renderer.domElement.toDataURL();
+      imgElement.src = screenshotCanvas.toDataURL();
       return;
     }
     if (type == 'png') {
--- a/src/components/screentock.js
+++ b/src/components/screentock.js
@@ -21,7 +21,7 @@ AFRAME.registerComponent('screentock', {
     type: { type: 'string', default: 'jpg' }, // png, jpg, img
     imgElementSelector: { type: 'selector' }
   },
-  takeScreenshotNow: function (filename, type, imgElement) {
+  takeScreenshotNow: async function (filename, type, imgElement) {
     const inspector = AFRAME.INSPECTOR;
     const renderer = AFRAME.scenes[0].renderer;
 
@@ -32,7 +32,7 @@ AFRAME.registerComponent('screentock', {
       if (inspector && inspector.opened) inspector.sceneHelpers.visible = show;
     }
 
-    function createCanvasWithScreenshot(aframeCanvas) {
+    const createCanvasWithScreenshot = async (aframeCanvas) => {
       let screenshotCanvas = document.querySelector('#screenshotCanvas');
       if (!screenshotCanvas) {
         screenshotCanvas = document.createElement('canvas');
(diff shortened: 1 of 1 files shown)


kfarr:
Hi @Algorush great work so far! I'm combining the "hide helpers" [ticket](https://github.com/3DStreet/3dstreet-editor/issues/306) feedback here too.

On that ticket you wrote:
> I changed the code of the screentock component a little. I moved the tock event code to the update event, I think it's more efficient. Now this code is not called at every next toсk of the entire scene. But now its not a screentock component...
> Before saving the screenshot, I run renderer.render() so that the scene is rendered first.
> @kfarr Let me know if its ok or no.

Yes it's ok, seems to work great!

> We can also leave the screentock code as it was.
> And rename the resulting component to screenshot??
> [...]
> I've done everything so far as part of the screentock component. Maybe it's better to rename component to screenshot now? Since there is no need to use the tock method here, as I believe. But maybe there's something I don't know?

I don't care what the component name is, but it can't be "screenshot" as that's already reserved for an a-frame component used by the system; might as well keep screentock for now unless there is a need to rename.

> And also for the screenshot, perhaps it’s worth hiding the round teleport control pointer?

Yes, I agree that should be hidden just like the other helper objects.

> For the logo
> (from PR https://github.com/3DStreet/3dstreet-editor/pull/343) Add id to logo elment to use 3DStreet logo SVG for screenshot

Thanks, I just merged that and later this week I will do a release for the current version of the Editor repo `master` branch.

kfarr:
From @Algorush 
> Thank you, so another question is: should the logo in the screenshot consist of "3DStreet Viewer/Editor" text? or only "3DStreet"?

It should just only say "3DStreet"

Algorush:
Okay, I added the code to get the logo for the screenshot from the svg element. It's time to use async/await to convert svg format to image. This is the first time I'm converting svg to image. 
Perhaps it would be easier not to remove the logo element `img.viewer-logo-img` that appears first but hide it. And use the logo from there. Take a look, @kfarr, if this code suits you.

src/components/screentock.js
```diff
@@ -14,44 +14,116 @@
...
+    // If `oldData` is empty, then this means we're in the initialization process.
+    // No need to update.
+    if (Object.keys(oldData).length === 0) { return; }
+    
     // this should be paused when not in use. could be throttled too
     if (this.data.takeScreenshot) {
-      this.el.setAttribute('screentock', 'takeScreenshot', false);
-      takeScreenshotNow(this.data.filename, this.data.type, this.data.imgElementSelector);
-    }
+      this.data.takeScreenshot = false;
```
kfarr, line 125: Hi @Algorush was there any specific reason to change this
`      this.el.setAttribute('screentock', 'takeScreenshot', false);`
into this
`       this.data.takeScreenshot = false;`

In general we're not supposed to do that with a-frame, always using `setAttribute` so that lifecycle methods / events are properly triggered, but maybe you needed to in this case for some reason?

kfarr:
Ok in testing this here is one final issue I see:
- the 3dstreet logo does not appear in editor mode, although the street title does appear -- let's wait to fix this if that's ok, since there are a lot of unmerged changes in progress on 3dstreet-editor repo

src/components/screentock.js
Algorush: through setAttribute the update event is triggered once again. In this case it is not necessary, in other cases it is different. Here takeScreenshot is used only inside the component

kfarr:
@Algorush yes please do make a correction if possible to the 3dstreet-editor repo through a new PR to master branch.

Algorush:
I've added commits with 

> @Algorush yes please do make a correction if possible to the 3dstreet-editor repo through a new PR to master branch.

I haven’t yet understood why the screenshot appears without a logo when using 3dstreet-editor.js from 3dstreet.app. But if I use the build with master branch of 3DStreet-editor then everything works well

PR #413, opened 2023-11-22
More randomly placed objects

Algorush:


kfarr:
looks great to me @Algorush !

PR #414, opened 2023-11-22
Notify methods in global STREET object

Algorush:


kfarr:
works when I tested loading a local 3dstreet json scene file

PR #417, opened 2023-11-26
Add bicycle and electro scooter animation

Algorush:
add random cyclist models also


kfarr:
Looks great!
https://github.com/3DStreet/3dstreet/assets/470477/93f098b2-08ad-486e-b0c7-4415020d38d6

PR #418, opened 2023-11-27
save grouped assets in the attribute of street-assets element

Algorush:
This function is needed to display mixins categories in the mixin list in the editor. To make this PR work: https://github.com/3DStreet/3dstreet-editor/pull/354


Algorush:
I added commit with changes. The mixin category is now stored in the `category` attribute fro each mixin. Instead of storing a list of mixins in one attribute of `street-assets` element.

kfarr:
I tested this only to make sure it doesn't break anything else. Once this is released on core then we can test on editor.

PR #428, opened 2023-12-05
remove shadow component from saved JSON

Algorush:


kfarr:
@Algorush we don't want to save this to the json because it's already provided via the mixin?

Algorush:
> @Algorush we don't want to save this to the json because it's already provided via the mixin?

yes, exactly

PR #435, opened 2023-12-15
Animation tasks (from "Animation prevent editing" issue)

Algorush:
Add first commit with first task done: toggle global animation in live mode.
No, I was wrong. Toggling only works if global animation is enabled by default. But if globalAnimation = false by default, then toggle does not work. I'm still working on it


Algorush:
Now works:

- `animation-element` component to control element animation
- enable/disable animation of a single element
- change direction of movement: inbound/outbound
- changing the position of an element in the editor and continuing the animation from the new position
- enable/disable global animation without reloading the scene

kfarr:
outdated

PR #440, opened 2023-12-29
fix save/load mapbox scene

Algorush:
- move environment and cameraRig to a-acene parent
- fix saving mapbox component data

It turned out that I had not corrected saving the scene with the mapbox component before. I fixed the intersection scene, not mapbox. Now after these changes, saving and loading works, but there are still strange errors when loading.
I'll check further. We will also need to change json-utils.js in the 3DStreet-editor repo. 
I know it's time to load json-utils.js from index.js. I did it in this PR, which probably needs some work: https://github.com/3DStreet/3dstreet/pull/352

### DIFF
--- a/src/json-utils.js
+++ b/src/json-utils.js
@@ -211,7 +211,7 @@ function filterJSONstreet (removeProps, renameProps, streetJSON) {
       }
     }
 
-    return compAttributes ?? value;
+    return compAttributes || value;
   });
   // rename components
   for (var renameKey in renameProps) {
--- a/src/json-utils.js
+++ b/src/json-utils.js
@@ -116,15 +116,7 @@ function getAttributes (entity) {
   return elemObj;
 }
 
-// properties or attributes that should be stored as a string in saved JSON
-// to prevent parsing and splitting URL 
-const propsToString = ["mapbox", "streetmixStreetURL", "streetmixAPIURL"];
-
 function toPropString (componentName, propData) {
-  if (propsToString.includes(componentName)) {
-    // return a string if key is an attribute or property listed in the propsToString array
-    return AFRAME.utils.styleParser.stringify(propData);
-  }
   if (
     typeof propData === 'string' ||
     typeof propData === 'number' ||
@@ -182,6 +174,10 @@ const renameProps = {
   street: 'not-street',
   intersection: 'not-intersection'
 };
+
+// properties or attributes that should be stored as a string in saved JSON
+// to prevent parsing and splitting URL 
+const propsToString = ["mapbox", "streetmixStreetURL", "streetmixAPIURL"];
 
 function filterJSONstreet (removeProps, renameProps, streetJSON) {
   function removeValueCheck (removeVal, value) {
--- a/src/json-utils.js
+++ b/src/json-utils.js
@@ -117,7 +117,8 @@ function getAttributes (entity) {
 }
 
 // properties or attributes that should be stored as a string in saved JSON
-const propsToString = ["mapbox"];
+// to prevent parsing and splitting URL 
+const propsToString = ["mapbox", "streetmixStreetURL", "streetmixAPIURL"];
 
 function toPropString (componentName, propData) {
   if (propsToString.includes(componentName)) {
--- a/src/json-utils.js
+++ b/src/json-utils.js
@@ -188,14 +188,11 @@ function filterJSONstreet (removeProps, renameProps, streetJSON) {
   }
 
   let stringJSON = JSON.stringify(streetJSON, function replacer (key, value) {
-    if (propsToString.includes(key)) {
-      // don't parse if key is an attribute or property listed in the propsToString array
-      return value;
-    }
-    const compAttributes = AFRAME.utils.styleParser.parse(value);
+    let compAttributes;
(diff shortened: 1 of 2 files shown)


src/json-utils.js
```diff
@@ -108,15 +108,22 @@ function getAttributes (entity) {
         if (isEmpty(modifiedProperty)) {
           elemObj['components'][componentName] = '';
         } else {
-          elemObj['components'][componentName] = toPropString(modifiedProperty);
+          elemObj['components'][componentName] = toPropString(componentName, modifiedProperty);
```
kfarr, line 111: @Algorush would this change apply to saving other components (not just mapbox) and is that ok?

kfarr:
@Algorush

I can't quite understand what is going on here with the changes to saving the component data, sorry. That's probably ok, I don't need to understand everything.

However, it makes me think that we need a standardized manual test process for testing changes to json save / load functions. Right now I just go in there and click around and try randomly to load something, make changes, save them, and reload them. But that's not a standardized procedure and you probably test differently than I do.

Or stated differently, how would you suggest testing a change like this to make sure it doesn't adversely affect saving other components?

src/json-utils.js
Algorush: Sorry, I didn't describe the reasons why I made this change.
The `mapbox` attribute's `style` property has a specific URL format that contains a colon (:) character. Example:
`style: mapbox://styles/mapbox/satellite-streets-v11`
The parser from json-utils and specifically from the JSON.stringlify function here (https://github.com/3DStreet/3dstreet/blob/938e55f00b4b1305edcde892851d9362875e2396/src/json-utils.js#L186) converts this property into this form in saved JSON file:
{style: {mapbox: {'//styles/mapbox/satellite-streets-v11'}}}.
This was the superficial reason why, when loading a JSON file with scene containing a mapbox component, the map was not loaded.
I came up with this quick solution, probably not the best one. In theory, it is better to check the contents of the stored property on the URL content using regEx, if the string contains the characters `://`

Algorush:
Yes, I completely agree that it would be very useful to write tests to test all possible cases of save/load JSON scenes.

src/json-utils.js
kfarr: Thanks. Would this have affected saving the streetmix URL in the streetmix-loader component for the same reason since it includes a colon character in the https:// ... ? Or we aren't saving that now for a different reason?

Algorush: I just looked at examples of saved JSON and only now I saw that streetmix URLs are also saved in separated form:
{'http': '//streetmix.net...'}
But now this does not affect the work, as far as I understand... Since streetmix-loader is saved as not-streetmix-loader.
So its better to use regex to check for such cases. I can make it in this branch

kfarr:
@Algorush I think this is ready for merge except for your suggestion above:
> I just looked at examples of saved JSON and only now I saw that streetmix URLs are also saved in separated form:
{'http': '//streetmix.net...'}
But now this does not affect the work, as far as I understand... Since streetmix-loader is saved as not-streetmix-loader.
So its better to use regex to check for such cases. I can make it in this branch

I'm not sure if it is needed or not to add this regex check?

Algorush:
> I'm not sure if it is needed or not to add this regex check?

not now. This may be useful when we need to save the parameters of the `streetmix-loader` component in JSON. Or if we save a url containing `://` for some attribute for future scenes

kfarr:
thanks for the update @Algorush 

Here is my plan to test:
- go the hardcoded mapbox url
- save to cloud
- restart app
- load from cloud
- see if mapbox thing displays

that is test for this PR

Then, once we are ready for a new 3dstreet "release" I should write up a flow that does a couple things like this, and testing a few other flows like importing from streetmix, etc.

Algorush:
I ended up fixing the filterJSON code. I discovered what the flaw was. Now there is no need for additional checks. I also noticed that the ocean material in the scene is now saved and loaded correctly. And now there are no error messages when loading JSON. 
I'll still test it on different scenes. But I think save/load should work as well or better

Algorush:
> thanks for the update @Algorush
> 
> Here is my plan to test:
> 
> * go the hardcoded mapbox url
> * save to cloud
> * restart app
> * load from cloud
> * see if mapbox thing displays
> 
> that is test for this PR
> 
> Then, once we are ready for a new 3dstreet "release" I should write up a flow that does a couple things like this, and testing a few other flows like importing from streetmix, etc.

this test is currently passed. And with other scenes too. Only I noticed a strange thing in the DOM inspector after load JSON: components not-street="[object object ]"

Algorush:
> Only I noticed a strange thing in the DOM inspector after load JSON: components not-street="[object object ]"

I'll see how to fix this

Algorush:
This PR is only about saving, it does not affect the JSON loading logic, so the functionality will be backwards compatible with previous JSON scenes

Algorush:
About the fact that similar things appear: `not-street='[object object ]'`
This has happened before, so it's better to create a separate issue

Algorush:
I changed operator ?? by ||. After this, it began to work better, the problem with splitting the URL into two parts went away.

kfarr:
merging this into a combined epic for testing all these changes together

PR #448, opened 2024-01-08
remove parent elements if they exists

Algorush:
Now, when changing the parameters of the `street` component, such as global-animated, show-ground, show-vehicles and others, the scene is reloaded with the new parameters. I noticed one problem - when the length of the street changes, the length of the buildings is also changes, but the length of the ground element always remains equal to 60. Also, switching the showGround parameter does not work yet

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,6 +35,14 @@ AFRAME.registerComponent('street', {
     }
 
     const streetmixSegments = JSON.parse(data.JSON);
+
+    // remove .street-parent element, if it exists, with old scene elements. 
+    // Because it will be created next in the processSegments function
+    const streetParent = this.el.querySelector('.street-parent');
+    if (streetParent) {
+      streetParent.remove();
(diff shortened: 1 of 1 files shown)


Algorush:
I noticed that buildings are still duplicated after updating the street parameters. I'm fixing it now

Algorush:
I noticed one thing with the checkbox in the right panel interface. The switch doesn't work when I click on an area outside of the small circle. But at the same time the values ​​of the input checkbox change and the checked class is added/removed from the parent div.checkboxAnim.
Here the switch is set to true. The value of the label at the top (showVehicles - value:true)
image

Here the switch is in false mode. But the value of the label above does not change (showVehicles - value:true). Vihicles  are on the scene.
image

here after I pressed the circle inside the switch. Everything works as it should
image

kfarr:
@Algorush this works really well!!!!

Issue #439, opened 2023-12-27
shadow map does not cover entire scene


kfarr:
not urgent but comes up from time to time

![image](https://github.com/3DStreet/3dstreet/assets/470477/0a4532af-b01d-48d1-8063-63fc943bf2ed)
https://3dstreet.app/#/scenes/224d2558-1b35-4a7a-bd37-8bc1c2c4dfe9.json

Algorush:
this happens because the scope of light.shadow.camera (top/bottom/left/right) does not cover the entire scene. We can make automatically resizing the shadow camera according to the size of the scene or just do these values very big for maximum scene size

PR #449, opened 2024-01-08
add shadowCameraAutomatic option

Algorush:
- add shadowCameraAutomatic option with #default-street element for automatic shadow camera sizing
- delete shadowCamera parameters

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -403,11 +403,11 @@ AFRAME.registerComponent('street-environment', {
       sky.setAttribute('color', '#FFF');
       sky.setAttribute('src', '#sky');
       sky.setAttribute('rotation', '0 20 0');
-      light2.setAttribute('light', 'intensity: 2.2; castShadow: true; shadowCameraBottom: -20; shadowCameraLeft: -30; shadowCameraRight: 40; shadowCameraTop: 30; shadowMapHeight: 2048; shadowMapWidth: 2048');
+      light2.setAttribute('light', 'intensity: 2.2; castShadow: true; shadowMapHeight: 2048; shadowMapWidth: 2048');
       light2.setAttribute('position', '-40 56 -16');
     } else if (this.data.preset === 'sunny-morning') {
       light1.setAttribute('light', 'intensity', 0.8);
-      light2.setAttribute('light', 'intensity: 2.2; castShadow: true; shadowCameraBottom: -20; shadowCameraLeft: -30; shadowCameraRight: 40; shadowCameraTop: 30; shadowMapHeight: 2048; shadowMapWidth: 2048');
+      light2.setAttribute('light', 'intensity: 2.2; castShadow: true; shadowMapHeight: 2048; shadowMapWidth: 2048');
       light2.setAttribute('position', '-60 56 -16');
       sky.setAttribute('visible', true);
       sky.setAttribute('color', '#FFF');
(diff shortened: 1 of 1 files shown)


Algorush:
I'm still testing this option for automatic shadow camera sizing. It is good for determining the size of the shadow area. But I want to test whether the performance has changed

Algorush:
I noticed that if use the shadowCameraAutomatic option for the light component, the scene slows down noticeably. This happens after turning on the editor and going to the viewer again. Perhaps because with this option the size of the object passed to shadowCameraAutomatic is calculated every tick, but not sure.
I sized the shadow camera according to the length of the street and added 10 meters to each side since the camera is aimed at an angle. For all enviroment modes, shadows from all objects are displayed on the scene. Tested with different street length

kfarr:
A couple thoughts:
- I'm not sure "default-street" length is the source of truth for how to position camera shadow in a scene -- there may be multiple streets, intersections or other user-defined geometry in the scene. This element id may not always be present.
- What is the best way to test this? What is the goal here? Is it to have shadow on any object that is in the camera field of view? Is that possible with A-Frame?
- You mentioned performance issues with `shadowCameraAutomatic`. Is `shadowCameraAutomatic` supposed to automate the above? Does it work for this use case ignoring the performance issue? If so maybe it's worth digging a bit more to fix the root cause?

Algorush:
> I'm not sure "default-street" length is the source of truth for how to position camera shadow in a scene -- there may be multiple streets, intersections or other user-defined geometry in the scene. This element id may not always be present.

Yes, I thought about that too. Maybe it's worth storing length in a STREET object?

>What is the best way to test this? What is the goal here? Is it to have shadow on any object that is in the camera field of view? Is that possible with A-Frame?

Yes, the goal here is to have shadows on all objects. I came from here: 
https://github.com/3DStreet/3dstreet/issues/439
Yes, it's possible in AFrame, why not? Just need to adjust the `light.shadow.camera` area. The problem is that this area is adjusted to the light source, not the viewer's camera. It would be possible to simply specify very large values ​​for the shadow.camera area, but then the shadow becomes pixelated in small scenes

>  Is `shadowCameraAutomatic` supposed to automate the above? Does it work for this use case ignoring the performance issue? If so maybe it's worth digging a bit more to fix the root cause?

Yes, this works if we specify in  `shadowCameraAutomatic` a scene element in the scene that the light and shadow should cover. But what the root cause you mean? Not quite understand

Algorush:
> I'm not sure "default-street" length is the source of truth for how to position camera shadow in a scene -- there may be multiple streets, intersections or other user-defined geometry in the scene. This element id may not always be present.

Maybe we just can use #street-container instead of #default-street?

kfarr:
@Algorush yes one level higher? So in theory `shadowCameraAutomatic` automatically adjusts for the combination of the camera frustum and the bounding box of the target entity?

Algorush:
I looked at the code of the light component and it turns out that cameraAutomaticShadowds is calculated every tick if enabled:
https://github.com/aframevr/aframe/blob/770f95a0cda0ad083c6d6f8d6ce4399804f238d5/src/components/light.js#L155
Therefore, this led to severe performance losses when using it. I can take a function from the light component that determines the exact dimensions of the shadowCamera, but I think that it is very large and in our case it is enough to simply specify the maximum side of the vector with the size of streetContainer. And add 10 meters to take into account that the camera light is at an angle.

Algorush:
calculating the dimensions of a `street-container` is a little more difficult than it seems. We can't use the element's `loaded` event since we are loading the content via JS. I tried different options, the option worked well when I added shadowCameraAutomatic with the `street-container` elements, and then turned off this option through `setTimeout`, after few seconds (5 is enough). Works good. But on this scene (https://streetmix.net/kfarr/3/3dstreet-demo-street) an error occurs. It looks like the problem is that one of the segments did not load and has `NaN` values in one of properties

Algorush:
I'll try to do something about it next time. For now I'll deal with other issues

PR #454, opened 2024-01-16
fix boundingbox for tram and foodTruck elemtens

Algorush:
I found a way to fix the wrong `boundingBox` for tram element and also this applies to the foodTruck element


kfarr:
- I tested to see if tram and food truck are placed correctly, they are (no regression)
- I could not reproduce the bounding issue to verify if it is fixed

since there is no regression error I will merge anyway

PR #455, opened 2024-01-16
Fix hash marks

kfarr:
to complete
- [x] - revert bus segment to previous behavior
- [x] - fix repeat count for other divider cases


kfarr:
Possible implementation methods:

1) replace the existing divider surface with the new design and superimpose the existing road surface behind it -- easier but adding another texture and naive method will change angle of hash
2) new logic -- if a variant is "divider" and the ground mixin id has not been changed to elevated grass or sidewalk, then use default drive-lane mixin for segment texture and add repeating hash like other stencils

kfarr:
progress using method 1
image

kfarr:
before: 80kb (53+27kb) for hatched and hatched normals
now: 42kb just for seamless-lane-with-hatch-half.jpg (saving ~40kb)

PR #461, opened 2024-01-19
Add new animated chars

Algorush:
Not done yet. I also need to make different speeds for different characters. 
And there is one issue to fix. After switching the globalAnimation in the street component to true, the added characters stand with their arms outstretched to the sides in the Editor, but after switching to View everything is fine


Algorush:
It looks like Node v12 doesn't support "nullish coalescing operator (??)". I'll change it

kfarr:
> It looks like Node v12 doesn't support "nullish coalescing operator (??)". I'll change it
we can upgrade node, i think 12 is pretty old now right?

kfarr:
a few questions @Algorush 
- so there is no more "people-rigged" category any more? now it is just "people" category that includes "achar..." mixin IDs.
- how will this affect scenes with the old character mixin ID saved? How do support these legacy scenes?
- what are the "default categories" of objects that are loaded now? all of them? it was hard for me to tell in assets.js.

PR #465, opened 2024-01-29
Fix saving user modified texture

Algorush:
For this change to work, we need first make changes from this PR: https://github.com/3DStreet/3dstreet/pull/440. 
This is necessary so that the texture url is saved in the correct format in JSON. 
I tested on different scenes, so far everything works well.


Algorush:
These changes are now working well. But maybe it's worth refactoring

PR #468, opened 2024-02-01
call json-utils.js from index.js

Algorush:
move functions used in 3DStreet-editor and in index.html (and another HTMLs) from json-utils to a global `STREET` object


Algorush:
I will also update the function calls from json-utils.js in the Editor repo

Algorush:
In this PR, all changes are related to requesting json-utils from index.js and calling json-utils functions from the global STREET.utils object. But I also added corrections to this PR from here (https://github.com/3DStreet/3dstreet/pull/440) regarding texture saving. Because incorrect saving and loading of textures led to errors on the screen when saving/loading from the Editor. I forgot to mention this.

src/json-utils.js
```diff
@@ -175,7 +178,7 @@ const renameProps = {
   intersection: 'not-intersection'
 };
 
-function filterJSONstreet (removeProps, renameProps, streetJSON) {
```
kfarr, line 178: is this function called from 3dstreet editor? it may cause this error
https://github.com/3DStreet/3dstreet-editor/blob/f763a949ce4cd643573ac7f3f7e321d8bbfa8c6a/src/components/scenegraph/Toolbar.js#L130C26-L130C37

kfarr:
changes needed on the editor:
- need to update index.html to match
- may need to update calls to `filterJSONstreet` function from editor, only pass `streetJSON` and not removeProps, renameProps

src/json-utils.js
kfarr: Yes, that is why this PR should be applied to editor after to prevent that error:
https://github.com/3DStreet/3dstreet-editor/pull/387

PR #480, opened 2024-02-25
Svg extruder issues

Algorush:


kfarr:
@Algorush this fixes texture saving issues, so this is pareto superior

there is still rotation issue but we can resolve later

PR #482, opened 2024-02-26
fix rotation and other issues

Algorush:


kfarr:
all seems to work great, merging!

Issue #476, opened 2024-02-23
support new streetmix metric plan


kfarr:
Currently streetmix json is in imperial units. Instead it will soon offer metric units.

The key code change is:
* For schema version >= 30 -- All units in API response are metric
* For schema version < 30, all units in API response are imperial

More notes and sample API responses for new metric responses: 
https://github.com/streetmix/streetmix/issues/2900

kfarr:
Example response from streetmix API with new metric:
[streetmix-metric-example-response.json](https://github.com/3DStreet/3dstreet/files/14412226/streetmix-metric-example-response.json)

Algorush:
I suggest to change once in one place all the width values ​​in the object with the data received from the streetmix API. So as not to convert to meters everywhere where necessary. @kfarr what do you think about this? 
Because, the processSegments function in aframe-streetmix-parser recieve only information about the contents of segments. But the schema version, as well as units, are at a higher level in recieved JSON. That is, there will be no need to change the code

Algorush:
> More notes and sample API responses for new metric responses:

PR #486. Tested with all cases from here [streetmix/streetmix#2900](https://github.com/streetmix/streetmix/issues/2900)

kfarr:
Just sharing notes testing with a streetmix staging server providing metric responses

Testing technique, load the 3dstreet app and type the following in console:
```
newEl = document.createElement('a-entity')
newEl.setAttribute('streetmix-loader', 'streetmixAPIURL: https://streetmix-staging.herokuapp.com/api/v1/streets?namespacedId=295')
parentEl = document.querySelector('#street-container')
parentEl.appendChild(newEl)
```

Behavior on 3dstreet app (which does not yet support metric) as expected is incorrect:
![image](https://github.com/3DStreet/3dstreet/assets/470477/2205fe4f-a05e-49e8-952d-ee5b2903eb73)

Behavior on github.3dstreet.org with NEW streetmix metric API response which supports the new metric api is correct:
![image](https://github.com/3DStreet/3dstreet/assets/470477/bbfb73f8-e665-495d-a521-a60b7b2cac16)

Behavior on github.3dstreet.org with OLD streetmix imperial API response:
![image](https://github.com/3DStreet/3dstreet/assets/470477/5a4dd07c-8c16-4d2f-92f9-181dcf6d477c)

Therefore both imperial and metric responses work as expected on the new release. I think we're good!

PR #486, opened 2024-02-28
support for new and old metric plan

Algorush:
https://github.com/3DStreet/3dstreet/issues/476


src/tested/streetmix-utils.js
```diff
@@ -51,10 +51,21 @@ module.exports.streetmixAPIToUser = streetmixAPIToUser;
...
-    const segmentWidthInMeters = segmentWidthInFeet * 0.3048;
-    cumulativeWidthInMeters = cumulativeWidthInMeters + segmentWidthInMeters;
+    cumulativeWidthInMeters = cumulativeWidthInMeters + currentSegment.width;
   });
   return cumulativeWidthInMeters;
 }
 module.exports.calcStreetWidth = calcStreetWidth;
+
+// convert all feet values to meters for schemaVersion < 30
+function convertStreetValues (streetData) {
```
kfarr, line 61: Thanks @Algorush , I really like this approach to modify the widths "in situ" only if needed instead of distributing that logic elsewhere in the app.

kfarr:
@Algorush in `index.js` I would suggest that we renamed the variable `streetmixSegmentsFeet` to `streetmixSegmentsMetric`

Everything else looks ok from a code review perspective, testing will be next

kfarr:
> Also might it be better to remove the calcStreetWidth function from streetmix-utils? Because it is simple now.

sure!

kfarr:
This looks good to me. Here was my testing procedure:

I temporarily modified line 92 of index.html https://github.com/3DStreet/3dstreet/blob/main/index.html#L92

old
`        <a-entity id="default-street" street streetmix-loader set-loader-from-hash></a-entity>`

new
`        <a-entity id="default-street" street streetmix-loader="streetmixAPIURL: http://localhost:7001/examples/streetmix-schema28.json"></a-entity>` for testing old schema, and
`        <a-entity id="default-street" street streetmix-loader="streetmixAPIURL: http://localhost:7001/examples/streetmix-schema30.json"></a-entity>` for testing new schema

both seem to work ok with dimensions as expected (per visual inspection)

PR #496, opened 2024-03-06
add bevel and second material support

Algorush:


kfarr:
hey @Algorush 


* using thie repo I am loading this file [1]

here is result - now all svg extrude elements have a grass top face
image

Then I set `bevelEnabled` and `topElement` default to `false` and the scene loaded as expected

Then I edited the scene and on the svg path - circle bioswale layer I set `bevelEnabled` and `topElement` to true:
* I noticed that once `topElement` is set to true in the Editor, even when setting to `false` the topElement face is still visible

Then I saved the file into a new json file and reloaded. Then I noticed that the bevel has a grass texture instead of concrete
image

Here is the new json file
[test-file-after-enable-bevel-and-top-element.json](https://github.com/3DStreet/3dstreet/files/14519337/test-file-after-enable-bevel-and-top-element.json)

[1] [svg-test-file-data.json](https://github.com/3DStreet/3dstreet/files/14519236/svg-test-file-data.json)

Algorush:
I fixed issue with visible `topLevel` in `false`.


> Then I saved the file into a new json file and reloaded. Then I noticed that the bevel has a grass texture instead of concrete

It because JSON saved with material component from previous version of svg-extruder. And this material is applied to element. Quick solution - remove material component string from JSON in this element. And also should I make cleaning component material data in init svg-extruder?

Algorush:
A short guide to the svg-extruder component.
   `depth` - float. Depth to extrude the shape. Default is 1.
   `bevelEnabled` - bool. Apply beveling to the shape. Default is true.
   `bevelThickness` - float. How deep into the original shape the bevel goes. Default is 0.2.
   `bevelSize` - float. Distance from the shape outline that the bevel extends. Default is bevelThickness - 0.1.
   `bevelOffset` - float. Distance from the shape outline that the bevel starts. Default is 0.
   `bevelSegments` - int. Number of bevel layers. Default is 3.
   `topElement` - bool. Add a separate top element - covering. default is false.
   `color` - type: 'color'. Top element color, coating. Default is 'grey'.
   `src` - string. Address of the link to the texture html element for the top element, cover. Default is '#grass-texture'.
   `bevelColor` - type: 'color'. Color of extruded element. Default is 'white'.
   `bevelSrc` - string. Address of the texture for the extruded element.
   `lineColor` - type: 'color'. Border line color.
If old scenes contain a version of the `svg-extruder` component without bevel and topElement, their materials will not be applied to the new version of the component. You need to reassign the `src` of texture for the top (covering) and for the extruded element (border)

kfarr:
Thanks! This is looking great:
image

PR #500, opened 2024-03-15
Load from url encoded json

Algorush:
to load from url encoded streetmix json you need:
1. Url encoded streetmix JSON. JSON can be obtained from the Editor, `StreetmixAPIURL` parameter in the `streetmix-loader` component of the #default-street element. String can be encoded through encodeURIComponent func in console or here for example: https://www.urlencoder.org/
2. Paste encoded string after `http://localhost:7001/#streetmix-json:` or `http://localhost:7001/#3dstreet-json:` for 3dstreet data

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -67,13 +67,57 @@ AFRAME.registerComponent('streetmix-loader', {
     streetmixStreetURL: { type: 'string' },
     streetmixAPIURL: { type: 'string' },
     showBuildings: { default: true },
-    name: { default: '' }
+    name: { default: '' },
+    streetmixJSON: { type: 'string', default: '' }
+  },
+  loadFromJSON: function (streetmixResponseObject) {
+    if (!streetmixResponseObject) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -67,14 +67,58 @@ AFRAME.registerComponent('streetmix-loader', {
     streetmixStreetURL: { type: 'string' },
     streetmixAPIURL: { type: 'string' },
     showBuildings: { default: true },
-    name: { default: '' }
+    name: { default: '' },
+    streetmixJSON: { type: 'string', default: '' }
```
kfarr, line 71: I think this could be set directly on the `street` component instead of `streetmix-loader` component? there is already a property called `JSON` on the `street` component?

kfarr:
Thanks please see:

https://github.com/3DStreet/3dstreet/pull/500/files#r1529712137
and
https://github.com/3DStreet/3dstreet/issues/502

PR #503, opened 2024-03-22
Streetplan support

Algorush:
The first version is currently working with the test response from Streetplan: exampleJOSN.json. Buildings are not yet supported. The remaining segments mostly now work with default options.

What's next: 
- add other subtype options that can be displayed in 3DStreet. 
- add Buildings, setback.
- add support for basic materials and markings. Currently, materials are selected only for bus-lane, bike-lane, streetcar, light-rail. But it would be nice to add other material options. 
- add support mixed segments in 3DStreet. 
- It would also be nice to add support for some other segment options in 3DStreet, which are quite easy to add: for example, empty roads, empty bus-lane, bike-lane,...

Also I'll make suggestions for Streetplan developer for JSON structure. For example it will be good to save units in JSON - feet/meters


Algorush:
Added Setback segment support in last commits.
I forgot to say - streetplan.html is a test file with the streetplan-loader component. Currently, while cors on the Streetplan side is not configured, a test JSON file with the Streetplan scene is used, located in the `src/streetplan/exampleJSON.json`

PR #508, opened 2024-04-10
3dtiles. Add orthographic camera support to example

Algorush:
Here I added new minified version of aframe-loader-tiles-component with orthographic camera support. 
This is a temporary version of the component until I add ortho-camera support to the three-loader-3dtiles library. But here the orthographic camera works just as well, I haven't noticed any issues yet.


kfarr:
outdated pr, closing

Issue #485, opened 2024-02-28
new scene function - reference layer is not removed when loading new scene


Algorush:
If open a scene with a reference layer, for example with a mapbox component, then if then load another scene from JSON, the reference layer of the old scene remains in the new scene

kfarr:
I propose to have a "new scene" function that is called before loading a new scene, and will also be a user interface button that allows a user to create a blank new 3dstreet scene.

A blank scene should have all the base layers as though loading from index.html without a scene.

The new scene function should also reset metadata of scene id and title to empty strings.

On the editor, we have 4 places to call this "new scene" function:
- when loading a 3dstreet scene from cloud via open modal
- when loading a 3dstreet scene from upload json via open modal
- when loading a streetmix street from open modal
- when the user presses a "new" button in the user interface

kfarr:
@Algorush is this ready for review?

Algorush:
yes

PR #509, opened 2024-04-20
New scene function

Algorush:
Added a newScene function to clear the old scene and create an empty scene. Perhaps there is something else to consider in this function, @kfarr ?
In the next commits I will add call of this function to the appropriate places in the code.
https://github.com/3DStreet/3dstreet/issues/485


Algorush:
what should be the initial structure of Blank Scene? Like this one?
```
      <a-entity id="street-container" data-layer-name="3D Street Layers" data-layer-show-children>
        <a-entity id="default-street" set-loader-from-hash></a-entity>
      </a-entity>
      <a-entity id="reference-layers" data-layer-name="Reference Layers" data-layer-show-children></a-entity>
      <a-entity id="environment" data-layer-name="Environment" street-environment="preset: day;"></a-entity>
```

Algorush:
I have added newScene funtion as STREET.utils.newScene. I added a call to this function in the appropriate scene loading methods, including in the Editor code, in this PR: https://github.com/3DStreet/3dstreet-editor/pull/420

kfarr:
things this does not reset yet:
* currently selected object blue bounding box
* long / lat / elevation in geopanel above scene title
* camera pose (position and rotation) -- this is probably ok to leave for a later date

image

kfarr:
similar to scene title edit modal, when the street-geo component is updated, we should also fire an event such that the react app is listening for this event and 'resets' the geo panel?

vincentfretin:
> similar to scene title edit modal, when the street-geo component is updated, we should also fire an event such that the react app is listening for this event and 'resets' the geo panel?

@kfarr and I started a GeoContext on @kfarr machine. Almost finished, we just need to reset the context value after calling newScene. We will finish that Friday.

vincentfretin:
Also probably put the "New" button behind a flag or just comment it before merging.
Like discussed, just having a new blank scene, without some templates or onboarding is confusing for the user.

kfarr:
@vincentfretin i committed our WIP work and tested, all seems to work with exception of resetting geo stuff as you noted so we can resume that on Fri

PR #512, opened 2024-05-03
Change street component parameters

Algorush:
Added the ability to change the parameters of the street component (showGround, showVehicles, showStriping) without reloading the scene. It remains to add globalAnimated. So this PR not ready yet.
Also, now there is no need to rename street to not-street, like streetmix-loader and intersection. I added a check that the scene is loaded from jsonFile and do not run the update function in these components

### DIFF
--- a/src/components/streetplan-loader.js
+++ b/src/components/streetplan-loader.js
@@ -75,6 +75,8 @@ AFRAME.registerComponent('streetplan-loader', {
       return;
     }
 
+    STREET.sourceType = 'streetplanURL'; // it also could be jsonFile/streetmixURL
+    
     var request = new XMLHttpRequest();
     console.log('[streetplan-loader]', 'GET ' + data.streetplanAPIURL);
 
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ if (typeof VERSION !== 'undefined') { console.log(`3DStreet Version: ${VERSION}`
 var streetmixParsers = require('./aframe-streetmix-parsers');
 var streetmixUtils = require('./tested/streetmix-utils');
 require('./json-utils_1.1.js');
+require('./street-utils.js');
 require('./components/gltf-part');
 require('./components/ocean');
 require('./components/svg-extruder.js');
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,6 @@ if (typeof VERSION !== 'undefined') { console.log(`3DStreet Version: ${VERSION}`
 var streetmixParsers = require('./aframe-streetmix-parsers');
 var streetmixUtils = require('./tested/streetmix-utils');
 require('./json-utils_1.1.js');
-require('./street-utils.js');
 require('./components/gltf-part');
 require('./components/ocean');
 require('./components/svg-extruder.js');
@@ -39,17 +38,6 @@ AFRAME.registerComponent('street', {
     }
 
     const streetmixSegments = JSON.parse(data.JSON);
-
-    // remove .street-parent and .buildings-parent elements, if they exists, with old scene elements.
-    // Because they will be created next in the processSegments and processBuildings functions
-    const streetParent = this.el.querySelector('.street-parent');
-    if (streetParent) {
-      streetParent.remove();
-    }
-    const buildingParent = this.el.querySelector('.buildings-parent');
-    if (buildingParent) {
-      buildingParent.remove();
-    }
 
     const streetEl = streetmixParsers.processSegments(streetmixSegments.streetmixSegmentsMetric, data.showStriping, data.length, data.globalAnimated, data.showVehicles);
     this.el.append(streetEl);
@@ -109,6 +97,7 @@ AFRAME.registerComponent('streetmix-loader', {
 
         const streetmixName = streetmixResponseObject.name;
         console.log('streetmixName', streetmixName);
+
         el.setAttribute('streetmix-loader', 'name', streetmixName);
 
         let currentSceneTitle;
--- a/src/index.js
+++ b/src/index.js
@@ -38,6 +38,18 @@ AFRAME.registerComponent('street', {
       return;
     }
 
+    // remove .street-parent and .buildings-parent elements, if they exists, with old scene elements.
+    // Because they will be created next in the processSegments and processBuildings functions.
+    // this is also necessary when changing the parameters of the street component to reload the scene    const streetParent = this.el.querySelector('.street-parent');
+    const streetParent = this.el.querySelector('.street-parent');
+    if (streetParent) {
+      streetParent.remove();
+    }
+    const buildingParent = this.el.querySelector('.buildings-parent');
+    if (buildingParent) {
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -28,19 +29,59 @@ AFRAME.registerComponent('street', {
...
+  },
+  toggleEntitiesVisibillity: function (entitiesArray, visible) {
+    entitiesArray.forEach(entity => entity.setAttribute('visible', visible));
+  },
+  toggleVehicles: function (showVehicles) {
+    const vehicleEntities = streetUtils.getVehicleEntities();
+    this.toggleEntitiesVisibillity(vehicleEntities, showVehicles);
+  },
+  toggleGround: function (showGround) {
+    const groundEntities = Array.from(document.querySelectorAll('.ground-left, .ground-right'));
```
vincentfretin, line 44: Array.from is not needed.

Algorush:
Updated this PR to the latest version of the 3dstreet repository. This issue has now been fixed: https://github.com/3DStreet/3dstreet/issues/673

src/components/streetplan-loader.js
```diff
@@ -78,6 +78,8 @@ AFRAME.registerComponent('streetplan-loader', {
       return;
     }
 
+    STREET.sourceType = 'streetplanURL'; // it also could be jsonFile/streetmixURL
```
kfarr, line 81: @Algorush there will be more than 1 street per scene so I don't understand why this needs to be set ?

src/index.js
```diff
@@ -34,9 +34,50 @@ AFRAME.registerComponent('street', {
...
+  },
+  toggleStriping: function (showStriping) {
+    const stripingEntities = streetUtils.getStripingEntities();
+    this.toggleEntitiesVisibillity(stripingEntities, showStriping);
+  },
   update: function (oldData) {
-    // fired once at start and at each subsequent change of a schema value
-    var data = this.data;
+    // do not call the update function when the component is initially loaded and sourceType is jsonFile
+    if (STREET.sourceType === 'jsonFile' && this._initLoad) {
```
kfarr, line 60: @Algorush we should not be using a global for sourceType that describes this specific entity; multiple streets may be loading from different types of sources on the same scene;

src/index.js
```diff
@@ -101,8 +142,11 @@ AFRAME.registerComponent('streetmix-loader', {
     // fired at start and at each subsequent change of any schema value
     // This method may fire a few times when viewing a streetmix street in 3dstreet:
     // First to find the proper path, once to actually load the street, and then subsequent updates such as street name
-    var data = this.data;
-    var el = this.el;
+    const data = this.data;
+    const el = this.el;
+
+    // do not call update function while loading scene from JSON file
+    if (STREET.sourceType === 'jsonFile') return;
```
kfarr, line 149: same feedback, not a global

kfarr (commented):
please add a new property to street component such as `synchronized` true/false

if synchronized is true (it should be true by default at the start) then it will create children entities; after creating children entities for the first time synchronized should be set to false; then use that instead of the global STREET.sourceType

src/components/streetplan-loader.js
Algorush: Yes, I thought about more than 1 street per scene and was tested with this case also. `STREET.sourceType` refers to the currently loaded street. 
And yes, maybe its not good idea, but I didn’t find a better option when was thinking about it.

Algorush:
> please add a new property to street component such as `synchronized` true/false
> 
> if synchronized is true (it should be true by default at the start) then it will create children entities; after creating children entities for the first time synchronized should be set to false; then use that instead of the global STREET.sourceType

Ok, it works good, when I added `synchronized` flag for streetmix-loader and stretplan-loader too. But I dont understand the logic why `synchronized` should be true by default?

Algorush:
question: should visibility options showGround. showStripping, showVehicles be saved in JSON or should they only be valid during the session? 
I have now made that these parameters are updated without reloading the scene. The reason for this was that if the user changed one of these parameters, the entire street would be reloaded and all his changes would be lost.

src/index.js
```diff
@@ -185,6 +228,8 @@ AFRAME.registerComponent('streetmix-loader', {
           JSON.stringify({ streetmixSegmentsMetric: streetmixSegments })
         );
         el.emit('streetmix-loader-street-loaded');
+        // the streetmix data has been loaded, set the synchronized flag
+        data.synchronized = true;
```
vincentfretin, line 232: Don't set it like that, otherwise the change won't be properly exported. Use
`this.el.setAttribute('street', 'synchronized', true}`

src/index.js
```diff
@@ -171,7 +214,7 @@ AFRAME.registerComponent('streetmix-loader', {
           );
         }
 
-        el.setAttribute('data-layer-name', 'Streetmix • ' + streetmixName);
+        el.setAttribute('data-layer-name', 'Streetmix � ' + streetmixName);
```
vincentfretin, line 217: unicode issue?

src/index.js
```diff
@@ -86,6 +123,8 @@ AFRAME.registerComponent('street', {
       );
       this.el.append(buildingsEl);
     }
+    // the scene has been loaded, set the synchronized flag
+    this.data.synchronized = true;
```
vincentfretin, line 127: use setAttribute

vincentfretin (commented):
`synchronized` means the entities are recreated each time from streetmix json.
If we don't want the behavior, we set `synchronized` to false after loading the json the first time.

kfarr:
@Algorush sorry for the miscommunication, this is close but the logic is inverted.

Instead of "this.data.synchronized = true;" we would like to have `synchronize` be the variable and the default is true. If the value of synchronize is true, then please yes do the actual element creation; if it is false, then do not. Effectively this is the opposite logic of what you have implemented.

src/index.js
Algorush: Ok, I'll do tis way. But I tested export and import and it works well

vincentfretin: The current code may work on aframe 1.5.0 but I think it won't work on aframe 1.6.0 or master because of internals restructuring, calling getAttribute('street') won't return the modified data object.

src/index.js
Algorush: yes, something strange with my code editor. Will fix in next commit too

src/index.js
Algorush: Thanks for clarifying. Yes, I have read the documentation why it is better to use setAttribute. But in general it would be better to use an internal variable of the component in these cases, which is saved in JSON, but will not be displayed in the editor. There is no such mechanism in the component implementation, but it could be done using __innerAttrName and internal get/set functions

kfarr: @Algorush it's ok for this variable to be exposed to the end-user

kfarr:
test
- import a streetmix scene
- save / reload that 3dstreet scene
- see if we see props panel
- if we are able to use the component properites -- great! that's a bonus

Issue #527, opened 2024-05-21
fix linting errors


kfarr:
there are many linting errors that require manual work to fix them

Algorush:
Now eslint is configured more strictly. Therefore, I will list some controversial points that had to be made.
- About different versions of `json-utils`. I think we can leave only last one now? I'll delete the rest.
- In `index.js` I'll comment:
```
if (typeof VERSION !== 'undefined') {
  console.log(`3DStreet Version: ${VERSION}`);
}
```

vincentfretin:
Yes keep only json-utils_1.1.js that's the one used, you can remove the other versions.

Keep the 3 lines, just add `/* global VERSION */` on top. webpack is setting this VERSION variable.

Algorush:
> Yes keep only json-utils_1.1.js that's the one used, you can remove the other versions.
> 
> Keep the 3 lines, just add `/* global VERSION */` on top. webpack is setting this VERSION variable.

Ok, thanks. I have problems running the command in my localhost: `npm run lint`.
```
No files matching the pattern "'src/**/*.js'" were found.
Please check for typing mistakes in the pattern.
```
Although I run it from 3DStreet root dir.

Algorush:
> That's weird. Which OS are you using? Windows? Can you try to change the lint command to `eslint --ext .js,.jsx src` to see if that helps?

Yes, windows. The problem went away when I removed the quotes here: `'src/**/*.js'`
Or if escape double quotes, it will work on unix/windows, as chatGPT says:
`"lint": "eslint \"src/**/*.js\" \"src/**/*.jsx\"",`

vincentfretin:
Ok, let's change the command to `eslint --ext .js,.jsx src` there is no quotes in it.

Algorush:
Its is done now: https://github.com/3DStreet/3dstreet/pull/530

kfarr:
closed by #530

PR #530, opened 2024-05-23
fix lint errors

Algorush:
Fix issue: https://github.com/3DStreet/3dstreet/issues/527


src/components/streetplan-loader.js
```diff
@@ -63,17 +61,10 @@ AFRAME.registerComponent('streetplan-loader', {
...
-    // /*  ***** debugging *****  */
-    // setTimeout(()=> {
-    //   this.streetplanResponseParse(exampleJSON);
-    // }, 1000);
-    // return;
 
     // load from URL encoded Streetplan JSON
     if (data.streetplanEncJSON) {
-      const streetplanJSON = decodeURIComponent(encodedString);
+      const streetplanJSON = decodeURIComponent(data.streetplanEncJSON);
```
vincentfretin, line 67: It seems you fixed a real issue here. Great.

src/editor/components/components/SceneCard/SceneCard.component.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
```
vincentfretin, line 1: You need to use useCallback for handleClickOutside

src/editor/components/components/SceneEditTitle/SceneEditTitle.component.jsx
```diff
@@ -60,7 +60,7 @@ const SceneEditTitle = ({ sceneData }) => {
           <input
             className={styles.title}
             value={title}
-            onChange={handleChange}
+            // onChange={handleChange}
```
vincentfretin, line 63: remove?

Algorush: Judging by how I understood the code, there is no need to use the `onChange` handler for this `input`, at least now. Because the new scene title is entered in the prompt window. Related issue: https://github.com/3DStreet/3dstreet-editor/issues/311

PR #537, opened 2024-05-28
New json export

vincentfretin:
Main differences between the previous implementation and the new implementation:
Previous implementation has its own `toPropString` function that does its own thing and using `AFRAME.utils.coordinates.stringify(propData)` only for vectors instead of using `stringifyComponentValue` from entity.js that uses `schema.stringify` that have the implementation of stringify for each property types. 

In the new export I'm doing the filters early while doing `prepareForSerialization`, that's cloning the entity (but not attached to the DOM) and then call `removeAttribute` if needed.
In previous implementation, it's iterating over the existing DOM, serialize each component with the "not include default values logic" that exists in entity.js but functions were copied from entity.js (not sure if there are some changes to those), and do a filter afterwards on the json with `filterJSONstreet`.

### DIFF
--- a/src/editor/components/scenegraph/Toolbar.js
+++ b/src/editor/components/scenegraph/Toolbar.js
@@ -20,7 +20,7 @@ import { uploadThumbnailImage } from '../modals/ScreenshotModal/ScreenshotModal.
 import { sendMetric } from '../../services/ga.js';
 import posthog from 'posthog-js';
 import { UndoRedo } from '../components/UndoRedo';
-// const LOCALSTORAGE_MOCAP_UI = "aframeinspectormocapuienabled";
+import { downloadJSON, exportSceneToJSON } from '../../lib/entity.js';
 
 function filterHelpers(scene, visible) {
   scene.traverse((o) => {
@@ -128,7 +128,7 @@ export default class Toolbar extends Component {
     }
   };
 
-  static convertToObject = () => {
+  static convertToObjectOld = () => {
     try {
       posthog.capture('convert_to_json_clicked', {
         scene_id: STREET.utils.getCurrentSceneId()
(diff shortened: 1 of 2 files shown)


vincentfretin:
The previous export also doesn't round the vec3, we can get
`"position": "0 -1.622889224056475e-15 -0.6911566952756658"`
but that's still valid.

vincentfretin:
Other differences between the two implementations.
The old json export is setting the "primitive" key
https://github.com/3DStreet/3dstreet/blob/afb7aa2862a27ff427854be412d16c74800a0a02/src/json-utils_1.1.js#L108-L111

and the loader read it here
https://github.com/3DStreet/3dstreet/blob/afb7aa2862a27ff427854be412d16c74800a0a02/src/json-utils_1.1.js#L412-L423
not sure why we have that `delete entityData.components[attr]; ` line here.

example of exported json
```json
                    {
                      "element": "a-entity",
                      "mixin": "markings solid-stripe",
                      "primitive": "plane",
                      "components": {
                        "geometry": "buffer: false; skipCache: true; width: 0.2; height: 150",
                        "material": "repeat: 1 5; transparent: true; src: #markings-atlas",
                        "scale": "1 0.4 1",
                        "rotation": "270 0 0",
                        "position": "0 0.01 0"
                      }
                    }
```
There is no primitive prop in geometry component, I'm not sure if this is deleted explicitly in the code, I didn't found that.
I don't know why this was implemented this way initially and what was the issue that resulted doing this implementation.

Another difference is the exported class is an array.

vincentfretin:
old implementation has a way to remove some properties inside a component
https://github.com/3DStreet/3dstreet/blob/afb7aa2862a27ff427854be412d16c74800a0a02/src/json-utils_1.1.js#L175-L184

vincentfretin:
Related to the export with the original functions, there is two changes in aframe inspector:
- https://github.com/aframevr/aframe-inspector/commit/84cb492ebded10bfbfe12fa952a654a5331efd40
- https://github.com/aframevr/aframe-inspector/commit/a2aefc829c3906bfc3e06c65752c6ead03db176b

kfarr:
outdated

PR #548, opened 2024-05-31
Fix regression of exporting vec2 and vec3 like rotation (fix #534)

### LINKED ISSUE #534: The saved JSON scene is not loading correctly
Scene before saving: 
image
Scene after loading: 
image

Perhaps the reason is as follows. I noticed in the dev version that some of the vector values ​​were saved differently: instead of `rotation: '0 90 0'` it is now saved as `rotation: 'x:0 y: 90: z: 0'`. Also some of the material parameters: instead of `offset: 0.415 0;` became `offset: x: 0.415; y: 0;`
I'll look in more detail tomorrow

vincentfretin:
Regression from fixing lint errors in #530
This closes #534

### DIFF
--- a/src/json-utils_1.1.js
+++ b/src/json-utils_1.1.js
@@ -136,7 +136,7 @@ function toPropString(propData) {
     propData.isVector3 ||
     propData.isVector2 ||
     propData.isVector4 ||
-    (propData['x'] && propData['y'])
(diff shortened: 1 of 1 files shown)


vincentfretin:
`propData['x']` returned potentially 0 so that's false
`propData.hasOwnProperty('x')` returns true

PR #552, opened 2024-06-04
Ar mode

kfarr:


kfarr:
I've been focusing my work today on determining magnetic north and have succeeded in a glitch demo for ios but cannot yet get android working as expected. The iOS demo needs some more work.

Sort of Working iOS implementation: https://glitch.com/edit/#!/compass-orientation?path=index.html
Broken android implementation: https://glitch.com/edit/#!/android-compass-orientation?path=index.html

Android notes: I was trying to implement `deviceorientationabsolute` listener which is supposed to be the android flavor, but it wasn't working.

iOS caveats:
* only works upon session start; if you go to another window, load another scene the green north arrow will be pointing in the wrong direction; workaround is to force quit the appclip and reopen
* might only works for some values of alpha / webkitcompass (ie it might only work for certain angles as you rotate around and not for others), needs some logic similar to this to handle positive and negative cases so that we constraint values between 0 and 360 for example: https://stackoverflow.com/questions/60624644/deviceorientation-compass-android

Further research:
* implementatoin from ar.js https://github.com/AR-js-org/AR.js/blob/master/three.js/src/location-based/js/device-orientation-controls.js#L133
* testing ios vs. android https://stackoverflow.com/questions/21741841/detecting-ios-android-operating-system
* deviceorientationabsolute event https://developer.mozilla.org/en-US/docs/Web/API/Window/deviceorientationabsolute_event

kfarr:
* Add an image to geomodal if the user is not logged in and not a pro user

kfarr:
Update:
* I've been focusing my work today on determining magnetic north on ANDROID and have succeeded in a glitch demo for this.

Android implementation: https://glitch.com/edit/#!/android-compass-orientation?path=index.html

This also seems to work on iOS as well!!! 

Remaining major issue:
* on both iOS and Android, when a new immersive session begins the magnetic north indication is incorrect
* only works upon session start; if you go to another window, load another scene the green north arrow will be pointing in the wrong direction; workaround is to force quit the appclip / browser and reopen

Other issues:
* this is a very clumsy implementation, could be componified, use three.js object3d instead of a-frame's setAttribute, etc.

I started to try to fix "immersive session restart loses magnetic north" issue by allowing the user to manually reset their location.

kfarr:
QRCode work:
* I tried to address QR code download issue but was not able to quickly port the QR code from bollard buddy QR maker (https://glitch.com/edit/#!/bollard-buddy-qr-maker)
* I did commit basic code to generate the URL for AR mode, but it needs a new react-friendly qr generating library
* something to try next: https://github.com/ushelp/EasyQRCodeJS

vincentfretin:
I added QR Code generation in
https://github.com/3DStreet/3dstreet/pull/540/commits/f3539d016e35daf1e808c4a5df9b35c8fa78638e and https://github.com/3DStreet/3dstreet/pull/540/commits/2ba2375d0227f9edc5080c8afc886af1a6eba31b
It shows the svg QR Code when you click on your "Create Augmented Reality QR Code" button and it downloads a bigger size in jpg format when you click on the QR code.

PR #557, opened 2024-06-04
optimization: added lazy loading for all modals

rostyslavvnahornyi:


rostyslavvnahornyi:
@kfarr you can already check, i fixed all problems with

PR #558, opened 2024-06-04
add posthog

rahulkgupta:
adding posthog tracking for session replay and analytics.

Following Option 1 in https://posthog.com/docs/libraries/js

Posthog is a session analytics and replay tool that will help us understand how users are using the product. In particular, it will help us understand what features are most valuable and where there are dropoffs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,6 +13,13 @@
     <!-- viewer controls - vr teleport -->
     <script src="https://cdn.jsdelivr.net/npm/aframe-blink-controls@0.4.3/dist/aframe-blink-controls.min.js"></script>
 
+    <!-- posthog -->
+    <script>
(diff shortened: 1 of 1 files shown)


kfarr:
> do we need to tag certain events? how does it know which events to track?

RG: there might be automated collection of pageview or other "standard" web events as well of session replay. easier to add tracking code now, see what we get, and then add custom events as needed.
VF: using self-hosted plausible to track specific custom events

> who has access

rahul + kieran


> how much does it cost

generous free tier


> what do we want to get out of it (like what report can we run?)

ideally to see why users drop off, where we can improve the initial product experience

kfarr:
only want to have this working on production

whitelist only: https://3dstreet.app
blacklist (do not track from): dev-3dstreet.web.app

PR #565, opened 2024-06-07
add posthog via import

rahulkgupta:
posthog still fails when clicking "share"


rahulkgupta:
oof looks like this makes the built file too large

vincentfretin:
You can increase the values here 
https://github.com/3DStreet/3dstreet/blob/main/webpack.prod.config.js#L10-L11

vincentfretin:
I just tested your branch, we have the same error with sessionrecording, it doesn't change anything.
The issue is in the posthog lib that does the stringify of an event that references an aframe entity.

PR #568, opened 2024-06-07
dont save entities with "autocreated" class

Algorush:


src/json-utils_1.1.js
```diff
@@ -65,20 +65,22 @@ function convertDOMElToObject(entity) {
...
-    elementTree['children'] = [];
+    const savedChildren = [];
     for (const child of children) {
       if (child.nodeType === Node.ELEMENT_NODE) {
-        elementTree['children'].push(getElementData(child));
+        const elementData = getElementData(child);
+        if (elementData) savedChildren.push(elementData);
       }
     }
+    if (savedChildren) elementTree['children'] = savedChildren;
```
vincentfretin, line 83: The condition is always true because empty list is true. You should write
`if (savedChildren.length > 0)` here.

vincentfretin (changes requested):

src/json-utils_1.1.js
Algorush: Yes, its my mistake, thanks for this!

kfarr:
Tested and worked

vincentfretin (approved):

PR #572, opened 2024-06-08
Add react-refresh to make hmr work for jsx

vincentfretin:
Add react-refresh to make hmr work for jsx and speed up development


kfarr:
Primary change:
- if you are editing a-frame core code, you will need to run `npm run dist` to see those changes

Issue #553, opened 2024-06-04
New scene title is not saved in the JSON file


Algorush:
There's always title='scene'

kfarr:
@Algorush thanks I tested and reproduced, however this is just for downloaded 3dstreet.json files and not cloud, the cloud title and rename saving seems to work as expected.

Here seems to be the offending line:
https://github.com/3DStreet/3dstreet/blob/main/src/json-utils_1.1.js#L58

Issue #532, opened 2024-05-25
Firebase error when changing scene title 


Algorush:
if I try to click on the scene title field, then after entering a new name for the street in a prompt window, an error appears:
```
Error with update title Error: FirebaseError: [code=invalid-argument]: Function doc() cannot be called with an empty path.
    at scene.js:87:11
```

kfarr:
Yes, this occurs if the file has not been saved yet. somewhere there should be the check to see if a doc exists, if not then don't try to update the title

Algorush:
I think this error appears because the sceneTitle update cannot happen without sceneId. Now the sceneId is generated only when the scene is saved to the cloud. That is, if the scene is loaded from Streetmix or from a JSON file from disk, then the sceneId is missing. And therefore, saving sceneTitle in Firebase cannot occur and this error appears.

kfarr:
closed by #619

PR #573, opened 2024-06-08
save new scene title in JSON file

Algorush:
Issue: https://github.com/3DStreet/3dstreet/issues/553
Now the title is saved in JSON and also updated in Editor mode. But there is another error related to Firebase - https://github.com/3DStreet/3dstreet/issues/532. I will fix it in another branch.

### DIFF
--- a/src/json-utils_1.1.js
+++ b/src/json-utils_1.1.js
@@ -492,11 +492,6 @@ AFRAME.registerComponent('scene-title', {
     titleDiv.setAttribute('id', 'sceneTitle');
     titleDiv.appendChild(newContent);
     document.body.append(titleDiv);
-    // update title if we are in Editor mode now
-    if (AFRAME.INSPECTOR && AFRAME.INSPECTOR.opened) {
-      const sceneTitleEditor = document.querySelector('#scene-title p');
-      sceneTitleEditor.textContent = titleText;
-    }
   },
   updateTitleText: function (titleText) {
     this.titleElement.textContent = titleText;
--- a/src/json-utils_1.1.js
+++ b/src/json-utils_1.1.js
@@ -56,7 +56,7 @@ function convertDOMElToObject(entity) {
   }
 
   return {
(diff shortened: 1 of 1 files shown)


vincentfretin:
Is this fixing #577 too?
That's really not a good idea to update a DOM element that is handled by React.
Can't we use an effect in SceneEditTitle to listen to the newTitle event?

Algorush:
> Is this fixing #577 too? That's really not a good idea to update a DOM element that is handled by React. Can't we use an effect in SceneEditTitle to listen to the newTitle event?

But we are loading json-utils.js before the Editor code which contains Events.
I don't understand a little how this can be implemented. Now if I import 'editor/lib/Events' into `SceneEditTitle.component`, where I assign `Events.on('newtitle' ()=> ...)`. And I also import `Events` into json-utils.js. But since json-utils is called earlier, Events is imported with an empty list of events.
Also I noticed that now when loading a saved scene with title, initially title=Undefined. But after selecting any entity, its title appears, this is not clear to me. This happens even without calling the 'newtitle' event handler

vincentfretin:
Events.js `on` and `emit` functions is used only for parts of editor between aframe world and React. The `newTitle` event emitted in the `metadata` aframe component is not using that but the `emit` function from aframe entity. That's not the same. To register a listener for that event you need to use `AFRAME.scenes[0].addEventListener('newTitle', ...)`.

If you see at one point the title when you select an entity, probably the SceneEditTitle react component is rerendering, you can check what going on with the props and the effect there.

Algorush:
> Events.js `on` and `emit` functions is used only for parts of editor between aframe world and React. The `newTitle` event emitted in the `metadata` aframe component is not using that but the `emit` function from aframe entity. That's not the same. To register a listener for that event you need to use `AFRAME.scenes[0].addEventListener('newTitle', ...)`.
> 

Thanks for this tip. I didn’t realize at first that I could use an existing event

Issue #577, opened 2024-01-08
scene title does not display when loading directly from URL in new window


kfarr:
when loading directly from URL in a new window the title does not display correctly such as ([link](https://3dstreet.app/#/scenes/77f6e654-072f-41bd-9ed9-f62ec2f705c6.json)) the title does not display correctly. Reloading the page sometimes correctly displays the title. KF guess -- the React app loads before the metadata component on the scene has been created. Instead, the correct title should display.

kfarr:
https://github.com/3DStreet/3dstreet/pull/609/

kfarr:
closed by #609

PR #609, opened 2024-06-11
add newtitle event

Algorush:
Issue: https://github.com/3DStreet/3dstreet/issues/577
But I also checked with another issue. After loading a JSON file from disk, the title is not immediately updated.
Its not fixed now. Added PR just to ask what needs to be changed in Events code


vincentfretin:
see my comment https://github.com/3DStreet/3dstreet/pull/573#issuecomment-2160183873

Algorush:
Now its fixed

Issue #519, opened 2024-05-21
Backport some changes from aframe-inspector


vincentfretin:
Latest sync with aframe-inspector was 2022-12-26. Since then I did some changes to aframe-inspector. Some changes were done in parallel to aframe-inspector and 3dstreet-editor, but some changes are not in 3dstreet's editor and may be interesting to backport here.

- [x] https://github.com/aframevr/aframe-inspector/pull/670 done in #540
- [x] https://github.com/aframevr/aframe-inspector/pull/671 done in #541
- [x] https://github.com/aframevr/aframe-inspector/pull/680 I think you have a similar changes but not exactly the same
- [x] https://github.com/aframevr/aframe-inspector/pull/706 and https://github.com/aframevr/aframe-inspector/pull/707 replace clipboard by clipboard-copy, replace classnames by clsx, replace fontawesome cdn and usage of fa- class by @fortawesome/free-solid-svg-icons and AwesomeIcon component
- [x] https://github.com/aframevr/aframe-inspector/pull/716 done in #542
- [x] https://github.com/aframevr/aframe-inspector/pull/715 done in #543

kfarr:
The highest prio ones to add are these 2
* https://github.com/aframevr/aframe-inspector/pull/685
* https://github.com/aframevr/aframe-inspector/pull/699

vincentfretin:
- https://github.com/aframevr/aframe-inspector/pull/714

kfarr:
@Algorush can you just do the following 3 or 4

- [ ] https://github.com/aframevr/aframe-inspector/pull/685
- [ ] https://github.com/aframevr/aframe-inspector/pull/699
- [ ] https://github.com/aframevr/aframe-inspector/pull/711
- [ ] https://github.com/aframevr/aframe-inspector/pull/714

Algorush:
> * Properly update the UI when adding or removing a mixin aframevr/aframe-inspector#699

Does this still need to be done?

vincentfretin:
Probably not, you have a different UI where you may not have the issue, I don't think you currently show the mixin in scenegraph, topbar or top right. The PR contains some cleanup to avoid unnecessary rerendering that would be nice to backport. I probably should do it and double check how much 3dstreet React code changed compared to aframe-inspector.

kfarr:
done!

PR #620, opened 2024-06-11
Properly update the UI when adding or removing a mixin

Algorush:
From this list: https://github.com/3DStreet/3dstreet/issues/519#issuecomment-2133712167
moved from aframe-inspector repository: https://github.com/aframevr/aframe-inspector/pull/699
Now UI is still not updating with the mixin name at the top of the right panel when you update a mixin, only after close/open right panel. But UI updating in the left panel.


vincentfretin:
Ok good.
In aframe-inspector, the entity representation is in CommonComponents, but in 3dstreet it's hidden with css and the title is really in Sidebar. I fixed it.

PR #621, opened 2024-06-11
[wip] Make add available for free users, but prevent adding geospatial and custom layers

rahulkgupta:
This is a WIP. Let me know if this makes sense generally and how to properly convey that a user should upgrade.
Screenshot 2024-06-11 at 2 00 31 PM
Screenshot 2024-06-11 at 2 00 35 PM

### DIFF
--- a/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx
+++ b/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx
@@ -17,10 +17,11 @@ import {
 } from './createLayerFunctions';
 import Events from '../../../lib/Events';
 
-const AddLayerPanel = ({ onClose, isAddLayerPanelOpen }) => {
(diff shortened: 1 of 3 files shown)


src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx
```diff
@@ -17,10 +19,11 @@ import {
 } from './createLayerFunctions';
 import Events from '../../../lib/Events';
 
-const AddLayerPanel = ({ onClose, isAddLayerPanelOpen }) => {
+const AddLayerPanel = ({ onClose, isAddLayerPanelOpen, currentUser }) => {
```
vincentfretin, line 22: Please use 
`const { currentUser } = useAuthContext();`
instead of passing it as prop.

vincentfretin (changes requested):

kfarr:
good enough!

vincentfretin (approved):

Issue #603, opened 2024-06-10
triage 3dtiles not loading when added mid-session


kfarr:
Sometimes as a user who selects google 3d tiles, they do not load.

No user visible error. Sometimes it might appear and just be taking a while to load, sometimes it never loads.

To reproduce:
- Choose lat/long
- Choose mapbox
- Then choose google3d
- the google 3d tiles do not appears (sometimes)
- Workaround: save the scene. Then reload, then the google 3d tiles appear correctly (usually)

Algorush:
Yes, I also noticed this behavior. I can study the reasons and fix it

Algorush:
the reason was that the `data-no-pause` attribute only works when navigating from Viewer->Editor. Therefore, it was necessary to add a `play` method call to the element with `loader-3dtiles` to start loading tiles if we are in the Editor

kfarr:
closed by #624

PR #624, opened 2024-06-12
fix 3dtiles loading

Algorush:
Issue: https://github.com/3DStreet/3dstreet/issues/603.
Now it loads in Editor mode. But I'll test in different cases more deeply

### DIFF
--- a/src/components/street-geo.js
+++ b/src/components/street-geo.js
@@ -112,6 +112,13 @@ AFRAME.registerComponent('street-geo', {
         copyrightEl: '#map-copyright'
       });
       google3dElement.classList.add('autocreated');
+
+      if (AFRAME.INSPECTOR && AFRAME.INSPECTOR.opened) {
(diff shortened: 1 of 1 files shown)


src/components/street-geo.js
```diff
@@ -112,6 +112,13 @@ AFRAME.registerComponent('street-geo', {
         copyrightEl: '#map-copyright'
       });
       google3dElement.classList.add('autocreated');
+
+      if (AFRAME.INSPECTOR && AFRAME.INSPECTOR.opened) {
+        // emit play event to start loading tiles in Editor mode
+        google3dElement.addEventListener('loaded', () => {
+          google3dElement.play();
+        });
```
vincentfretin, line 120: Yeah it makes sense. Just add `{once: true}` as the third param here.

Algorush:
I found another issue that sometimes happend. Will add new commit here

PR #638, opened 2024-06-14
New BoxHelper that take into account the rotation

### LINKED ISSUE #588: bounding box hover area does not match target geometry
As a user attempting to select segments of a street that has been rotated to match real world conditions, I expect the hover bounding box to match the segment area.

The hover area bounding box appears to be oriented on a central axis (axis aligned bounding box) instead of a bounding box oriented to the geometry (oriented bounding box)

image

considerations:
* check out https://threejs.org/docs/#examples/en/math/OBB
* is obb used in a-frame inspector already? three.js editor?

vincentfretin:
This fixes #588

Remaining to do:
- [x] Fix rotation of the hoverBoxFill

### DIFF
--- a/src/editor/lib/viewport.js
+++ b/src/editor/lib/viewport.js
@@ -6,6 +6,48 @@ import { initRaycaster } from './raycaster';
 import Events from './Events';
 import { sendMetric } from '../services/ga.js';
 
+const auxEuler = new THREE.Euler();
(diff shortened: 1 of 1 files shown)


vincentfretin:
You can test with a rotation on street-container
`document.getElementById('street-container').setAttribute('rotation', "0 90 0")`

PR #639, opened 2024-06-14
Fix mouse cursor when switching the move mode

### LINKED ISSUE #634: Update move action to use 5 finger hand
update from pointer to 5 hand

vincentfretin:
This fixes #634

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -194,6 +194,7 @@ Inspector.prototype = {
     Events.on('hidecursor', () => {
       this.cursor.pause();
       this.selectEntity(null);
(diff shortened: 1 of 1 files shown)


rahulkgupta:
i dont think the mouse cursor is resetting back to the pointer when the select button is clicked

vincentfretin:
It's coming back to the pointer only if you hover on a clickable entity, if you hover on a non clickable entity it's the hand move cursor.
But there is indeed an issue that if you hover on the previous hovered entity it doesn't change to the pointer cursor.
I disable now the raycaster, this properly fixes the mouse cursor.

PR #641, opened 2024-06-14
allow saving scenes after initial save as

### LINKED ISSUE #636: Allow for saving after importing from streetmix
Right now, user can only save as.
After the save as event, the save button is not enabled but should be.

rahulkgupta:
Quick fix to allow for saving scenes. There is an edge case where if the user decides to open another community scene after saving, the save button will show clickable. However, when it's clicked, it creates a separate scene, so i think we're ok here. In order to resolve that issue, i think it would require some decent refactoring of save logic.

fixes #636


kfarr:
can't reproduce, let's wait for posthog to see what's going on

PR #657, opened 2024-06-19
posthog improvements

rahulkgupta:


kfarr:
code looks ok to me @rahulkgupta but i didn't actually test it

kfarr (approved):

rahulkgupta:
tested it, seeing events show up in posthog for localhost. can i merge?

src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -101,6 +103,7 @@ export default class SceneGraph extends React.Component {
         if (this.props.onChange) {
           this.props.onChange(entity);
         }
+        posthog.capture('entity_selected', { entity: entity.tagName });
```
vincentfretin, line 106: Do you want to make a distinction between selecting an entity in scene graph and in the scene (entityselect is also emitted there)?
If you don't care and you want to also track when you select in the scene, you should add in `Main.js` the following:
```js
Events.on('entityselect', (entity) => {
  posthog.capture('entity_selected', { entity: entity.tagName });
})
```

but `entity.tagName` won't me much of interest here, it will always be `A-ENTITY`, you probably want `entity.getAttribute('mixin')` to know the type of asset, so
```js
  posthog.capture('entity_selected', { mixin: entity.getAttribute('mixin') });
```

src/editor/api/auth.js
```diff
@@ -1,18 +1,22 @@
...
+      name: user.displayName
+    });
     // first signIn to ga
-    if (creationTime !== lastSignInTime) return;
+    if (user.metadata.creationTime !== user.metadata.lastSignInTime) return;
     sendMetric('Auth', 'newAccountCreation');
+    posthog.capture('user_signed_up', {
+      email: user.email,
+      name: user.displayName
+    });
```
vincentfretin, line 19: If you do business in EU, you will need to have a proper Privacy Policy document that says you send email and the fullname of the user to a third party service.

src/editor/components/scenegraph/SceneGraph.js
rahulkgupta: i want to record when entities are selected in the scene graph for now. Maybe down the line it will be interesting to see the other one.

src/editor/api/auth.js
kfarr: https://github.com/3DStreet/3dstreet/issues/664

Issue #613, opened 2024-06-11
add image, splat, intersection layers


kfarr:
The existing image, splat, intersection components should be available in the add layer panel

- [x] (1) Geo - street-geo maptype: google3d - (MBES / TREAT)
- [x] (2) Additional Streetmix Street - to have multiple sections (MBES)
- [ ] (3) Image - to see my work (TREAT)
- [ ] (4) Splats - (TREAT) street-splat URL specified
- [x] (5) Gltf - to add custom objects (MBES)
- [x] (6) SVG - to add custom geometry to my scene (MBES)
- [x] (7) 3DStreet models. I will describe the questions for discussion in a separate section below
- [x] (8) QR anchor (street-qr-anchor)
- [ ] (9) Intersection - to allow for automatic intersection
- [x] (10) MapBox - (MBES) street-geo maptype: mapbox2d

Algorush:
Does Image mean screenshot or something else?

Algorush:
I'm making an auto-intersection component now from this issue: https://github.com/3DStreet/3dstreet/issues/438. I'll write my thoughts there.

kfarr:
> Does Image mean screenshot or something else?

image is a plane that accepts a material src path to jpg or png file

example in this scene with image 1914 sanborn fire insurance layer

https://dev-3dstreet.web.app/#/scenes/6a342577-e482-4d90-81b5-44ae3645a1ff.json



![Image](https://github.com/3DStreet/3dstreet/assets/470477/45d204b9-93dd-4829-969d-00705951ca03)

Algorush:
About splat layer. Services that generate splats, such as [poly.cam](poly.cam), provide them in a variety of formats, including `gltf`.
Then, in fact, we already have the opportunity to add splat to the scene through the `gltf` import function on Pro Layers.
My question is what capabilities should Splat Layers provide and look like?

kfarr:
Hey @Algorush a few updates

> I'm making an auto-intersection component now from this issue: https://github.com/3DStreet/3dstreet/issues/438. I'll write my thoughts there.

To keep things simple, for this first ticket #613 let's just have a default intersection placed without any input field; then we can support dynamic editing of intersections, and then after that we can worry about automation

> My question is what capabilities should Splat Layers provide and look like? Also what's mean "street-splat URL specified"?

To keep it simple, it should just accept a splat file like line 59 here:
https://glitch.com/edit/#!/maps-and-splats?path=index.html%3A59%3A7

[...]

PR #672, opened 2024-06-25
Add intersection layer

Algorush:
added intersection Layer with `inersection` component (not automatic version).
As a part of issue: https://github.com/3DStreet/3dstreet/issues/613


src/components/intersection.js
```diff
@@ -0,0 +1,353 @@
...
+      rotationVec
+    }) {
+      const sd = document.createElement('a-entity');
+      const repeatCountInter = [];
+      repeatCountInter[0] = width / 2;
+      // every 2 meters repeat sidewalk texture
+      repeatCountInter[1] = parseInt(length / 2);
+
+      sd.setAttribute('geometry', 'primitive', 'box');
+      sd.setAttribute('geometry', 'height: 0.4');
```
vincentfretin, line 79: I know you just moved the code in a separate file, but I'm reading this code for the first time, so making suggestions here.
This should be done in one call.
```js
sd.setAttribute('geometry', 'primitive: box; height: 0.4');
```

src/components/intersection.js
```diff
@@ -0,0 +1,353 @@
...
+      sd.setAttribute('position', positionVec);
+      sd.setAttribute('scale', scaleVec);
+      sd.setAttribute('geometry', 'depth', length);
+      sd.setAttribute('geometry', 'width', width);
```
vincentfretin, line 83:
```js
      sd.setAttribute('geometry', {depth: length, width: width});
```

src/components/intersection.js
```diff
@@ -0,0 +1,353 @@
...
+    );
+
+    function createSidewalkElem({
+      length,
+      width,
+      positionVec,
+      scaleVec = { x: 1, y: 1, z: 1 },
+      rotationVec
+    }) {
+      const sd = document.createElement('a-entity');
```
vincentfretin, line 72: Don't we want to add the autocreated class on all those created entities so those are not exported?

src/editor/components/components/AddLayerPanel/createLayerFunctions.js
```diff
@@ -149,11 +149,22 @@ function createPrimitiveGeometry() {
   Events.emit('entitycreated', newEl);
 }
 
+function createIntersection() {
```
vincentfretin, line 152: Do you know you can just use `export function createIntersection` here and for all the functions you want to export instead of having an export { ... } at the end of the file?
You don't need to modify that in this PR, I'm just asking if it was a stylistic choice in the code base or if there was any other reason I'm not aware of.

src/editor/components/components/AddLayerPanel/layersData.js
```diff
@@ -0,0 +1,85 @@
...
+  createMapbox,
+  createStreetmixStreet,
+  create3DTiles,
+  createCustomModel,
+  createPrimitiveGeometry,
+  createIntersection
+} from './createLayerFunctions';
+
+// data for PRO layers cards
+const layersData = [
```
vincentfretin, line 12: same here you can just use `export const`

src/components/intersection.js
Algorush: yes I know about this, thanks for reminding

src/components/intersection.js
Algorush: of course

vincentfretin:
Ok, you can create the PR in draft or convert it to draft now if you're still working on it, and put it on review when it's ready to review.

src/editor/components/components/AddLayerPanel/createLayerFunctions.js
Algorush: I stuck to the style that already is. Initially, I don’t know what reasons the programmer had to use this style:)

src/editor/components/components/AddLayerPanel/createLayerFunctions.js
```diff
@@ -149,11 +149,22 @@ function createPrimitiveGeometry() {
...
+  const newEl = document.createElement('a-entity');
+  newEl.setAttribute('intersection', '');
+  newEl.setAttribute('data-layer-name', 'Street intersection');
+  const parentEl = document.querySelector('#street-container');
+  parentEl.appendChild(newEl);
+  // update sceneGraph
+  Events.emit('entitycreated', newEl);
+}
+
```
vincentfretin, line 161: To properly implement undo/redo, we need to stop using such code to create an entity. We should use the proper api 
```js
Events.emit('entitycreate', {
  'data-layer-name': 'Street intersection',
  components: {intersection: ''}
});
```
but we need the changes for createEntity from
https://github.com/3DStreet/3dstreet/pull/537/files#diff-2f5a94d2538a12f206cf431af70f02620687da9738381f7be44feeaf809bc5b3R618
We can do that in another PR and modify all the entity creation here.

Algorush: I used `emit('entitycrated')` here to update the scene graph after adding a new element, not to create the entity. But I know that now there is an `updatescenegraph` event, so I'll change to it.

vincentfretin: No no don't change it. entitycreated here is better than updatescenegraph. But we'll need to rewrite all those create functions later that's what I'm saying.

Algorush: Ok, got it

Algorush:
I'm closing this PR because slpat PR (https://github.com/3DStreet/3dstreet/pull/684) is based on this PR.

Issue #665, opened 2024-06-21
Add sign up prompt when non-authed user comes in from streetmix


kfarr:
Possible methods to trigger this dialog:
1) We show this modal IF the # sign includes https://streetmix.net/... URL *AND* the user is not authenticated
2) We show the modal ONLY after successful load event from streetmix

PR #674, opened 2024-06-25
show a popup for when a streetmix scene is loaded for the first time

rahulkgupta:
Screenshot 2024-06-25 at 9 07 11 AM

Very simple change for now. Only happens when the page is loaded. Won't happen if the user clicks "load from streetmix" which i think is ok for now.

### DIFF
--- a/src/editor/components/Main.js
+++ b/src/editor/components/Main.js
@@ -84,6 +84,14 @@ export default class Main extends Component {
       '.viewer-logo-start-editor-button'
     );
     htmlEditorButton && htmlEditorButton.remove();
+
+    const isStreetMix = window.location.href.includes('streetmix');
+    if (isStreetMix) {
+      STREET.notify.warningMessage(
+        'Hit save if you want to save changes to the scene. Otherwise changes will be lost'
+      );
+    }
+
     Events.on(
       'opentexturesmodal',
       function (selectedTexture, textureOnClose) {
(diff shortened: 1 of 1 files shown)


src/editor/components/Main.js
```diff
@@ -84,6 +84,14 @@ export default class Main extends Component {
       '.viewer-logo-start-editor-button'
     );
     htmlEditorButton && htmlEditorButton.remove();
+
+    const isStreetMix = window.location.href.includes('streetmix');
```
vincentfretin, line 88: You can be even more specific here. `window.location.hash` but yeah that's the same.

vincentfretin (approved):

src/editor/components/Main.js
vincentfretin: If you want the message also when you click on "Load from Streetmix", you can use this:
```js
    function locationHashChanged() {
      const isStreetMix = window.location.hash.includes('streetmix');
      if (isStreetMix) {
        STREET.notify.warningMessage(
          'Hit save if you want to save changes to the scene. Otherwise changes will be lost'
        );
      }
    }

    locationHashChanged();
    window.addEventListener('hashchange', locationHashChanged);
```

rahulkgupta: good idea, thanks!

rahulkgupta (commented):
responded to comments

vincentfretin (approved):

rahulkgupta:
#665

PR #676, opened 2024-06-27
Undo/redo position/rotation/scale

### LINKED ISSUE #590: undo / redo for position/rotation/scale only as v1
As a user I wish to undo past actions and redo those once undone. It's ok to have a limit of undo actions such as 20 deep.

It's also ok to start with just 1 action, such as position/rotation/scale

vincentfretin:
This closes #590
Remaining to do:
- [x] shortcut ctrl+z and ctrl+shift+z
- [x] it should truncate to 3 decimals (instead of 5) when moving with transformcontrols
- [x] fix issue that we got another command when opening the right panel that truncate to 3 decimals
- [ ] use the entityupdatecommand for mixin

### DIFF
--- a/src/editor/components/components/CommonComponents.js
+++ b/src/editor/components/components/CommonComponents.js
@@ -6,7 +6,6 @@ import PropertyRow from './PropertyRow';
 import Collapsible from '../Collapsible';
 import classnames from 'classnames';
 import {
-  updateEntity,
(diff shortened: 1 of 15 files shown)


kfarr:
looks great!

vincentfretin:
Additional fixes were made in #708

Issue #632, opened 2024-06-14
ability to have image previews for models


kfarr:
* related to #576 and previous work from @vincentfretin and myself on a model json format
* what is the best way to provide for image previews for models on the add layer panel?

Algorush:
We can try to generate model images automatically. Write a function that will take images of all models. If all models are already loaded into page memory, then this should not take up many resources?? But I'm not sure if this is a good idea. 
Maybe It might be better to have such a generator as a separate script, which will create images of models and place them in a separate folder. The downside here is that we will have to load all images of models onto the page from the server

kfarr:
@Algorush was already have a script that can generate jpegs from gltf, see https://github.com/3DStreet/citybuilderjr

this task is the ability for the add entity panel to reference some sort of json like catalog.json for the image file. How we create the image file is a task for another ticket (or to be done manually)

Algorush:
> this task is the ability for the add entity panel to reference some sort of json like catalog.json for the image file.

Ok, got it.
Now there is a file `cardsData.js`, where img, icon, name for each mixinId can be stored. It is better to rename it to `mixinsData.js`, because there will be only mixins data now. Use this data in the `getGroupedMixinOptions` function in `addLayerPanel`: https://github.com/3DStreet/3dstreet/blob/95b81ab695050f92be0884dd48ecfb0d862a4d0b/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx#L52-L58
We can also remove the grouping logic from this function and instead divide the mixins into groups inside mixinsData.js. Like in src/assets.js.

kfarr:
@Algorush I tried adding an image to `Bicycle_1` mixin in this branch: https://github.com/3DStreet/3dstreet/pull/683 but I was not able to make it work. Can you check it out and suggest what I might be doing wrong?

Algorush:
> @Algorush I tried adding an image to `Bicycle_1` mixin in this branch: #683 but I was not able to make it work. Can you check it out and suggest what I might be doing wrong?

also need to import `mixinsData` from `cardsData.js` file in `AddLayerPanel.component.js` and add `img: mixinsData[mixinId]` here:
https://github.com/3DStreet/3dstreet/blob/95b81ab695050f92be0884dd48ecfb0d862a4d0b/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx#L52-L58
same for description

kfarr:
[...]

PR #683, opened 2024-06-28
use catalog.json for additional asset metadata

kfarr:
Issue: https://github.com/3DStreet/3dstreet/issues/632


src/catalog.json
Algorush, line 1: Should we use exactly this structure? Using a structure with mixin as objects id will be more convenient, since in the component mixins are already divided into groups by mixinId, and obtaining their data (name, img, descr) will take O(1): 
{
  mixinId1: { 
        "id": "bicycle_1",
        "name": "Bicycle Model 1",
        "img": "ui_assets/cards/models/bicycle_1.jpg",
        "description": "Bicycle, it's a bike!"
 }, 
  mixinId2: { 
        "id": "bus",
        "name": "Bus New Flyer XD40",
        "description": "The New Flyer XD40 is a modern, stylish, and comfortable bus that is perfect for people in urban settings."
} 
}

PR #684, opened 2024-06-29
Add intersection and splat layers

Algorush:
This branch is based on the `add-intersection-layer` branch (https://github.com/3DStreet/3dstreet/pull/672).
Suggestion from Kieran (https://github.com/3DStreet/aframe-gaussian-splatting/issues/2): In order for the splat object to be displayed, you need to set: `renderer='logarithmicDepthBuffer: false;'`


Algorush:
Save/load doesn't work, so I converted to draft

Algorush:
Now save/load for splat models works. So far I have loaded the `aframe-gaussian-splatting-component.min.js` from index.js, since it only takes up 20kb. And removed dynamic loading from `createLayerFunctions.js`

Issue #686, opened 2024-07-02
Error in in Right Panel after making New scene


Algorush:
Found this error after creating New scene by clicking "New scene" button.
To reproduce: 
- click on "New scene"
- click on default-street
- open right panel
- error appears:
```
Cannot read properties of undefined (reading 'JSON')
TypeError: Cannot read properties of undefined (reading 'JSON')
```

Algorush:
closed as completed: https://github.com/3DStreet/3dstreet/pull/689

Algorush:
closed as completed: https://github.com/3DStreet/3dstreet/pull/689

PR #687, opened 2024-07-02
Fix new street error

Algorush:
To fix this error: https://github.com/3DStreet/3dstreet/issues/686


Algorush:
strange commits were pulled up, closed

PR #697, opened 2024-07-11
Automatic intersection component

Algorush:
added a component to determine the size of the parameters required for the intersection component.
Its parameters specify selectors of street elements at the intersection:
```
northStreet: selector,
southStreet: selector, ...
```

Example of work: 
`testEl.setAttribute('auto-intersection', 'southStreet: #left_street; westStreet: #perpendicular_street')`

By the way, I found that when adding a Streetmix street from the input prompt, it created with `id='http_path'`, such a selector will not work as it is has wrong format


Algorush:
I need to fix some issue

Algorush:
Example of how it works now. 
Initial scene with 3 streets:
image
Add empty `a-entity#test` element and this line:
`test.setAttribute('auto-intersection', "northStreet: #north_street; westStreet: #perpendicular_street; southStreet: #south_street")`
Streets take their places automatically, but they must be rotated correctly first. Here is result:
image

Algorush:
But save/load doesn't work yet. I'm getting maximum call stack error, I don't remember why it could be, need to debug

Issue #706, opened 2024-07-12
Camera button


rahulkgupta:

kfarr:
here is an example of an older ui that we had on the right-side and the camera button would just download an image immediately https://3dstreet.github.io/aframe-loader-3dtiles-component/examples/google-tiles-3dstreet/

PR #707, opened 2024-07-12
Screenshot button

rahulkgupta:
#706 

Screenshot 2024-07-12 at 2 38 06 PM

clicking the image button takes a PNG.

Defaulting to PNG (dont see the need for JPG, but lmk).

Question: What should we do w/ the existing share functionality?
- thinking we remove the PNG and JPG options and add a notify or we just add a notify and remove it in a couple weeks?

### DIFF
--- a/src/editor/components/components/ActionBar/ActionBar.component.jsx
+++ b/src/editor/components/components/ActionBar/ActionBar.component.jsx
@@ -1,7 +1,8 @@
 import {
   faHand,
   faHandPointer,
-  faPlusSquare
+  faPlusSquare,
+  faImage
 } from '@fortawesome/free-regular-svg-icons';
 import { AwesomeIcon } from '../AwesomeIcon';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


src/editor/components/components/ActionBar/ActionBar.component.jsx
```diff
@@ -28,6 +29,19 @@ const ActionBar = ({ handleAddClick, isAddLayerPanelOpen }) => {
     setCursorEnabled(true);
   };
 
+  const handleCameraCLick = () => {
+    Events.emit('camera_clicked');
```
vincentfretin, line 33: Don't emit an event if you don't use it.

src/editor/components/components/ActionBar/ActionBar.component.jsx
```diff
@@ -28,6 +29,19 @@ const ActionBar = ({ handleAddClick, isAddLayerPanelOpen }) => {
...
+    const screenshotEl = document.getElementById('screenshot');
+    screenshotEl.play();
+
+    posthog.capture('screenshot_taken', {
+      scene_id: STREET.utils.getCurrentSceneId()
+    });
+
+    screenshotEl.setAttribute('screentock', 'type', 'png');
```
vincentfretin, line 41: I think jpg would be a better default, png can really be heavy in size.

src/editor/components/components/ActionBar/ActionBar.component.jsx
```diff
@@ -28,6 +29,19 @@ const ActionBar = ({ handleAddClick, isAddLayerPanelOpen }) => {
     setCursorEnabled(true);
   };
 
+  const handleCameraCLick = () => {
+    Events.emit('camera_clicked');
+    const screenshotEl = document.getElementById('screenshot');
+    screenshotEl.play();
```
vincentfretin, line 35: Is this really needed? I don't see any play function or events in the screentock component.

src/editor/components/components/ActionBar/ActionBar.component.jsx
rahulkgupta: png seems to be the better default for graphics design / presentation purposes. it also seems to be used a bit more than jpg in our usage.

src/editor/components/components/ActionBar/ActionBar.component.jsx
rahulkgupta: i just copy pasted what was in the other function!

PR #720, opened 2024-07-19
get geoid as javascript function

kfarr:
Good news -- we now have what appears to be a working method to reliably find the ground point of google 3d tiles for a 3DStreet scene.

However, how can we offer this updated feature without breaking existing user's geospatial scenes?

How we store elevation today:
- "elevation" value is stored in street-geo component
- this value is fetched from google elevation service which returns topographical elevation
- street-geo adds an arbitrary hack constant (32.49158) to convert topographical height to ellipsoidal height. Using a constant value is incorrect (outside of the SF Bay Area). Instead, `ellipsoidalHeight` value can be calculated only with geoid height which varies based on user-specified lat/lon.

change:
- now we have the a more accurate `ellipsoidalHeight` from our new cloud function (yay!), but how should we handle this new feature without breaking old stored user values?

proposed way to handle to be backwards compatible:
- add new property to `street-geo` component `ellipsoidalHeight` 
- if `ellipsoidalHeight` exists, then use that value for google3d tiles without the hack constant
- if it doesn't, then use old school `elevation` value with the existing hack constant

how to do this?
- add 2 conditional statements in street-geo `init` and `update` methods:
```
    // if data.ellipsoidalHeight, use it, otherwise use data.elevation less constant (deprecated)
    const height = data.ellipsoidalHeight
      ? data.ellipsoidalHeight
      : data.elevation - this.elevationHeightConstant;
```

test scenarios:
- [x] user makes a new geospatial scene which automatically specifies `ellipsoidalHeight` for new locations; elevation may be null, NaN, 0, we don't care; correct ground location should "just work" with max adjustment of ~2 meters needed
- [x] user opens a existing geospatial scene -- the elevation value is technically incorrect but the user has manually adjusted their scene to match the naive constant method; what do we want to happen? we want existing behavior
- [x] what happens when a user edits the location of an existing geospatial scene? we want it to use the new, correct value.
- [x] what happens if opens an old geospatial scene, makes edits but NOT to the geospatial settings? the user expects the 3dtiles "to stay where they are" so let's make sure they do!
- [ ] what happens if an `ellipsoidalHeight` value is actually 0 -- will javascript assume that the value is falsy and use the incorrect (or nonexistent) `elevation` value?

other issues need to fix:
* does not allow for negative numbers for ellipsoidal value


kfarr:
although this worked in local emulator, fun new error from firebase when trying to deploy:
`Error: [getGeoidHeight(us-central1)] Changing from an HTTPS function to a callable function is not allowed. Please delete your function and create a new one instead.`
- does this mean i should rename it, or "delete" from a gui?

src/components/street-geo.js
```diff
@@ -146,10 +156,15 @@ AFRAME.registerComponent('street-geo', {
   },
   google3dUpdate: function () {
     const data = this.data;
+    // if data.ellipsoidalHeight, use it, otherwise use data.elevation less constant (deprecated)
+    const height = data.ellipsoidalHeight
+      ? data.ellipsoidalHeight
+      : data.elevation - this.elevationHeightConstant;
+
```
vincentfretin, line 163: You can set this.height in init so you don't have the calculation logic twice in this component.

src/editor/components/components/GeoPanel/GeoPanel.component.jsx
```diff
@@ -37,7 +37,6 @@ const GeoPanel = () => {
           <a onClick={onClick}>Click to set location</a>
         )}
       </>
-      )
```
vincentfretin, line 40: Ah you already removed the parenthesis, I did it too in a separate PR #725

src/editor/components/modals/GeoModal/GeoModal.component.jsx
```diff
@@ -38,31 +40,33 @@ const GeoModal = ({ isOpen, onClose }) => {
...
-          setElevation(elevation);
+        if (!isNaN(ellipsoidalHeight)) {
+          setElevation(ellipsoidalHeight);
         }
       }
     }
   }, [isOpen]);
 
-  const requestAndSetElevation = (lat, lng) => {
+  const requestAndSetElevation = async (lat, lng) => {
```
vincentfretin, line 55: If you don't use await in the function, then it's not async, even if your use then/catch.

public/functions/package.json
```diff
@@ -14,14 +14,16 @@
   },
   "main": "index.js",
   "dependencies": {
+    "@googlemaps/google-maps-services-js": "^3.4.0",
     "firebase-admin": "^12.1.1",
     "firebase-functions": "^5.0.1",
-    "stripe": "^8.4.0"
+    "stripe": "^8.4.0",
+    "core-js": "^3.6.5"
```
vincentfretin, line 21: Was it necessary to specify core-js dependency here?

src/components/street-geo.js
```diff
@@ -7,9 +7,12 @@ const MAPBOX_ACCESS_TOKEN_VALUE =
 
 AFRAME.registerComponent('street-geo', {
   schema: {
-    longitude: { type: 'number', default: 0 },
-    latitude: { type: 'number', default: 0 },
-    elevation: { type: 'number', default: 0 },
+    longitude: { type: 'number', default: null },
+    latitude: { type: 'number', default: null },
```
vincentfretin, line 11: Is it okay to change default to null for longitude and latitude now?

public/functions/geoid.js
```diff
@@ -0,0 +1,292 @@
...
+            }
+
+            return model.offset + model.scale * h;
+        };
+    });
+};
+
+// function that takes in lat, long, and returns the height
+async function getGeoidHeightFromPGM(geoidFilePath, lat, long) {
+    const geoidModel = await loadGeoidModel(geoidFilePath);
```
rahulkgupta, line 288: are we loading this every invocation of this function?

rahulkgupta:
I'm having trouble following the description. For user's stored values, can't we just use that and not rely on the service? 
> "elevation" value is stored in street-geo component
this value is fetched from google elevation service which returns topographical elevation

What does this mean? Is the value stored in the scene data in the firebase db or is it fetched via google elevation?

src/components/street-geo.js
kfarr: this.height may change, it's used in `google3dCreate` and `google3dUpdate`, so the only other option I could see was to create a small function wrapper but if it's only 2 times then maybe not worth it

public/functions/package.json
kfarr: bug in google maps node api :/

src/components/street-geo.js
kfarr: I don't know. It's probably safest to keep 0

public/functions/geoid.js
kfarr: yes, however it's a "local file" in the google cloud function so it's not downloading it, it is in the same local directory. i honestly don't know how to keep it in memory and reference it, instead of reloading it upon each invocation, and this seems to work in <1 second each time, so I'll just cross my fingers and praise gcp cloud functions that it works

src/editor/components/modals/GeoModal/GeoModal.component.jsx
kfarr: thanks i'll remove it

PR #724, opened 2024-07-23
fire checkout success event only when payment is successful

rahulkgupta:


src/editor/components/modals/PaymentModal/PaymentModal.component.jsx
```diff
@@ -23,7 +23,7 @@ const PaymentModal = ({ isOpen, onClose }) => {
   const { currentUser } = useAuthContext();
   const [isLoading, setIsLoading] = useState(false);
 
-  if (currentUser?.isPro) {
+  if (location.pathname.includes('success')) {
```
vincentfretin, line 26: You want to check for success in `/#/modal/payment/success` url right?
That won't work. All the part after the # is not part of location.pathname but part of location.hash
You want location.href or location.hash here.

rahulkgupta:
Responded to comments, thanks

vincentfretin (approved):

PR #735, opened 2024-07-29
Use document.readyState !== 'complete' to be sure startObserving is always executed

vincentfretin:
Following #734
`readyState` could also be in the `interactive` state before the `DOMContentLoaded` event is fired.
`if (document.readyState !== 'complete') {`
would be more adequate here.

https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState


kfarr:
agreed this make sense to me thanks @vincentfretin !

Issue #730, opened 2024-07-26
user specified rotation is not respected upon reload


kfarr:
This is urgent and on production reproducible with a new scene. This also occurs locally with latest `main` build.
***

As a user I have set a position for my streetmix sourced street placed in my scene. Upon saving the position and rotation appear to be saved but in fact when I reload my scene from the cloud rotation of my street is not saved correctly.

Upon inspecting the raw scene json I notice that there is no position component. (This issue occurs both with download raw 3dstreet.json file as well as in the cloud using firefoo inspector)

In this scene for example:
https://3dstreet.app/#/scenes/a49cf5d8-c43b-4a1c-85a9-a9771cd70a8e.json

The json for the streetmix street doesn't include a rotation component:
```
        "components": {
          "position": "-31.008 3.278 -2.078",
          "street": {
            "left": "grass",
            "length": "120",
            "right": "grass",
            "synchronize": "false"
          },
          "streetmix-loader": "streetmixStreetURL: https://streetmix.net/pav.rysula/1/sitna; streetmixAPIURL: https://streetmix.net/api/v1/streets?namespacedId=1&creatorId=pav.rysula; name: Sítná; synchronize: false",
          "visible": ""
        },
```

also cannot modify rotation or scale for a streetmix scene in the right-hand panel. Manually editing rotation or scale values results in no change of the object.
![image](https://github.com/user-attachments/assets/26bd9701-6ae6-4bca-89cf-6255d90f9025)

kfarr:
confirmed working as expected in 6916af8426ab7f26cc1581b0ac3d9040fa942449

now going to step back through and see which commit caused this...

kfarr:
it appears that this commit is the culprit:
https://github.com/3DStreet/3dstreet/commit/983c365a4b61534e1b408714fa2a424c28fab507

kfarr:
deployed and resolved

vincentfretin:
I'm sorry I didn't see the issue when I did the change.
Reverting this specific commit was indeed working because I also added the else branch for unknown properties in https://github.com/3DStreet/3dstreet/commit/9b43fbb3902d7712b60e6bdade37a0f4bb65b537
That's ok for number and string properties, but the stringification behavior for vec2 and vec3 properties is not correct, it produces `{x: 0.78, y: 1}` instead of `0.78 1`.
#736 would be a better fix for the issue.

kfarr:
excellent, thanks for digging into this! I'll review 736

PR #736, opened 2024-07-29
Properly stringify some properties of dynamic schema, fallback to static schema

vincentfretin:
Properly stringify some properties of dynamic schema, but be sure sure to fallback to static schema of the component if not currently on entity.

This is the proper fix for #730


kfarr:
tested and nothing seems to be broken

PR #738, opened 2024-07-29
change to map selector, reduce google3d tiles distanceScale

kfarr:
now only 1 map can be selected at a time, which is intended user behavior anyway

### DIFF
--- a/src/components/street-geo.js
+++ b/src/components/street-geo.js
@@ -13,7 +13,11 @@ AFRAME.registerComponent('street-geo', {
     orthometricHeight: { type: 'number', default: null },
     geoidHeight: { type: 'number', default: null },
     ellipsoidalHeight: { type: 'number', default: null },
-    maps: { type: 'array', default: [] }
+    maps: {
+      type: 'string',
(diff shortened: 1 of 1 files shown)


src/components/street-geo.js
```diff
@@ -13,16 +13,20 @@ AFRAME.registerComponent('street-geo', {
     orthometricHeight: { type: 'number', default: null },
     geoidHeight: { type: 'number', default: null },
     ellipsoidalHeight: { type: 'number', default: null },
-    maps: { type: 'array', default: [] }
+    maps: {
+      type: 'string',
+      default: 'google3d',
+      oneOf: ['google3d', 'mapbox2d']
+    }
```
vincentfretin, line 20: Changing the type would break existing saved json, right? Is it ok?

kfarr: @vincentfretin luckily a stringified array of 1 item appears to be identical to a string; and we didn't support choosing multiple at a time yet anyway so there shouldn't be any arrays in the db

PR #741, opened 2024-07-29
add osm3d map type

kfarr:
things to do before making this available

- [x] - better handle 3 dependencies of osm-tiles, osm-geojson, and osm2geojson, right now they're all stuffed into `osm-tiles-geojson.js` (now using webpack on this fork: https://github.com/3DStreet/osm4vr/)
- [x] buildings and maps should not be selectable / clickable
- [x] - better handle removing buildings when switching maps
- [x] - (!) do NOT save all the tiles generated by osm3d as children in 3dstreet json, instead ensure that any children of street-geo are not saved; look at example from environment component
- [x] - better handle copyright notices adding / hiding for OpenStreetMap
- [x] - test saving / loading


kfarr:
wip: 
image

kfarr:
things this doesn't support:
- this uses the default tile.openstreetmap.org service. The osm4vr component does not yet support proper appending of api key in tile image requests.

PR #746, opened 2024-08-08
Clean up code in AddLayerPanel component and drag and drop

### LINKED ISSUE #576: add "drag and drop" to add new entity panel
part of "add layer" button

figma design wip:
https://www.figma.com/file/W9jid3A0jgssnIKBFEvG38/3DStreet-2022?type=design&node-id=3583-76149

example implementation and further discussion on expected behavior:
https://github.com/c-frame/aframe-editor/pull/1

Relevant quote:
> As a user hovers over a model the destination cursor is shown at the default location. If they click, it will be spawned in that location. If they click to drag the model onto the scene, the cursor previews the drop zone where the model will be placed upon click up.

vincentfretin:
This is mainly code cleanup and creating pure functions (not using variables from outside the function) to better understand this code.
I fixed some issues and implemented drag and drop.
This closes #576

- Don't recreate AddLayerPanel each time we want to add an element, and make sure getSelectedMixinCards is not recalculated unnecessarily.
- Don't create preview entity each time the panel open.
- Don't set y to 0.2 for preview on non segment (for example bicycle on waterfront :p) and properly use worldToLocal when creating the entity so the preview and created entity are at the same position.
- Placing a second element moved the existing custom-group entity instead of moving the second entity inside the existing custom-group entity.
- The preview entity wrongly used a previous position instead of scene origin when no entity was selected.
- Instead of scene origin, place element at the intersection of ground and current camera position.
- Select created entity.
- Allow to drag and drop card to create an element at the drop position

### DIFF
--- a/src/editor/components/Main.js
+++ b/src/editor/components/Main.js
@@ -341,7 +341,7 @@ export default class Main extends Component {
             <Compass32Icon />
           </Button>
         )}
-        {this.state.inspectorEnabled && this.state.isAddLayerPanelOpen && (
+        {this.state.inspectorEnabled && (
           <AddLayerPanel
             onClose={this.toggleAddLayerPanel}
             isAddLayerPanelOpen={this.state.isAddLayerPanelOpen}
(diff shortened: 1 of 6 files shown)


vincentfretin:
@kfarr Drag an drop implemented, creating the element in 3dstreet-container.
You can add your cursor and it's good to merge.

Issue #742, opened 2024-07-30
pin geo to top of scenegraph


kfarr:
as a user I always want to see the geospatial entity at the top of the layers panel (scenegraph) if one exists for the scene

eventually do that for environment as well

PR #760, opened 2024-08-13
Layers panel

rahulkgupta:
Wanted to get a WIP Draft for the side panel changes. The main difference is that i've removed a lot of logic from `SceneGraph`. I'm just using the `scene` node itself and iterating through that to render the scenegraph.


src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -62,307 +51,63 @@ export default class SceneGraph extends React.Component {
...
-    }
-    return entities.filter((entityOption) => {
-      return filterEntity(entityOption.entity, filter || this.state.filter);
-    });
-  }
-
-  isVisibleInSceneGraph = (x) => {
-    let curr = x.parentNode;
-    if (!curr) {
+  shouldRenderNode = (node) => {
```
rahulkgupta, line 64: can add to this to determine what to render in the main scene graph. Notice how we're hiding things in `HIDDEN_CLASSES` like`teleportRay`

src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -62,307 +51,63 @@ export default class SceneGraph extends React.Component {
...
-        });
-        resultEntities.push(
-          <div className={className} key={i}>
-            {layerEntities}
+      const children = node.children || [];
+      const { renderedChildren, isExpanded: childIsExpanded } =
+        this.renderSceneNodes(Array.from(children), depth + 1);
+
+      const isSelected = node === this.props.selectedEntity;
+      isExpanded = isExpanded || isSelected || childIsExpanded;
```
rahulkgupta, line 97: im figuring out whether or not the node should be expanded based on the a selected child node in here. There seems to be an issue where after a close of an expanded node and then selection on the editor, it doesn't seem to re-expand

src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -377,25 +122,10 @@ export default class SceneGraph extends React.Component {
...
 
-    const clearFilter = this.state.filter ? (
-      <a onClick={this.clearFilter} className="button fa fa-times" />
-    ) : null;
-
     return (
       <div id="scenegraph" className="scenegraph">
         <div className="scenegraph-toolbar">
           <ToolbarWrapper />
-          <div className="search">
```
rahulkgupta, line 388: this wasn't being used at the moment, so i decided to remove all the search functionality. can add it back if we decide we need it

src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -20,36 +22,23 @@ export default class SceneGraph extends React.Component {
...
       1000
     );
-    this.updateFilteredEntities = debounce(
-      this.updateFilteredEntities.bind(this),
-      500
-    );
   }
 
   componentDidMount() {
-    this.setFirstLevelEntities();
```
rahulkgupta, line 52: removed all notion of first and second level entities.

rahulkgupta:
@vincentfretin i know this is a draft, but would love your eyes on it!

vincentfretin:
Yes sure I'm looking at it right now. I wasn't available today to look at it earlier.

vincentfretin (commented):
I first reviewed the original implementation to familiarize with it and eventual shortcomings.
When an entity is created, it is selected, so state.entity is updated in Main, that triggers SceneGraph componentDidUpdate that calls Scenegraph.selectEntity (note Events.emit('entityselect', entity, true); the true here is used nowhere in the current listeners, there is no Event.on('entityselect', (entity, thisBoolean) => {}) in the code base), although the code here should expand to the created entity, in fact it doesn't. That's most probably because this.state.filteredEntities is not yet updated, it's updated in rebuildEntityOptions (debounced at 1s after the createdentity event) that calls
```
    this.setState({
      entities: entities,
      filteredEntities: this.getFilteredEntities(this.state.filter, entities)
    });
```
Even if we remove the debounce on both rebuildEntityOptions and getFilteredEntities, that still don't work because the setState is asynchronous and so this.state is not yet up to date when  Scenegraph.selectEntity execute. This logic should be put in a createEffect really, but here it's a class component so in componentDidUpdate(prevProps, prevState). And this is the right place to implement a scrollToElement.

Note: debounce is okay for getFilteredEntities when it's used in the search input, but here when it's called by rebuildEntityOptions it should bypass the debounce and do it right away.
We should remove the debounce on rebuildEntityOptions, be sure that entityupdate of position/rotation/scale is ignored to not trigger rebuildEntityOptions.

In the original code, the expanded logic is in SceneGraph component, not in Entity component mainly because of the search input, that way when you clear the filter, you have back the expanded branches.

In your new implementation, you break accessibility with arrow keys, expanding/collapse with right/left arrow keys, and selecting previous/next with up and down keys. You didn't try to fix expanding the tree to the created entity, and fixing the issue with the state being in the Entity component would be even harder now I think.
The new implementation now expand on click on the line, not just the expand icon, not sure that's expected. Collapsing a parent and expand it back, you lose the state of expanded children.

I think it would be better to restart from the original aframe-inspector SceneGraph code that didn't have the first and second levels entities stuff, add back your HIDDEN_CLASSES/shouldRenderNode change, getEntityType for posthog, removing the filter input and associated code, and fix the expand on entitycreated issue.

vincentfretin:
If that's okay, I can take a stab at updating the code like I said above tomorrow, before our meeting.
Main focus I think is to not break what was working with the keys, resolve the expand and scroll to the created entity, and hide elements we don't want.

vincentfretin:
I finished with the changes, please do some tests to see if all is working for you.

vincentfretin:
This closes 
- #659 
- #742

PR #762, opened 2024-08-14
Geo and environment on top

rahulkgupta:
Screenshot 2024-08-14 at 10 37 17 AM


kfarr:
@rahulkgupta looks good to me. should we also put User Layers as fixed in 3rd spot now or wait till later?

rahulkgupta:
o that's a good idea. can do

vincentfretin (approved):
Looks good

Issue #757, opened 2024-08-12
drag and drop pro layer doesn't work on production


kfarr:
I think it's because function names were clobbered during minimization so they cannot be fired after the drag event by using the function name as string

console error is:
`AddLayerPanel.component.jsx:412 Function $V not found
onDrop	@	`

So ... maybe there is a way to mark those functions to not be minimized?

vincentfretin:
If we add a new function, we need to make sure to add it in the webpack config as well, it's a bit brittle.
I'm wondering if there is another way to modify the code to avoid having a name mismatch when it's minified.

kfarr:
I think we could add a key value to the function name that won't be
minified? Or a lookup table with a string lookup key that won't be minified?

PR #771, opened 2024-08-17
Alternate fix for drag and drop pro layer in production

vincentfretin:
Alternate fix for #757 to avoid mangling name issue with minified build, get the card from the id to access handlerFunction.


kfarr:
thanks @vincentfretin this is a much better solution! noting for myself that most of the create layer functions don't support position yet, but i can add that later

PR #775, opened 2024-08-18
create new "Streets" group in add entity panel

kfarr:
also update the other layer groupings to make sure we have all models in `assets.js` accessible to users


kfarr:
@vincentfretin i could use your help with this too -- I added a new add entity panel dropdown group "Streets Layers" which has at the moment 1 card. When I click on the card it works as expected to run `createStreetmixStreet`. However when I drag and drop this card instead it appears to run `createMapbox`

Excerpt from `/src/editor/components/components/AddLayerPanel/layersData.js`
```
const streetLayersData = [
  {
    name: 'Street from Streetmix URL',
    img: 'ui_assets/cards/streetmix.jpg',
    icon: 'ui_assets/cards/icons/streetmix24.png',
    requiresPro: false,
    description:
      'Create an additional Streetmix street in your 3DStreet scene without replacing any existing streets.',
    id: 1,
    handlerFunction: createStreetmixStreet
  }
];

// data for PRO layers cards
const layersData = [
  {
    name: 'Mapbox 2D Aerial',
    img: 'ui_assets/cards/mapbox2d.jpg',
    icon: 'ui_assets/cards/icons/mapbox24.png',
    requiresPro: true,
    description:
      'Create entity with mapbox component, that accepts a long / lat and renders a plane with dimensions that (should be) at a correct scale.',
    id: 1,
    handlerFunction: createMapbox
  },
  ```

**Comment by vincentfretin** [comment]
Be sure to increment the id, it should be unique. I see in your code snippet there are both "id: 1". We use the id now when we drop to call the function since #771

**Comment by vincentfretin** [comment]
I misread your comment. In #771 I hard coded layersData in the drop function so it calls createMapbox, and not getting the id from streetLayersData, we should really get the cards from selectedCards somehow.

File: src/editor/components/components/AddLayerPanel/LayersOptions.js [hunk 1]
```diff
@@ -1,50 +1,50 @@
 const LayersOptions = [
   {
-    value: 'Models: Personal Vehicles',
-    label: 'Models: Personal Vehicles',
-    mixinGroups: ['vehicles', 'vehicles-rigged'],
-    onClick: () => console.log('Models: Personal Vehicles')
+    value: 'Street Layers',
+    label: 'ꈨ Street Layers',
```
vincentfretin, line 4: I just see an empty square or a square with ? on Ubuntu 22.04 Chrome for this one.

PR #780, opened 2024-08-18
Properly unregister listeners when React component unmount

vincentfretin:
This contains some additional cleanup I did in https://github.com/aframevr/aframe-inspector/pull/745 and backport the unregister listeners on unmount I did in https://github.com/aframevr/aframe-inspector/pull/748


kfarr:
don't know how to test but the basics seem to still work for me

Issue #298, opened 2023-06-14
reduce image size by 1MB


kfarr:
the top 6 images take 5MB! Can we reduce this to 4MB or lower?

* Recompress at higher compression ratio and / or resize
* Test in app to see experience
* Consider VR distances in app test
* Consider future "high quality texture" toggle as a concession for lowering quality for now to optimize load time

![image](https://github.com/3DStreet/3dstreet/assets/470477/b1a6f55f-8a3e-42f2-a8b8-9c7affd874f3)

vincentfretin:
Use https://squoosh.app and save to webp.

PR #786, opened 2024-08-22
Add intro video

rahulkgupta:
Screenshot 2024-08-22 at 1 42 36 PM


vincentfretin:
3dstreet-editor.js went from 1,6M to 2,5M with the mux video player. We should probably look into lazy loading the modals again soon #557

kfarr:
@vincentfretin agreed that would be a good thing. in addition, to lazy loading, we can reduce size with:
* removing legacy mapbox component in favor of new osm: https://github.com/3DStreet/3dstreet/issues/787 and 
* more aggressive texture compression https://github.com/3DStreet/3dstreet/issues/298

Issue #770, opened 2024-08-17
default google 3d tiles not shown in scenegraph after setting location


kfarr:
after adding 3dtiles, somehow we needs to fire another entitychanged or whatever scenegraph update event needed

the tiles are added how? i think in the modal maybe after the update button is pressed?

vincentfretin:
I'm not sure if this is this one, but the `onSaveHandler` in GeoModal here
https://github.com/3DStreet/3dstreet/blob/a7dbaca9db29b71ff6e10fc44b740ecad34a761e/src/editor/components/modals/GeoModal/GeoModal.component.jsx#L144-L154
should definitely be rewritten to use `EntityUpdateCommand` similar to https://github.com/3DStreet/3dstreet/pull/789/files#diff-a70d189aca39070b7c4bd36206c8df559633374a33865caf6616323d1e403e4cR49 after the following PR is merged.
- #788

vincentfretin:
I did the above change in #789 but that was not the issue. I did fix it in https://github.com/3DStreet/3dstreet/pull/789/commits/69904017a9557f80e0881f8bb641fd655c50d28e though.

PR #788, opened 2024-08-26
change how elevation is set

rahulkgupta:
elevation isnt an input in the geomodal. On save, the elevation is fetched from the cloud function and set


vincentfretin (commented):
I didn't test, but the changes seems okay to me.

vincentfretin:
You can address my comment https://github.com/3DStreet/3dstreet/issues/770#issuecomment-2311979099 in this PR if you want, or we can do the change after the merge.

kfarr:
this is a good step forward:
- it was unclear after clicking update why the modal wasn't closing. obviously the answer is waiting for remote server. i wonder if there's a way to copy the save indicator

kfarr (approved):

rahulkgupta (commented):
nice!

PR #806, opened 2024-08-29
change the geo right panel

rahulkgupta:
Screenshot 2024-08-29 at 2 48 23 PM
Screenshot 2024-08-29 at 2 48 09 PM
Screenshot 2024-08-29 at 2 51 36 PM

todo
- [x] fix css for advanced in non-geo entities


vincentfretin:
There is a conflict to resolve. Do you need help?

rahulkgupta:
will do that today

Issue #798, opened 2024-08-29
(wb/cam) double click camera focus should not be placed on ground, add y (height)


kfarr:
today 
![image](https://github.com/user-attachments/assets/fc4bdffc-5e7f-4c1d-a844-562578c253d4)

instead
![image](https://github.com/user-attachments/assets/a62869fa-7e79-44ef-a6d4-85440a190d82)

Issue #797, opened 2024-08-29
(wb/cam) double-click entity to center camera in scenegraph should also apply to viewport double click


kfarr:
upon double clicking the object is selected and the camera is placed at a distance relative to the size of the bounding box of the selected object, including a y (height)

vincentfretin:
You disabled the dblclick behavior long time ago, I think mainly because of #710 that we fixed now, so you can compare viewport.js with the one from aframe-inspector to get back the behavior. There is still the y position issue though.

Issue #693, opened 2024-07-05
wrong camera angle when switch to Cross Section View after zooming in on Plan View


Algorush:
I found another issue with ortho camera. If zoom in on the camera in Plan view and then switch to Cross Section View, the camera angle takes on random values, and this is not related to changes in window resize:
![after_zoom](https://github.com/3DStreet/3dstreet/assets/32197767/c24ccb6e-c0a6-4d06-85d8-dba3fd884ef6)

vincentfretin:
Even after #694 changes, I indeed had that issue on 3dstreet, and also on my project when switching to ortho face view, but I had the issue one time, it's really random, I can't reproduce the issue consistently.

vincentfretin:
I reproduced this weird view if I focus an object in the scene graph by double clicking the item. I'll disable focus when in orthographic camera.

PR #807, opened 2024-08-30
Focus object with better camera position and transition animation

vincentfretin:
This closes:
- #797 
- #693 
- #798 
- #799

### DIFF
--- a/src/editor/lib/EditorControls.js
+++ b/src/editor/lib/EditorControls.js
@@ -66,6 +66,10 @@ THREE.EditorControls = function (_object, domElement) {
     }
     var distance;
 
+    // Save current camera position/quaternion
+    scope.transitionCamPosStart.copy(object.position);
+    scope.transitionCamQuaternionStart.copy(object.quaternion);
+
     box.setFromObject(target);
 
     if (box.isEmpty() === false && !isNaN(box.min.x)) {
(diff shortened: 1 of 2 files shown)


src/editor/lib/EditorControls.js
```diff
@@ -76,12 +83,73 @@ THREE.EditorControls = function (_object, domElement) {
...
+          this.transitioning = false;
+          object.position.copy(this.transitionCamPosEnd);
+          object.quaternion.copy(this.transitionCamQuaternionEnd);
+          scope.dispatchEvent(changeEvent);
+        }
+      }
+    }
   };
+  // Register the tick function with the render loop
+  AFRAME.scenes[0].addBehavior(this.fakeComponent);
```
vincentfretin, line 152: Note that addBehavior is not a public aframe api. The fakeComponent here with the isPlaying flags works on aframe 1.5.0, but it doesn't seem to work on aframe master, the tick is not called. I know there was some changes to behaviors with component order. I'm investigating it. I'll do another PR once I found something that works on both 1.5.0 and master, if this one is already merged.

vincentfretin:
The logic of the transition animation with the slerp was taken from https://github.com/c-frame/aframe-cursor-teleport/blob/master/index.js so I don't expect any issue.

src/editor/lib/EditorControls.js
kfarr: noted, thanks @vincentfretin for the heads up

vincentfretin: I tried to register a real component, then you need to add an entity to the scene with this component, then you need to handle for it to play, and keep playing if you switch back and forth between viewer and editor mode. That's more brain wracking than continuing with this hack, at least the changes are isolated there. I fixed it for aframe 1.6.0 in #817

PR #809, opened 2024-08-31
New sky

kfarr:


src/components/street-environment.js
```diff
@@ -19,138 +19,115 @@ AFRAME.registerComponent('street-environment', {
...
+      this.createLight('env-light1', { type: 'ambient', color: '#FFF' });
+    this.light1.setAttribute('data-layer-name', 'Ambient Light');
+    this.light2 =
+      this.el.sceneEl.querySelector('#env-light2') ||
+      this.createLight('env-light2', { type: 'directional', castShadow: true });
+    this.light2.setAttribute(
+      'data-layer-name',
+      'Directional Light • Shadow Caster'
+    );
+    this.setEnvOption();
```
vincentfretin, line 37: You don't need to call `this.setEnvOption();` in init. It's already called in update. Here it's executed twice when the component initialize.

src/street-utils.js
```diff
@@ -37,7 +37,7 @@ function newScene(
     'Environment'
   );
   environmentEl.removeAttribute('street-environment');
-  environmentEl.setAttribute('street-environment', 'preset', 'day');
+  environmentEl.setAttribute('street-environment');
```
vincentfretin, line 40: There is an error here. You need a second param ''

vincentfretin (commented):
I reviewed just the logic here, I didn't test.

vincentfretin (approved):
I tested, this seems to work as expected.

kfarr:
thanks, one remaining issue noted here: https://github.com/3DStreet/3dstreet/commit/04dc429ad68628396e1a9aef3be9245512677ef0#commitcomment-146063965 not sure how to fix

kfarr:
thanks @vincentfretin it seems like there are a lot of race conditions in a-frame that we don't encounter until doing dynamic loading in apps like these. (ie just loading a scene from scratch in html is a much easier case to solve for) 

looking forward to all of the rest and fixing them together so that a-frame is a very stable option for everyone :)

vincentfretin:
I can't reproduce it. How do you test it? Switching from an image (for example cloudy) to color from the UI panel take me 1s, the color is set, I see the image still downloading, and when the download is finished, I still see the color.

kfarr:
> I can't reproduce it. How do you test it? Switching from an image (for example cloudy) to color from the UI panel take me 1s, the color is set, I see the image still downloading, and when the download is finished, I still see the color.

here's a video, perhaps what i didn't explain is that i was not changing the color in the right-panel but instead opening a scene with a gray color but it is not loading unless I use the timeout hack, in this case it's the "Funny Waymo with renaming" scene:
https://github.com/user-attachments/assets/88a66343-f303-4058-b27d-87f7da9000c7

vincentfretin:
Got it, I reproduce it.
When you load a scene, it executes the newScene function with
```js
  environmentEl.removeAttribute('street-environment');
  environmentEl.setAttribute('street-environment', '');
```
so removing the component, adding it back setting day preset
then createEntities is called that will remove the environmentEl entity with the setBackground still pending, and remove an create the entity with a new component instance having the color preset.
The guard we put works only for the same component instance.
So what is happening is that the setBackground call of the old environment entity with an old component instance finish and don't see we now switched to color preset with a new component instance. So we can add an additional check for this.el.parentEl null and ignore the loaded texture to fix the issue.

Another fix is the following, instead of removing the component and setting again, you can use the third param clobber to true, that will reset the component to default values. https://aframe.io/docs/1.6.0/core/entity.html#putting-multi-property-component-data
From my tests, it seems that if you remove a component, set a component and remove entity, it will still initialize the component, this is weird but I think this i what I saw. But if you reset a component with clobber and then remove the entity, it won't update the component.

I did both fixes.

kfarr:
Thanks @vincentfretin ! Works great, thanks for using both methods

PR #814, opened 2024-08-31
Remove h shortcut that does nothing now we removed the help modal

vincentfretin:


vincentfretin:
Also removed in documentation https://github.com/3DStreet/3dstreet-docs/pull/47

PR #816, opened 2024-08-31
Remove ctrl+x action

vincentfretin:
See https://github.com/aframevr/aframe-inspector/pull/759 for more context.

Also here ctrl+x then ctrl+v didn't do anything anymore with our EntityCloneCommand, the cut entity is not in the DOM so getElementById return null.
https://github.com/3DStreet/3dstreet/blob/3cf8194117ef02004d9763dfe0401dbc2c072cfe/src/editor/lib/commands/EntityCloneCommand.js#L20

I know we don't use the help modal anymore here, but I still updated it.


vincentfretin:
I also updated the shortcuts doc here https://github.com/3DStreet/3dstreet-docs/pull/47

kfarr:
@vincentfretin ok with me to remove for now. may need to bring back someday

PR #818, opened 2024-08-31
Update transform controls scale on camera change

### LINKED ISSUE #812: (wb/cam) transform control bug fixing - don't display in "grab" mode, sizing incorrect
Issue1: As a user traveling to show points of interest by double-clicking on layers, I do not want the transform controls to appear when the "grab" tool is selected and I double-click on a layer.

Issue2: Further, if I am using the "move" or "rotate" tool I want the transform controls to appear on the newly selected (double-clicked) object, however I do not want them to be huge after the camera moves to the newly selected object.

Diagnosis on issue 2
- when the transform controls are created for a newly selected object, it appears to use the "screen space" to determine the size of the controls to ensure consistency between object selection (see capture1)
- when the camera moves after the transform controls are created, they will appear larger the further away the newly selected object is from the previous camera location. For example the next nearest objects (see capture2 below) and (see capture3 below)

capture1:
Image

capture2:
Image

capture3:
Image

Idea for alternative behavior for issue 2:
- instead, perhaps the transform controls can be resized in realtime depending on the screenspace size as it varies during camera transitions?

vincentfretin:
This includes
- https://github.com/aframevr/aframe-inspector/pull/735
- https://github.com/aframevr/aframe-inspector/pull/761 (issue2 of #812)
- fix for issue1 of #812

This closes #812

### DIFF
--- a/src/editor/lib/EditorControls.js
+++ b/src/editor/lib/EditorControls.js
@@ -96,7 +96,6 @@ THREE.EditorControls = function (_object, domElement) {
     scope.transitionSpeed = 0.001;
     scope.transitionProgress = 0;
     scope.transitioning = true;
-    // The changeEvent is emitted at the end of the transition below
   };
 
   function easeInOutQuad(t) {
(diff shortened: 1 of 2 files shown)


kfarr:
@vincentfretin works great

Issue #793, opened 2024-08-26
save layer visibility state, layer visibility should be undo-able


kfarr:
As a user who hides a layer using the scene-graph panel layer checkbox, I expect the layer that I have hidden to remain hidden upon opening the scene again in a later session.

Also, hiding / showing layers via the scene-graph panel should be a undo/redo-able action

vincentfretin:
The undoable part is fixed in #820 
For saving the visible component, not sure where the issue is. I'm starring at json-utils, nothing caught my eyes.

kfarr:
Is visibility treated like a component named `visible` with a single value of `true` or `false`? That is different than material.visibility under the hood? Or the same thing?

Looking at the json of a saved scene, we are not saving `visible`. It might be something to explicitly support in a whitelist somewhere

vincentfretin:
It should be a `visible` component yes. The `visible` component modify `el.object3D.visible`, so nothing to do with material.

kfarr:
closed by #820

PR #820, opened 2024-08-31
Toggling entity visibility is undoable

vincentfretin:
This fixes one part of #793

### DIFF
--- a/src/editor/components/scenegraph/Entity.js
+++ b/src/editor/components/scenegraph/Entity.js
@@ -29,7 +29,11 @@ export default class Entity extends React.Component {
       entity.tagName.toLowerCase() === 'a-scene'
         ? entity.object3D.visible
         : entity.getAttribute('visible');
-    entity.setAttribute('visible', !visible);
+    AFRAME.INSPECTOR.execute('entityupdate', {
+      entity,
+      component: 'visible',
+      value: !visible
+    });
   };
 
   render() {
--- a/src/editor/components/scenegraph/SceneGraph.js
+++ b/src/editor/components/scenegraph/SceneGraph.js
@@ -39,8 +39,8 @@ export default class SceneGraph extends React.Component {
     );
   }
 
-  onMixinUpdate = (detail) => {
-    if (detail.component === 'mixin') {
+  onEntityUpdate = (detail) => {
+    if (detail.component === 'mixin' || detail.component === 'visible') {
       this.rebuildEntityOptions();
     }
   };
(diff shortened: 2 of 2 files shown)


kfarr:
@vincentfretin ok i made an attempt to save/load visible. It worked in my testing but I would appreciate you to look it over again. I was worried about javascript's confusion sometimes to detect the existence of a property whose value is false vs. a property that does not exist, but `entity.getAttribute('visible') === false` seems to result in the desired behavior.

vincentfretin:
The visible key should really be in the components object, so not a correct fix, but you gave me an idea of the real issue.

vincentfretin:
There were two issues here to be able to export `"visible": "false"`
First is the condition `if (modifiedProperty)` that should check explicitly `if (modifiedProperty !== null)`
After fixing this, it wrongly exported `"visible": ""`
The second issue was the `isEmpty(modifiedProperty)`. This function didn't check for an object before using `Object.keys`. `Object.keys(false)` gave `[]` so `[].length === 0` and was putting `""` in the export.
Also in javascript, to detect if a variable is an object, `typeof variable === "object"` is not enough, because null and an array are also this type, so you need to check for not null and not array before using `Object.keys` here.

kfarr:
thinking out loud, one way to quickly test this is to compare scene json export from before / after the fix on a complicated scene (without hidden visibility, since that is new behavior) and see if the output is identical?

kfarr:
here are results so far from my testing

old and new comparing json output for the same scene

* position and rotation switch places, and a few other props change ordering, no big deal (pic1)
* in one case, "scale" component values now appear in the new version where they didn't before with mixin = "stencils bike-arrow" (pic2) 
* previously some components included a `"visible": ""` json entry, this persists but the order changes, no big deal just fyi (pic3)
* in another case, "scale" component values do not appear in the new version where they did before for "stencils word-taxi" (pic4)

pic1
image

pic2
image

pic3
image

pic4
image

vincentfretin:
Yes, all the things you just reported existed before this change.
The scale component appearing sometimes, I reproduced it before the change. That's another issue I didn't want to adresse in this PR. I have my guess where the issue may be in the code, probably in https://github.com/3DStreet/3dstreet/blob/afb7aa2862a27ff427854be412d16c74800a0a02/src/json-utils_1.1.js#L312-L322 where we have an hard coded list of mixin properties to ignore instead of checking dynamically if componentName is used in one of the mixin and the value didn't change from the value of the mixin.

vincentfretin:
You can create a separate issue if you want to fix removing the scale, geometry, material components if that's the same value as the mixin, make the component order deterministic, or round the coordinates of position and rotation.
As far as I can tell this PR is okay to be merged.

vincentfretin:
exporting `"visible": ""` when you toggled twice an entity is expected, the entity now have the visible component and will be exported, it's empty string because the value true is the default for this component, so this the expected export here.

PR #841, opened 2024-09-05
add new compacted gravel and sandy drive lane textures

kfarr:
example however drive lane texture pending revisions
image


kfarr:
to use compacted gravel on sidewalk, select (sidewalk) segment, click show advanced, material, change src from `#seamless-sidewalk` to `#compacted-gravel-texture`
image

to use sandy drive lane, select (road) segment, click show advanced, material, change src from `#seamless-road` to `#seamless-sandy-road`. You may want to change the Y repeat value to 4 instead of 25.

PR #848, opened 2024-09-09
Add support for undo mixin change

### LINKED ISSUE #813: changing mixin should be undo-able
I forgot about that one.
https://github.com/3DStreet/3dstreet/blob/a1fc99c9e044b37e60ba9f2e0fb1424250d901c9/src/editor/components/components/Mixins.js#L57-L87

vincentfretin:
I added support for undo mixin change.
I also added support for undo id and class but there is no UI in 3dstreet for that.
This closes #813


kfarr:
works as expected

PR #865, opened 2024-09-13
autosave

rahulkgupta:
Works..

Todo
- [x] improve css of the saving icon
- [x] figure out save button functionality. do we keep the saving modal or change it?
- [x] test if the user isn't logged in
- [x] is it ok that autosave creates a new scene if the user is editing a community scene?
- [x] 1 second seems too slow
- [x] the save function itself isn't the quickest

### DIFF
--- a/src/components/notify.js
+++ b/src/components/notify.js
@@ -44,19 +44,22 @@ AFRAME.registerComponent('notify', {
     if (STREET) {
       STREET.notify = {};
       STREET.notify.successMessage = (messageText) => {
-        this.message(messageText, 'success');
+        return this.message(messageText, 'success');
       };
       STREET.notify.errorMessage = (messageText) => {
-        this.message(messageText, 'error');
+        return this.message(messageText, 'error');
       };
       STREET.notify.warningMessage = (messageText) => {
-        this.message(messageText, 'warning');
+        return this.message(messageText, 'warning');
+      };
+      STREET.notify.dismissNotification = (notification) => {
+        return this.notify.dismiss(notification);
       };
     }
   },
   message: function (messageText, messageType = 'info') {
     if (messageText && this.types.includes(messageType)) {
-      this.notify.open({
+      return this.notify.open({
         type: messageType,
         message: messageText
       });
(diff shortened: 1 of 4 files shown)


vincentfretin:
It can autosave locally, but I think it should only save to server if you already saved the scene and are the author of the scene.

rahulkgupta:
can keep manual save and just make it unclickable w/ a similar saving state (save button is a loading circle)

rahulkgupta:
If the user isn't logged in, we dont allow autosave

rahulkgupta:
i think it's kind of cool that opening a community scene and editing it runs an auto save as, but open to convo

rahulkgupta:
Going through posthog data, i think saving a scene from streetmix will be really handy. I'm seeing users who don't save streetmix scenes and are logged in.

kfarr:
- this is amazing!
- css changes: yes, I see an offset C instead of spinner. Also too large, would be nice to fit spinner within Save icon.
- offline to online may need more testing: prompts for save as whenever a change is made and internet is not available. when switching offline, no error or warning that offline, then upon online many save success notifications but did not test if the changes made offline then online when opened are as expected
- more than 1 save success notification visible at a time will obscure the toolbar. options: limit success notify to only 1 at a time. or relocate success notify to the Save button. Or, get rid of success notify and only notify upon error. However offline to online as noted above, and initial saving operations, may warrant a success notify.
- small chance of user error may result in "corrupt" files that require redoing or significant changes. options: save undo stack; save "checkpoints" as subcollections
- bandwidth uses whatever the scene size is from ~20k-100kb upload per save (depending on complexity). as long as a user isn't constantly saving every moment it is not a big use and doesn't warrant optimization, but bandwidth constricted users may experience issues. some day in the future there could be a way to stream diffs to update the json scene on the server for 1/1000th of the upload bandwidth use, firebase does support the concept of merge: true to update specific parts of the json doc instead of rewriting the whole thing.

rahulkgupta:
fixed offline to online. it was a function of checking scene author and returning false if the device went offline

rahulkgupta:
i wasn't able to repro the save notifications, mind sharing with me how you did it? I moved a bunch of things around and the code is set up to remove the old save notification

rahulkgupta:
CleanShot 2024-09-23 at 17 17 07

PR #869, opened 2024-09-16
upgrade posthog

rahulkgupta:


vincentfretin:
you need to commit the package-lock.json file as well

PR #873, opened 2024-09-19
Model card previews

rahulkgupta:
- assets now reads from catalog.json and will create mixins from there.
- catalog.json values contains a `src` field which is used for both glb and images.

todo
- [x] remove images from here.


src/assets.js
```diff
@@ -307,6 +293,20 @@ function buildAssetHTML(assetUrl, categories) {
...
+
+  // Iterate through catalog.json and add mixins to assetsHTML
+  catalog.forEach((item) => {
+    if (item.id && item.src) {
+      const mixinHTML = `
+        <a-mixin
+          id="${item.id}"
+          shadow
+          gltf-model="url(${item.src}/${item.id}.glb)"
+          category="${item.category || ''}"
```
vincentfretin, line 305: You should use `data-category` here unless you're creating a category component.

src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx
```diff
@@ -49,6 +49,9 @@ const getGroupedMixinOptions = () => {
     let mixinDescr = '';
 
     if (mixinDataFromCatalog) {
+      if (mixinDataFromCatalog.src) {
+        mixinImg = `${mixinDataFromCatalog.src}/${mixinId}.jpg`;
+      }
```
vincentfretin, line 54: Isn't it overwritten right away after the condition?

src/assets.js
rahulkgupta: i was following this.

```
  function addCategoryNamesToMixins(html, categoryName) {
    const parser = new DOMParser();
    const doc = parser.parseFromString(html, 'text/html');
    const mixinNodes = doc.querySelectorAll('a-mixin');
    mixinNodes.forEach((mixinNode) => {
      mixinNode.setAttribute('category', categoryName);
    });
    return doc.documentElement.innerHTML;
  }
```

PR #875, opened 2024-09-20
remove cross section since its broken

rahulkgupta:


vincentfretin:
I was going to say how is it broken, and I just reproduced right away #693 that I thought we fixed, just by clicking on Cross Section, back to 3D View, moved a bit and back to Cross Section
image

I tried to reproduce again and I couldn't reproduce. TT

vincentfretin:
Oh I just remembered what I just did, I have the steps to reproduce:
- open scene
- click Cross Section
- pan a bit to the right with left click 
- click 3D View
- click Cross Section

view is broken

PR #877, opened 2024-09-25
Clipping box v1

rahulkgupta:
remaining before merging:

- [x] ability to save and load -- saves correctly, but when loading a scene it cannot apply the clipping plane to the underground box presumably because the scene is still loading and the box does not exist yet
- [x] the box is still AABB instead of OBB, should be revised. See this example code that uses Box3 but also rotates it, and then applies world matrix https://github.com/aframevr/aframe/blob/v1.6.0/src/components/obb-collider.js#L141-L163

### DIFF
--- a/src/components/clipping-planes.js
+++ b/src/components/clipping-planes.js
@@ -1,0 +1,77 @@
+/* global AFRAME */
+
+AFRAME.registerComponent('clipping-planes', {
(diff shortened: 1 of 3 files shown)


kfarr:
closed by https://github.com/3DStreet/3dstreet/pull/882

PR #882, opened 2024-09-29
obb clipping

kfarr:
barely working v1

enable by pasting into console after loading a scene into editor with google3d tiles layer:
```
document.querySelector('#clipping-box').setAttribute('obb-collider2', "clipPlaneFunction: obb;"); 
document.querySelector('#clipping-box').play();
```

video:
https://github.com/user-attachments/assets/b29a31aa-d312-42b0-b87a-15988f23d775

some things this needs:
- see the other branch on how make the component usable as a boolean option in street geo component, instead of copy/paste console code --> when street-geo sets clipping ON, then the box is created with this component on it
- rename to 'obb-clipping' (component and system) 
- try setting the "known good" clipping box to be the same dimensions as the default street in the scene
- try moving the box to a child layer and/or adding obb-collider2 component to the underground box entity itself
- try moving obb-collider2 component to the 3dtiles itself using `trackedObject3D` property to reference the box
- try removing top/bottom planes to have infinite top/bottom
- once we figure out the right mechanism to allow the user to save the clipping settings, ensure it saves/loads correctly

### DIFF
--- a/src/components/obb-clipping.js
+++ b/src/components/obb-clipping.js
@@ -10,8 +10,7 @@ AFRAME.registerComponent('obb-clipping', {
       // empty string means use the same element as the source
       type: 'string'
     },
-    minimumColliderDimension: { default: 0.02 },
-    enabled: { type: 'boolean', default: true }
+    minimumColliderDimension: { default: 0.02 }
   },
 
   init: function () {
@@ -146,8 +145,8 @@ AFRAME.registerComponent('obb-clipping', {
   },
 
   applyClippingPlanes: function (clipPlanes) {
-    console.log('applyclipping', this.elementToClip);
-    console.log('applyclipping', this);
+    // console.log('applyclipping', this.elementToClip);
+    // console.log('applyclipping', this);
     if (!this.elementToClip) {
       this.fetchElementToClip();
     }
@@ -185,7 +184,6 @@ AFRAME.registerComponent('obb-clipping', {
   },
 
   remove: function () {
-    this.data.enabled = false;
     this.removeClippingPlanes();
   },
 
@@ -344,7 +342,7 @@ AFRAME.registerComponent('obb-clipping', {
       this.trackedObject3D = trackedElement.object3D;
       this.updateCollider();
     }
-    console.log('trackedElement', this.trackedObject3D);
+    // console.log('trackedElement', this.trackedObject3D);
     return this.trackedObject3D;
   },
 
@@ -355,10 +353,6 @@ AFRAME.registerComponent('obb-clipping', {
     var auxMatrix = new THREE.Matrix4();
 
     return function () {
-      console.log('tick');
-      if (!this.data.enabled) {
-        return;
-      }
       var obb = this.obb;
       var renderColliderMesh = this.renderColliderMesh;
       var trackedObject3D = this.checkTrackedObject();
--- a/src/components/obb-clipping.js
+++ b/src/components/obb-clipping.js
@@ -1,0 +1,624 @@
+/* global AFRAME */
+
+AFRAME.registerSystem('obb-collider2', {
+  schema: {
+    showColliders: { default: false }
+  },
+
+  init: function () {
+    this.collisions = [];
+    this.colliderEls = [];
+  },
+
+  addCollider: function (colliderEl) {
+    this.colliderEls.push(colliderEl);
+    if (this.data.showColliders) {
+      colliderEl.components['obb-collider2'].showCollider();
+    } else {
+      colliderEl.components['obb-collider2'].hideCollider();
+    }
+    this.tick = this.detectCollisions;
+  },
+
+  removeCollider: function (colliderEl) {
+    var colliderEls = this.colliderEls;
+    var elIndex = colliderEls.indexOf(colliderEl);
+    colliderEl.components['obb-collider2'].hideCollider();
+    if (elIndex > -1) {
+      colliderEls.splice(elIndex, 1);
+    }
+    if (colliderEls.length === 0) {
+      this.tick = undefined;
+    }
+  },
+
+  registerCollision: function (componentA, componentB) {
+    var collisions = this.collisions;
+    var existingCollision = false;
+    var boundingBoxA = componentA.obb;
+    var boundingBoxB = componentB.obb;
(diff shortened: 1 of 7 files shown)


src/components/obb-clipping.js
```diff
@@ -0,0 +1,624 @@
...
+        if (boxA.intersectsOBB(boxB)) {
+          this.registerCollision(componentA, componentB);
+        }
+      }
+    }
+    this.clearCollisions();
+  }
+});
+
+AFRAME.registerComponent('obb-collider2', {
```
rahulkgupta, line 161: there can be two `obb-collider2` registered?

rahulkgupta (commented):
the math in this makes my head spin

src/components/obb-clipping.js
rahulkgupta: o one is a system

rahulkgupta:
What's the other branch?

kfarr:
> What's the other branch?

https://github.com/3DStreet/3dstreet/pull/877

kfarr:
next steps:
- test if I can put the component on the street underground box instead of a root box

kfarr:
barely working 3

issues to fix:
- data-no-pause doesn't seem to work right when loading the box at first. need to enter viewer mode and then return to 
- overloading dirt / underground box with clipping seems not like desired user behavior
- it may make sense to have an explicit clipping box that is created for this purpose at the same level as the underground box but a separate box
- it is invisible by default, shown with a wireframe if visible

kfarr:
ok great progress so far:
- now you can specify geo, place a scene, and then use the "underground" layer to define your clipping area, then uncheck the visibility checkbox from underground

things to fix:
- improve logic to specify a target clipping element (the element to be clipped), right now '#google3d' id selector is hardcode
- improve logic to specify a source clipping element (the element to define clipping volume) ... such as if a street is not present, create a clipping box in the user root?

kfarr:
Progress notes:
- now this will look for a target of [street] and return bounding box of the first street and subelements it finds
- todo: consider throttle ticking, remove logging, consider removing things from tick as much as possible since this is no longer a collision component

Not working yet:
- if clipping is turned on, and then i turn it off during the session, it will not remove clipping and the tick appears to still run even though the component has been removed

kfarr:
Now ready for testing / code review:
* login with geospatial pro account
* Choose a location to enable google3d map source
* Add a street
* Click geospatial layer, open right panel, click advanced, click enableClipping
* Need more clipping? Change the size of our scene by placing objects higher or to the side. You can also place an object and toggle its visibility off.

rahulkgupta:
works!
- testing placing things higher also works. note, scaling a lightpole doesn't change the clipping box

src/components/obb-clipping.js
```diff
@@ -0,0 +1,402 @@
...
+    },
+    clippingDestinationSelectorString: {
+      // empty string means use the same element as the source
+      type: 'string'
+    },
+    minimumColliderDimension: { default: 0.02 }
+  },
+
+  init: function () {
+    this.tick = AFRAME.utils.throttleTick(this.tick, 250, this);
```
rahulkgupta, line 17: for my education, what's the reason behind this?

src/editor/components/components/AddLayerPanel/createLayerFunctions.js
```diff
@@ -61,7 +62,7 @@ export function createStreetmixStreet(position, streetmixURL, hideBuildings) {
   // position the street further from the current one so as not to overlap each other
   if (streetmixURL && streetmixURL !== '') {
     const definition = {
-      id: streetmixURL,
+      id: createUniqueId(),
```
rahulkgupta, line 65: why this change?

src/components/street-geo.js
```diff
@@ -113,6 +124,10 @@ AFRAME.registerComponent('street-geo', {
     const create3DtilesElement = () => {
       const google3dElement = document.createElement('a-entity');
       google3dElement.setAttribute('data-no-pause', '');
+      google3dElement.id = 'google3d';
+      if (data.enableClipping) {
+        google3dElement.setAttribute('obb-clipping', '');
```
rahulkgupta, line 129: just to confirm, setting this attribute is what enables the `obb-clipping` component to "run" right?

rahulkgupta (commented):
mostly just questions that dont need an immediate answer

src/components/obb-clipping.js
kfarr: this throttles tick to only fire every 250ms instead of every frame. recalculating the bounding box and then traversing all the tiles to apply the clipping planes is very expensive.

better solution is to monitor for pose changes on target object OR tiles loaded events from 3dtiles runtime object via threejs-loader-tiles library and ONLY THEN initiate updating obb and clipping planes, but throttling is a cheap easy way to do it for now

src/components/street-geo.js
kfarr: yes this adds the obb-clipping component to the 3dtiles which starts it running

src/editor/components/components/AddLayerPanel/createLayerFunctions.js
kfarr: streetmixURL is not unique, dragging and dropping the same street preset from the add layer panel will cause id collision errors

src/components/obb-clipping.js
```diff
@@ -0,0 +1,402 @@
...
+      // Update OBB visual representation.
+      if (renderColliderMesh) {
+        renderColliderMesh.matrixWorld.copy(auxMatrix);
+      }
+
+      // Reset OBB with AABB and apply entity matrix. applyMatrix4 changes OBB internal state.
+      obb.copy(this.aabb);
+      obb.applyMatrix4(auxMatrix);
+
+      const clipPlanes = this.createPlanesFromOBB(obb);
```
dmarcos, line 397: you don't need to update the clipping planes every frame. Can do similar to bounding box recalculation. Consider scenarios when you want to recalculate: rotation, scale, position changes

dmarcos:
```
    const clipPlanes = this.createPlanesFromOBB(obb);
    this.applyClippingPlanes(clipPlanes);
```

PR #887, opened 2024-10-02
Improve clipping performance

kfarr:
This makes use of a new event that is emitted from 3dtiles loader "contentPostProcess" after this modification:
https://github.com/3DStreet/aframe-loader-3dtiles-component/commit/339b735f6a454c1b6a8f823b1f0904f9ddb6b2d9#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346

Although most of the performance issues are the 3dtiles themselves regardless of clipping... That's a whole 'nother thing


src/components/obb-clipping.js
```diff
@@ -147,28 +161,28 @@ AFRAME.registerComponent('obb-clipping', {
...
+          ? obj.material
+          : [obj.material];
+        materials.forEach((material) => {
+          material.clippingPlanes = clipPlanes;
+          material.clipIntersection = true;
+        });
+      }
+    });
+  },
+
```
vincentfretin, line 185: If it's an hot path, you should avoid creating an array and prefer using for-of.
```js
   object3D.traverse((obj) => {
      if (obj.isMesh) {
        if (Array.isArray(obj.material)) {
          for (const material of obj.material) {
            material.clippingPlanes = clipPlanes;
            material.clipIntersection = true;
          }
        } else {
          obj.material.clippingPlanes = clipPlanes;
          obj.material.clipIntersection = true;
        } 
      }
    });
```

kfarr: thanks @vincentfretin, I just reverted back to this logic which was there before and i mistakingly tried to simplify

PR #888, opened 2024-10-02
more thumbnails

rahulkgupta:


src/assets.js
```diff
@@ -83,9 +83,6 @@ function buildAssetHTML(assetUrl, categories) {
         <a-mixin shadow id="Bicycle_1" gltf-part="src: #microMobilityDevices; part: Bicycle_1"></a-mixin>
         <a-mixin shadow id="ElectricScooter_1" gltf-part="src: #microMobilityDevices; part: ElectricScooter_1"></a-mixin>
         <!-- fantasy vehicles -->
-        <a-asset-item id="magic-carpet-glb" src="${assetUrl}sets/magic-carpet/gltf-exports/draco/magic-carpet.glb"></a-asset-item>
```
rahulkgupta, line 86: got rid of the magic carpet stuff. didn't feel like it needed to be here, but lmk if that's incorrect

kfarr: I think it's a valid segment type from streetmix but it is rare. therefore it should be a "lazy load" reference from url instead of an a-asset-item that is loaded for all scenes. there is more backstory to this segment type from carlos, it may have been funded for streetmix while carlos was at numo.

kfarr:
added some more screenshots so that the path resolves https://github.com/3DStreet/3dstreet-assets-dist/commit/43c45305f0f1c2d7e8f321e77830058e783a6b14

src/assets.js
kfarr: added back for now, can discuss more

rahulkgupta: sounds good!

PR #891, opened 2024-10-09
quick prevention of accidentally hitting new when a scene is saving

rahulkgupta:
Saw this bug occur when demoing 3dstreet. if you hit new while the scene is auto-saving, it wipes the scene.


kfarr:
works great!

PR #900, opened 2024-10-24
save changes

rahulkgupta:
one interesting thing here is that screenshots are only set on the first time a scene is saved. won't this cause drift with the screenshot and what the user has created?

EDIT: actually, this relates to "set scene thumbnail". I dont think we need to ask the user to do this and instead we should just do this for the users ourselves.

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,6 +31,7 @@
 </head>
 
 <body>
+  <img id="screenshot-img" src="ui_assets/3DStreet-Viewer-Start-Editor.svg" alt="Invisible Image" style="display:none;">
   <!-- loading animation start -->
   <div class="loader__wrapper">
     <div class="loader">
@@ -47,7 +48,6 @@
       <div class="road">Loading 3DStreet</div>
     </div>
   </div>
-
   <!-- viewer ui start -->
   <div class="viewer-header-wrapper">
     <button class="viewer-logo-start-editor-button" onclick="startEditor()">
--- a/src/components/screentock.js
+++ b/src/components/screentock.js
@@ -21,7 +21,7 @@ AFRAME.registerComponent('screentock', {
     type: { type: 'string', default: 'jpg' }, // png, jpg, img
     imgElementSelector: { type: 'selector' }
   },
-  takeScreenshotNow: async function (filename, type, imgElement) {
+  takeScreenshotNow: function (filename, type, imgElement) {
     const inspector = AFRAME.INSPECTOR;
     const renderer = AFRAME.scenes[0].renderer;
 
@@ -32,7 +32,7 @@ AFRAME.registerComponent('screentock', {
       if (inspector && inspector.opened) inspector.sceneHelpers.visible = show;
     }
 
-    const createCanvasWithScreenshot = async (aframeCanvas) => {
+    const createCanvasWithScreenshot = (aframeCanvas) => {
       let screenshotCanvas = document.querySelector('#screenshotCanvas');
       if (!screenshotCanvas) {
         screenshotCanvas = document.createElement('canvas');
@@ -53,7 +53,7 @@ AFRAME.registerComponent('screentock', {
         screenshotCanvas.height
       );
       // add 3DStreet logo
-      await addLogoToCanvas(ctxScreenshot);
+      addLogoToCanvas(ctxScreenshot);
       return screenshotCanvas;
     };
 
@@ -68,21 +68,9 @@ AFRAME.registerComponent('screentock', {
       );
     }
 
-    const addLogoToCanvas = async (ctx) => {
-      const logoImg = document.querySelector('img.viewer-logo-img');
-      const logoSVG = document.querySelector('#aframeInspector #logoImg svg');
-
-      if (logoImg) {
-        ctx.drawImage(logoImg, 0, 0, 135, 43, 40, 30, 270, 86);
-      } else if (logoSVG) {
-        const image = new Image();
-        image.src = `data:image/svg+xml;base64,${window.btoa(logoSVG.outerHTML)}`;
-        await new Promise((resolve) => {
-          image.onload = resolve;
-        });
-
-        ctx.drawImage(image, 0, 0, 135, 23, 40, 40, 270, 46);
-      }
+    const addLogoToCanvas = (ctx) => {
+      const logoImg = document.querySelector('#screenshot-img');
+      ctx.drawImage(logoImg, 0, 0, 135, 43, 40, 30, 270, 86);
     };
 
     function downloadImageDataURL(filename, dataURL, scnrenshotCanvas) {
@@ -103,9 +91,7 @@ AFRAME.registerComponent('screentock', {
 
     // render one frame
     renderer.render(AFRAME.scenes[0].object3D, AFRAME.scenes[0].camera);
-    const screenshotCanvas = await createCanvasWithScreenshot(
-      renderer.domElement
-    );
+    const screenshotCanvas = createCanvasWithScreenshot(renderer.domElement);
 
     if (type === 'img') {
       imgElement.src = screenshotCanvas.toDataURL();
--- a/src/editor/api/scene.js
+++ b/src/editor/api/scene.js
@@ -51,6 +51,7 @@ const createScene = async (authorId, sceneData, title, version) => {
   });
   return newSceneId;
 };
+
 const deleteScene = async (sceneId) => {
   try {
     const sceneDocRef = doc(db, 'scenes', sceneId);
@@ -102,35 +103,6 @@ const getScene = async ({ sceneId }) => {
   const sceneRef = doc(db, 'scenes', sceneId);
   const sceneSnapshot = await getDoc(sceneRef);
   return sceneSnapshot;
-};
-
-const isSceneAuthorWithSnapshot = async ({ sceneSnapshot, authorId }) => {
-  console.log('sceneSnapshot', sceneSnapshot.blah);
-  if (!sceneSnapshot || !authorId) {
-    console.log('sceneSnapshot or authorId is not provided in isSceneAuthor2');
-    return false;
-  }
-  try {
-    if (sceneSnapshot.exists()) {
-      return sceneSnapshot.data().author === authorId;
-    } else {
-      console.error('Scene not found while running isSceneAuthor');
-      return false;
-    }
-  } catch (error) {
-    console.error('Error fetching scene while running isSceneAuthor:', error);
-    throw new Error('Error checking scene authorship');
-  }
-};
-
-const isSceneAuthor = async ({ sceneId, authorId }) => {
-  if (!sceneId || !authorId) {
-    console.log('sceneId or authorId is not provided in isSceneAuthor');
-    return false;
-  }
-  // Get a reference to the scene document
-  const sceneSnapshot = await getScene(sceneId);
-  return isSceneAuthorWithSnapshot({ sceneSnapshot, authorId });
 };
 
 let scenesSnapshot;
@@ -315,8 +287,6 @@ export {
   getCommunityScenes,
   getUserScenes,
   getScene,
-  isSceneAuthorWithSnapshot,
-  isSceneAuthor,
   updateScene,
   updateSceneIdAndTitle,
   uploadThumbnailImage,
--- a/src/editor/api/scene.js
+++ b/src/editor/api/scene.js
@@ -18,13 +18,13 @@ import { db, storage } from '../services/firebase';
 import { getDownloadURL, ref, uploadBytes } from 'firebase/storage';
 import posthog from 'posthog-js';
 
-const generateSceneId = async (authorId) => {
-  const userScenesRef = collection(db, 'scenes');
+const sceneRef = collection(db, 'scenes');
 
+const generateSceneId = async (authorId) => {
   // Generate a new UUID
   const newSceneId = uuidv4();
 
-  const newSceneDocRef = doc(userScenesRef, newSceneId);
+  const newSceneDocRef = doc(sceneRef, newSceneId);
 
   // Use setDoc to set data on the specified document
   await setDoc(newSceneDocRef, {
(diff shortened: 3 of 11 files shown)


index.html
```diff
@@ -31,6 +31,7 @@
 </head>
 
 <body>
+  <img id="screenshot-img" src="ui_assets/3DStreet-Viewer-Start-Editor.svg" alt="Invisible Image" style="display:none;">
```
rahulkgupta, line 34: created an invisible image so that the screentock functions can be synchronous

src/components/screentock.js
```diff
@@ -21,7 +21,7 @@ AFRAME.registerComponent('screentock', {
     type: { type: 'string', default: 'jpg' }, // png, jpg, img
     imgElementSelector: { type: 'selector' }
   },
```
rahulkgupta, line 23: removed all the async code from here

src/editor/api/scene.js
```diff
@@ -189,6 +191,102 @@ const checkIfImagePathIsEmpty = async (sceneId) => {
   }
 };
 
+const saveScreenshot = async (value) => {
```
rahulkgupta, line 194: moved screenshot functionality to the scene api file so that the toolbar component didn't call functions from the screenshotmodalcomponent

src/editor/api/scene.js
```diff
@@ -189,6 +191,101 @@ const checkIfImagePathIsEmpty = async (sceneId) => {
...
+    scene_id: STREET.utils.getCurrentSceneId()
+  });
+
+  screenshotEl.setAttribute('screentock', 'type', value);
+  screenshotEl.setAttribute('screentock', 'takeScreenshot', true);
+};
+
+const uploadThumbnailImage = async () => {
+  try {
+    // saveScreenshot('img');
```
rahulkgupta, line 217: note, i commented this out, since in every call `makeScreenshot` is called prior

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -30,7 +30,6 @@ export default class Toolbar extends Component {
     this.state = {
       // isPlaying: false,
       isSaveActionActive: false,
-      isCapturingScreen: false,
```
rahulkgupta, line 33: removed all state related functionality for this

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -298,34 +290,22 @@ export default class Toolbar extends Component {
...
-        component.setState((prevState) => ({
-          ...prevState,
-          isCapturingScreen: false
-        }));
-    });
-  // openViewMode() {
-  //   AFRAME.INSPECTOR.close();
-  // }
+  makeScreenshot = () => {
+    const imgHTML = '<img id="screentock-destination">';
```
rahulkgupta, line 294: use the invisible imgHTML

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -343,8 +323,8 @@ export default class Toolbar extends Component {
   };
 
   toggleSaveActionState = () => {
+    this.makeScreenshot();
```
rahulkgupta, line 326: explicitly call makescreenshot rather than state management

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -418,10 +398,7 @@ export default class Toolbar extends Component {
           <Button
             leadingIcon={<ScreenshotIcon />}
             onClick={() => {
-              this.setState((prevState) => ({
-                ...prevState,
-                isCapturingScreen: true
-              }));
+              this.makeScreenshot();
```
rahulkgupta, line 401: call makescreenshot instead of using state

rahulkgupta (commented):
I think there is more cleanup but this is a good enough start. should deal w/ the save as issue

src/editor/api/scene.js
```diff
@@ -44,24 +62,16 @@ const deleteScene = async (sceneId) => {
   }
 };
 
-const updateScene = async (sceneId, userUID, sceneData, title, version) => {
```
rahulkgupta, line 47: no need to pass user id in updating a scene

src/editor/api/scene.js
```diff
@@ -34,6 +36,22 @@ const generateSceneId = async (authorId) => {
   return newSceneId;
 };
 
+const createScene = async (authorId, sceneData, title, version) => {
```
rahulkgupta, line 39: added a createScene

src/editor/api/scene.js
```diff
@@ -88,28 +96,6 @@ const updateSceneIdAndTitle = async (sceneId, title) => {
   }
 };
 
-const isSceneAuthor = async ({ sceneId, authorId }) => {
```
rahulkgupta, line 91: no longer needed

src/editor/components/modals/ScenesModal/ScenesModal.component.jsx
```diff
@@ -64,10 +64,9 @@ const ScenesModal = ({ isOpen, onClose, initialTab = 'owner', delay }) => {
 
       const sceneId = scene.id;
       const sceneTitle = sceneData.title;
-
       AFRAME.scenes[0].setAttribute('metadata', 'sceneId', sceneId);
       AFRAME.scenes[0].setAttribute('metadata', 'sceneTitle', sceneTitle);
-
+      AFRAME.scenes[0].setAttribute('metadata', 'authorId', sceneData.author);
```
rahulkgupta, line 69: when a new scene is loaded, set the authorId

src/editor/components/modals/ScreenshotModal/ScreenshotModal.component.jsx
```diff
@@ -1,128 +1,18 @@
...
 import Toolbar from '../../scenegraph/Toolbar';
 import Modal from '../Modal.jsx';
 import posthog from 'posthog-js';
 import { saveBlob } from '../../../lib/utils';
 import Events from '../../../lib/Events';
-
+import { uploadThumbnailImage, saveScreenshot } from '../../../api/scene';
 // import { loginHandler } from '../SignInModal';
 
-export const uploadThumbnailImage = async (uploadedFirstTime) => {
```
rahulkgupta, line 26: move this to a shared location so that Toolbar isn't calling this component

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -28,23 +27,17 @@ export default class Toolbar extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      // isPlaying: false,
       isSaveActionActive: false,
-      isCapturingScreen: false,
       showLoadBtn: true,
-      savedNewDocument: false,
```
rahulkgupta, line 35: didnt seem like this was needed

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -28,23 +27,17 @@ export default class Toolbar extends Component {
...
       isSavingScene: false,
       pendingSceneSave: false,
-      signInSuccess: false,
-      isAuthor: props.isAuthor,
```
rahulkgupta, line 39: removed these states because they didn't need to exist

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -55,51 +48,23 @@ export default class Toolbar extends Component {
   }
 
   componentDidUpdate(prevProps) {
```
rahulkgupta, line 50: cleaned up componentdidupdate. only calls `cloudSaveHandlerWithImageUpload` if pendingSave is true and there is a currentUser

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -55,51 +48,23 @@ export default class Toolbar extends Component {
...
-
-    if (
-      this.state.isCapturingScreen &&
-      prevProps.isCapturingScreen !== this.state.isCapturingScreen
-    ) {
-      this.makeScreenshot(this);
-    }
   }
 
-  checkSignInStatus = async () => {
```
rahulkgupta, line 88: im not sure why this existed. deleted

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -55,51 +48,23 @@ export default class Toolbar extends Component {
...
-        setTimeout(this.checkSignInStatus, 500);
-      }
-    }
-  };
-
   componentWillUnmount() {
     document.removeEventListener('click', this.handleClickOutsideSave);
   }
 
+  isAuthor = () => {
```
rahulkgupta, line 64: stateless isAuthor function that checks currentUser and street util function

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -137,8 +102,13 @@ export default class Toolbar extends Component {
     }
   };
 
-  cloudSaveAsHandler = async () => {
-    this.cloudSaveHandler({ doSaveAs: true });
+  cloudSaveHandlerWithImageUpload = async (doSaveAs) => {
```
rahulkgupta, line 105: changed cloudSaveAs to be a wrapper of cloudsave w/ uploading an image, since image uploads happen primarily on save as (unless something goes wrong)

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -182,25 +152,20 @@ export default class Toolbar extends Component {
         Events.emit('openpaymentmodal');
         return;
       }
-      let isCurrentUserTheSceneAuthor;
```
rahulkgupta, line 185: no more network requests to see if the current user is the author. we already know that

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -218,48 +183,42 @@ export default class Toolbar extends Component {
         console.log(
           'no urlSceneId or doSaveAs is true, therefore generate new one'
         );
-        currentSceneId = await generateSceneId(this.props.currentUser.uid);
+        currentSceneId = await createScene(
```
rahulkgupta, line 186: instead of generating a scene id and then updating a scene, we are simply creating a scene w/ all the relevant info

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -218,48 +183,42 @@ export default class Toolbar extends Component {
...
+          this.props.currentUser.uid,
+          filteredData.data,
+          currentSceneTitle,
+          filteredData.version
+        );
         console.log('newly generated currentSceneId', currentSceneId);
-        window.location.hash = `#/scenes/${currentSceneId}.json`;
```
rahulkgupta, line 223: we are doing this later anyways, so this is duplicative

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -218,48 +183,42 @@ export default class Toolbar extends Component {
...
-
       // make sure to update sceneId with new one in metadata component!
-      AFRAME.scenes[0].setAttribute('metadata', 'sceneId: ' + currentSceneId);
-
-      const isImagePathEmpty = await checkIfImagePathIsEmpty(currentSceneId);
-      if (isImagePathEmpty) {
-        await uploadThumbnailImage(true);
-      }
+      AFRAME.scenes[0].setAttribute('metadata', 'sceneId' + currentSceneId);
+      AFRAME.scenes[0].setAttribute(
```
rahulkgupta, line 209: also updating the authorid now

src/editor/components/scenegraph/Toolbar.js
```diff
@@ -218,48 +183,42 @@ export default class Toolbar extends Component {
...
-      await updateScene(
-        currentSceneId,
-        this.props.currentUser.uid,
-        filteredData.data,
-        currentSceneTitle,
-        filteredData.version
-      );
-
       // make sure to update sceneId with new one in metadata component!
-      AFRAME.scenes[0].setAttribute('metadata', 'sceneId: ' + currentSceneId);
```
rahulkgupta, line 246: **note**, i change `sceneId: ` to `sceneId`

src/editor/components/scenegraph/ToolbarWrapper.js
```diff
@@ -1,31 +1,9 @@
...
-          console.error('Error:', error);
-        }
-      }
-    }
-
-    checkAuthorship();
-  }, [currentUser, currentSceneId]);
-
-  return <Toolbar currentUser={currentUser} isAuthor={isAuthor} />;
+  return <Toolbar currentUser={currentUser} />;
```
rahulkgupta, line 6: cleaned this up. no longer passing stuff that doesnt make sense

kfarr:
## meta comments:
- Before this PR, the state was duplicated across both React and A-Frame components with with Kieran and Alex's "spaghetti code." After this PR state is still duplicated, but there is a lot less spaghetti code and some things are much simpler such as handling of author. Active scene ID, title, author, etc. is probably a candidate for zustand or similar state management system when the time is right.

## testing:
I have a hard time testing the core issue of the "race conditions" of async preview generation not compatible

Instead, I have tested the following to see what is the outcome for preview image generation and uploading:
- press save button when not logged in (then logging in with either google or msft) -- results in uploading black box for thumbnail (if no prior screenshots taken) OR uploading the previously captured image (if successfully saved a capture earlier in the same session) [1] this is confusing as it could be a thumbnail from a scene that is not currently loaded. (ie a black box would be better than the incorrect thumbnail).
- when logged in, open another user's scene, then press "save as" from another user's scene -- WORKS as expected
- when logged in, open another user's scene, then move an item, then change the title in the title popup from another user's scene -- RESULT in no thumbnail (different than black, this is not uploading any thumbnail)
- when logged in, open my own scene, then press "save as" from my own scene -- WORKS as expected
- when logged in, open my own scene, then click share, then click "Set as scene thumbnail" Now I no longer see a success message after uploading (unless I look at console log) and I may think that something has gone wrong. I do see an unauthorized error if it is raised, which is expected behavior (if I attempt to change the thumbnail of a scene that is not my own)

screenshots:
[1] 
image

kfarr:
all looks great!

change noted as discussed: opening another user's scene, and moving an object, it no longer asks for save as

PR #928, opened 2024-11-13
Remove element tag and class

kfarr:


rahulkgupta:
for my knowledge, what was the purpose behind this change?

src/aframe-streetmix-parsers.js
```diff
@@ -1007,14 +1001,12 @@ function processSegments(
 
   // create and center offset to center the street around global x position of 0
   var streetParentEl = createCenteredStreetElement(segments);
-  streetParentEl.classList.add('street-parent');
```
vincentfretin, line 1010: Be careful, this class is checked in https://github.com/3DStreet/3dstreet/blob/2929a6d8b9aa5c1c2b3281fea2eb6659b26929e1/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx#L110-L111

src/aframe-streetmix-parsers.js
```diff
@@ -1726,7 +1722,6 @@ module.exports.processSegments = processSegments;
 function processBuildings(left, right, streetWidth, showGround, length) {
   const buildingElement = document.createElement('a-entity');
   const clonedObjectRadius = 0.45 * length;
-  buildingElement.classList.add('buildings-parent');
```
vincentfretin, line 1729: Be careful, this class is checked in https://github.com/3DStreet/3dstreet/blob/2929a6d8b9aa5c1c2b3281fea2eb6659b26929e1/src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx#L110-L111

kfarr:
suggestion: simplify this to remove element tag and tackle class later

src/aframe-streetmix-parsers.js
kfarr: ok we'll change this later

src/aframe-streetmix-parsers.js
kfarr: ok we'll change this later

vincentfretin (approved):

PR #930, opened 2024-11-14
Remove unused watch:css and add prettier-plugin-tailwindcss

vincentfretin:
- We don't need tailwindcss cli, webpack with postcss-loader is handling the css build
- Add prettier-plugin-tailwindcss to have a consistent order of tailwind classes through prettier.


vincentfretin:
Before
image

After
image

PR #941, opened 2024-11-20
Global clipping box option

kfarr:


kfarr:
Feature:
https://github.com/user-attachments/assets/d7cfa6f1-3085-44bd-bb9a-022d1fff0157

kfarr:
now using flattening instead

PR #949, opened 2024-11-29
Intersection right panel

rahulkgupta:
CleanShot 2024-11-29 at 11 30 12@2x


kfarr:
There is an error in the original intersection component before the changes in this PR. Although [directionOrder is defined in intersection component](https://github.com/3DStreet/3dstreet/blob/main/src/components/intersection.js#L18) as
```
    const directionOrder = ['west', 'east', 'north', 'south'];
```
for stopsign, trafficsignal, crosswalk and sidewalk, however the actual behavior of the component doesn't follow that expectation.

Instead, the actual behavior of the intersection component is as follows:
```
    stopsign: { type: 'string', default: '0 0 0 0' },      // east, west, north, south
    trafficsignal: { type: 'string', default: '1 1 1 1' }, // east, west, north, south
    crosswalk: { type: 'string', default: '1 1 1 1' },     // east, west, south, north
    sidewalk: { type: 'string', default: '0 0 0 0' },      // east, west, south, north
```

The UI follows directionOrder, so it exhibits the same error as the test above [1].

Instead, we could fix the intersection component to match the order defined in directionOrder. One strategy could be swapping the parameter names. 

[1] 
UI blackbox testing
 
correct:
north / south signal control option correct
all 4 curb corners

incorrect:
east / west traffic control option swapped
crosswalk & sidewalk north / south swapped
crosswalk & sidewalk east / west swapped

rahulkgupta (commented):
directions look good

PR #951, opened 2024-12-03
Default sidebar to be open

rahulkgupta:


src/editor/components/components/Sidebar.js
```diff
@@ -60,15 +60,15 @@ export default class Sidebar extends React.Component {
...
+    this.setState({ showSideBar: !this.state.showSideBar });
   };
 
   render() {
     const entity = this.props.entity;
     const visible = this.props.visible;
     const className = classnames({
       outliner: true,
-      hide: this.state.rightBarHide,
+      hide: this.state.showSideBar,
```
vincentfretin, line 71: I was very confused with this class name "hide" in combination with showSideBar. That class does a "display: block" :D The css that is enabled by this class is:
```css
.outliner.hide {
  display: block !important;
  height: 584px;
  width: 372px;
}
```

Issue #929, opened 2024-11-14
(uo) users editing intersection child objects are not aware that changes are not saved


kfarr:
For children of an intersection, it should not be a layer that is edited
- add data-no-transform
- add autocreated (if not already)

meta item:
- intersections need better ui

kfarr:
closed by #956

PR #956, opened 2024-12-07
do not allow transform of intersection children

kfarr:
https://github.com/3DStreet/3dstreet/issues/929


kfarr:
Feedback: also ignore raycaster (do not make these clickable)

PR #961, opened 2024-12-11
Don't transpile to ES5, no need to support IE11

vincentfretin:
See https://github.com/aframevr/aframe-inspector/pull/772 for details.


kfarr:
Thanks, works fine and reduces build size by ~300kb!

PR #963, opened 2024-12-12
Change layer name

kfarr:


src/editor/lib/commands/EntityRenameCommand.js
```diff
@@ -0,0 +1,61 @@
...
+    this.type = 'entityrename';
+    this.name = 'Rename Entity';
+    this.updatable = true; // Allow updating for consecutive renames
+
+    if (!entity.id) {
+      entity.id = createUniqueId();
+    }
+    this.entityId = entity.id;
+    this.oldName = entity.getAttribute('data-layer-name') || '';
+    this.newName = null; // Will be set during execute
```
vincentfretin, line 18: It needs to be set in the constructor otherwise you can't undo/redo

vincentfretin (commented):
There is no need for a new command. The entityupdate command already supports setting data attributes like data-layer-name. The prompt needs to be done before creating the command.

src/editor/lib/commands/EntityRenameCommand.js
kfarr: undo/redo works in my testing?

kfarr:
> There is no need for a new command. The entityupdate command already supports setting data attributes like data-layer-name. The prompt needs to be done before creating the command.

here is a branch using entityupdate however undo/redo doesn't work and i'm not sure what to change:
https://github.com/3DStreet/3dstreet/pull/964

i'm tempted to merge this one that works in my "black box" testing unless there's an easy way to make the entityupdate one work better

src/editor/lib/commands/EntityRenameCommand.js
vincentfretin: you're right it works, this.newName is set so it's fine afterwards

vincentfretin:
I fixed the issue in #964, the undo actually worked but the React UI wasn't updated.
In your rename command, it worked because you always called selectEntity that triggered the React update.

I would prefer we merge #964

kfarr:
https://github.com/3DStreet/3dstreet/pull/964

PR #964, opened 2024-12-14
change layer name -- without new command

kfarr:


kfarr:
works great, thanks @vincentfretin

PR #965, opened 2024-12-16
load from managed street object

kfarr:


src/components/managed-street.js
```diff
@@ -5,14 +5,137 @@ const { segmentVariants } = require('../segments-variants.js');
...
+  sourceValue: window.STREET.SAMPLE_STREET,
+  synchronize: true
+});
 userLayersEl.append(newStreetEl);
 */
 
+// Example Street object
+// 'width' this is the user-specified RoW width, not cumulative width of segments
+// some of this is redundant, many of the segment attributes defined below can be inferred by type
+window.STREET.SAMPLE_STREET = {
```
rahulkgupta, line 35: what's the purpose of having this and is there a way to have this outside of this file to keep it clean?

kfarr: it's for testing and yeah it should be somewhere outside of this file

kfarr: ok moved example scenes out of this file
https://github.com/3DStreet/3dstreet/pull/965/commits/cee039a1504ab40fa5b4a842afa57d64adb15c65

PR #971, opened 2024-12-19
Use random seed

kfarr:


kfarr:
issue: on first load, some random clones do not adhere to seed, but subsequent loads do.

causes: i thought it might be race conditions that the seed hadn't been set before the clones started generated on first load, so I attempted to fix that but didn't see improvement: https://github.com/3DStreet/3dstreet/pull/971/commits/c500b4b5d2eafb467254db196ec0edf5a51f0078

potential workaround: perhaps I can set seed value when instantiating the component from managed-street, although in a perfect world that wouldn't be necessary

screenshots:

load 1 (instantiated from streetmix -> managed street add layer button):
image

load 2 (loaded from cloud json)
image

load 3, 4, etc. (identical to load2)
image

kfarr:
also: pedestrian placement doesn't respect seed yet

vincentfretin:
The update function was called two times when creating the scene initially (you could see two floats with the console.log, not sure why) and only once when opening the scene. Just recreate the rng (that's just a function, that cheap) every time you're about to create the entities fixed the issue. I also reversed your logic, just checking things in update. Returning in init is a bit confusing because aframe will still call the update anyway.

kfarr:
@vincentfretin works great!

PR #981, opened 2024-12-21
In SceneGraph, only scroll node into view if it's not currently visible

vincentfretin:
Before:
[3dstreet-sg-scroll-before.webm](https://github.com/user-attachments/assets/4eea0e61-f320-4df8-8999-89eb2e979930)

After:
[3dstreet-sg-scroll.webm](https://github.com/user-attachments/assets/5f4f5e86-d90a-4793-a7f6-fba4d728934d)


kfarr:
@vincentfretin so much better, thanks!

Issue #967, opened 2024-12-17
(uo) intersection curbs should be curved


kfarr:
As a user I'd like to specify a radius for curves on corner curbs.

This would be a modification to the [intersection component](https://github.com/3DStreet/3dstreet/blob/main/src/components/intersection.js#L2) to accept a new property such as radius

A first version could take one radius, in the future there could be a different radius value for each corner.

here is an example diagram [1] of composing the northwest curb with a curve from primitive box and cylinder (with theta values to show only 90º arc for given radius)

Some possible methods:
1) combine multiple entities of geometry primitives (2 or 3 boxes + 1 arc cylinder) << easy to prototype
2) manually create the vertices for a custom geometry << more performant, more complicated
3) "boolean" geometric functions -- start with a box, then you "substract" the inverse of an arc cylinder  https://github.com/gkjohnson/three-bvh-csg << might be helpful for most complicated geometry such as adding curb ramps

[1]
![Image](https://github.com/user-attachments/assets/663cb245-52d5-4497-9c76-84d968e8154e)

PR #983, opened 2024-12-22
Curb radius option

ewei2406:
Adds the curb radius option for intersections: #967 

- Adds a new radius field under `width` and `height` in the intersection properties panel.
- Replaces the curb primitive geometry with a 3js mesh generated from the width, height, and radius parameters.
- Update the README with the changes to the `northeastcurb: ...`, `northwestcurb: ...`, ... API:
  The radius is the third value. Example: `northeastcurb: 4, 4, 1` for a 4m x 4m curb with a 1m radius in the corner of the intersection.
Note: the radius that is generated is bounded to the lower of the width and length of the curb. So a 4m x 6m curb will have a radius no smaller than 4m.

image
Screenshot 2024-12-21 at 10 01 30 PM

### DIFF
--- a/src/components/intersection.js
+++ b/src/components/intersection.js
@@ -97,10 +97,12 @@ AFRAME.registerComponent('intersection', {
       console.log(sidewalkTexture);
       sidewalkTexture.wrapS = THREE.RepeatWrapping;
       sidewalkTexture.wrapT = THREE.RepeatWrapping;
-      sidewalkTexture.repeat.set(2, 2);
+      sidewalkTexture.repeat.set(0.5, 0.5);
 
-      const material = new THREE.MeshBasicMaterial({
-        map: sidewalkTexture
+      const material = new THREE.MeshStandardMaterial({
+        map: sidewalkTexture,
+        roughness: 0.8,
+        color: 0xcccccc
       });
       const mesh = new THREE.Mesh(curbGeom, material);
       mesh.scale.setX(scaleVec.x);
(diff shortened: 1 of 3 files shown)


src/components/intersection.js
```diff
@@ -57,75 +59,112 @@ AFRAME.registerComponent('intersection', {
...
+      const material = new THREE.MeshStandardMaterial({
+        map: sidewalkTexture,
+        roughness: 0.8, // Same roughness as sidewalk mixin in src/assets.js
+        color: 0xcccccc // Darkens the texture to match existing sidewalk
+      });
+      const mesh = new THREE.Mesh(curbGeom, material);
+      mesh.scale.setX(scaleVec.x);
+      mesh.scale.setY(scaleVec.y);
+      mesh.scale.setZ(scaleVec.z);
+
```
vincentfretin, line 115: Because you create the three geometry and material yourself, the dispose of them are not handled by aframe, you need to take care of it yourself in `remove()`, assign the variables on `this` so you can then add:

```js
remove() {
  if (this.curbGeom) this.curbGeom.dispose();
  if (this.material) this.material.dispose();
}
```

The material previously could be shared among several meshes so for several intersections, that's a feature of aframe it does that automatically. Here we create one for each intersection. We could I guess use a global variable in this module to create only once the material if not currently exist (and not dispose of it in `remove()`).

vincentfretin (commented):
I just looked at the code and your image, that looks great!
I added a note about disposing the created resources.

src/components/intersection.js
ewei2406: Was not aware of that, thanks for the tip! I will add the global in `intersection.js` to avoid creating it multiple times then.

kfarr:
@ewei2406 this is awesome! I'll note that for users with existing intersections that do not have corner curb dimensions specified this update will use a default 1m curve radius. However if a curb corner does have height/width then the radius will be 0. I think this is reasonable behavior that a user would expect

image

src/components/intersection.js
vincentfretin: Yes, your changes are perfect. When I commented, I didn't pay attention at first we had 4 sidewalks in the intersection, so yes sharing the material for the 4 created sidewalks was a must indeed.
I was talking about having several intersections in the scene, we could have the material shared for all the sidewalks of all the intersections, but the hard part is to dispose the material only when all the intersections are removed. You can usually do that with an aframe system that do this reference counting, that's what the aframe material system is actually doing to dispose automatically the material when no other aframe entities is using it. But yeah maybe too much for this PR, and I don't think we'll have a lot of intersections in a single scene, probably only one or two. We can revisit later if we need some performance.

ewei2406: That makes sense. Thanks for the tip!

PR #987, opened 2024-12-25
Insert new segment

kfarr:
Insert a basic segment:

`document.querySelector('[managed-street]').components['managed-street'].insertSegment(0, 'drive-lane');`

Insert a segment with custom properties:

```
document.querySelector('[managed-street]').components['managed-street'].insertSegment(1, 'bike-lane', {
  width: 2.5,
  color: '#adff83',
  direction: 'inbound',
  name: 'Protected Bike Lane',
  generated: {
    stencil: [{
      model: 'bike-arrow',
      spacing: 20
    }],
    clones: [{
      mode: 'random',
      modelsArray: 'cyclist1, cyclist2, cyclist3',
      spacing: 2.03,
      count: 3
    }]
  }
});
```
Insert a segment with full configuration:

```
document.querySelector('[managed-street]').components['managed-street'].insertSegment(2, 'bus-lane', {
  width: 3.5,
  length: 60,
  level: 0,
  direction: 'outbound',
  color: '#ff9393',
  surface: 'asphalt',
  name: 'BRT Lane',
  generated: {
    stencil: [{
      stencils: 'word-only, word-bus',
      spacing: 40,
      padding: 10
    }],
    clones: [{
      mode: 'random',
      model: 'bus',
      spacing: 15,
      count: 1
    }]
  }
});
```

### DIFF
--- a/src/components/managed-street.js
+++ b/src/components/managed-street.js
@@ -53,8 +53,14 @@ AFRAME.registerComponent('managed-street', {
     // Bind the method to preserve context
     this.refreshFromSource = this.refreshFromSource.bind(this);
   },
-  // New method to insert a segment at a specific index
-  insertSegment: function (index, type) {
+  /**
+   * Inserts a new street segment at the specified index
+   * @param {number} index - The index at which to insert the new segment
+   * @param {string} type - The segment type (e.g., 'drive-lane', 'bike-lane')
+   * @param {Object} [segmentObject] - Optional configuration object for the segment
+   * @returns {Element} The created segment element
+   */
+  insertSegment: function (index, type, segmentObject = null) {
     // Validate index
     if (index < 0 || index > this.managedEntities.length) {
       console.error('[managed-street] Invalid index for insertion:', index);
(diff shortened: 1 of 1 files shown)


src/components/managed-street.js
```diff
@@ -53,6 +53,92 @@ AFRAME.registerComponent('managed-street', {
...
+      });
+    }
+
+    // Insert the segment at the specified index in the DOM
+    if (index === this.managedEntities.length) {
+      this.el.appendChild(segmentEl);
+    } else {
+      const referenceNode = this.managedEntities[index];
+      this.el.insertBefore(segmentEl, referenceNode);
+    }
```
vincentfretin, line 117: https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore#referencenode
_referenceNode_
The node before which newNode is inserted. If this is null, then newNode is inserted at the end of node's child nodes.
So instead of 
```js
    if (index === this.managedEntities.length) {
      this.el.appendChild(segmentEl);
    } else {
      const referenceNode = this.managedEntities[index];
      this.el.insertBefore(segmentEl, referenceNode);
    }
```
you can do:
```js
    const referenceNode = this.managedEntities[index] ?? null;
    this.el.insertBefore(segmentEl, referenceNode);
```

This is what we do already in
https://github.com/3DStreet/3dstreet/blob/172d36051e011365c21fc8a22310b47bd6d88ed2/src/editor/lib/commands/EntityRemoveCommand.js#L37-L39

kfarr: 1781d2b3baafc7d9c2d5ad919a2e0bab5acc04c7

PR #991, opened 2024-12-26
detaching clones

kfarr:
to test detaching, after adding a managed street, run in console: `document.querySelector('[street-generated-clones__0]').components['street-generated-clones__0'].detach()`

right now it's difficult to select the desired component of the desired segment. some ideas:
- add autocreated / Autocreated managed entity. Edit street segment settings, or convert to manual; this is shown on the entity of the 
- function to get parent entity/component, or display link to component settings on the street segment sidepanel
- in the parent segment have a button to convert to manual -- on the component level?


src/components/street-generated-stencil.js
```diff
@@ -99,6 +99,40 @@ AFRAME.registerComponent('street-generated-stencil', {
...
+        );
+      }
+    });
+
+    // Clear the createdEntities array without removing the actual entities
+    this.createdEntities.length = 0;
+
+    // Remove the component from the parent element
+    parentEl.removeAttribute(this.attrName);
+  },
```
vincentfretin, line 135: For that detach action to be undoable, I would keep the component and use a detached boolean property. In update if it changed to detached true, do what you do here without removing the component, and if it changed to detached false, do the reverse getting all entities with `[data-parent-component="${this.attrName}"]`. For the scene load to work properly, if detached is true, don't create entities and return.
For the detach button do `AFRAME.INSPECTOR.execute('entityupdate', {entity, component, property, value: {detached:true})`

kfarr: good idea i'll try that

kfarr: I tried this out and it is possible but is also complicated and requires a lot of extra condition checking in `update`. i may keep this as-is with no undo support with the idea that this needs to be done differently in the future. i could add a user warning.

here's another idea of an approach:
- when "detaching" any component, instead:
- create new clones of each under same parent
- remove component (perhaps component removal an undo-able command?)
- then there is a series of actions: create clones, remove component, select the clone of the previously selected object

vincentfretin: Yes, you could indeed use my multi command like this
```js
const multiCommands = [];
multiCommands.push(['componentremove', this.attrName]);
let entityObjToPushAtTheEnd = null; // so that the entity is selected after executing the multi command
this.createdEntities.forEach((entity) => {
  const entityObj = {
    mixin: entity.getAttribute('mixin'),
    'data-layer-name': entity
      .getAttribute('data-layer-name')
      .replace('Cloned Model', 'Detached Model'),
    components: {
      position: entity.getAttribute('position'),
      rotation: entity.getAttribute('rotation')
    }
  };
  if (AFRAME.INSPECTOR?.selectedEntity === entity) {
    entityObjToPushAtTheEnd = entityObj;
  }
  if (entityObjToPushAtTheEnd == null) {
    multiCommands.push(['entitycreate', entityObj]);
  }
});
if (entityObjToPushAtTheEnd !== null) {
  multiCommands.push(['entitycreate', entityObjToPushAtTheEnd]);
}
AFRAME.INSPECTOR.execute('multi', multiCommands);
```

vincentfretin: multi command in #995

kfarr:
see https://github.com/3DStreet/3dstreet/pull/996 instead

Issue #166, opened 2022-05-07
different colors for vehicles


burritojustice:
I apologize in advance for my ignorance about how 3D models work, but it would be nice if we had a way to change the color of a vehicle other than the default grey.

Could a random color be selected when the 3Dstreet view is loaded? Could a car color be selected in the editor?

Car colors are shockingly boring these days: https://blog.consumerguide.com/boring-car-colors/

but maybe we could select from 

- black
- white
- grey
- darkish blue
- red
- green (crazy, I know)

kfarr:
example glitch code that changes a material inside gltf https://glitch.com/edit/#!/grateful-sweltering-meter

kfarr:
@gnansai in addition to the removing rigging experiment, we would also like to be able to change vehicle colors dynamically.

Therefore, would it make sense to consider alternate coloring mechanism for vehicles instead of UV map? Such as vertex coloring or defining colors through material.

A good example to refer to is https://github.com/diarmidmackenzie/instanced-mesh/tree/main/examples/gltf-models note the gltf file whose color is modified in the various instances via the component

gnansai:
@kfarr 
I checked the example "https://github.com/diarmidmackenzie/instanced-mesh/tree/main/examples/gltf-models". 
The above example randomizes the first material colors of the car model.
![image](https://github.com/3DStreet/3dstreet/assets/25223402/497db419-5ff2-45b8-abc1-1d975d7f253a)
![image](https://github.com/3DStreet/3dstreet/assets/25223402/68739f51-997f-4fae-9868-d0995fca51fa)

If I reorder the material in blender and put the window material at first, The code randomized the window color.
![image](https://github.com/3DStreet/3dstreet/assets/25223402/ad3d5646-57d8-4b5c-870a-269f22d9bdec)
![image](https://github.com/3DStreet/3dstreet/assets/25223402/33e1bea9-8a9e-426d-b084-a02176080682)

gnansai:
I tried with our Isuzu truck model. I selected the few parts of the vehicle and assigned a new material "car-body". The example code randomizes the colors of "car-body" material.
![image](https://github.com/3DStreet/3dstreet/assets/25223402/6d0c5b81-87e6-4cc1-821f-b88f705cbdb4)
![image](https://github.com/3DStreet/3dstreet/assets/25223402/19a6874f-79bf-4244-8b99-87747fa59d70)

Algorush:
[...]

PR #992, opened 2024-12-27
[WIP] Adding color options to entities

ewei2406:
Based on #166, I was playing around with options for adding color. Here's a basic implementation of a simple approach to just modify the contents of meshes `object3D` to have a `color` property on their materials.

- Since each entity has multiple meshes it is hard to determine which one corresponds with which part (i.e. the Sedan has one mesh for the body, one mesh for the windows), I left an `index` field so that the user can specify which mesh to change the color for (though this requires a bit of guessing, and leave it to `-1` for no color applied).
- The data for the coloring is saved in the `accent-color` component which has the `color` and `index` values, and is grouped under the "advanced" section.
- The `color` field doesn't really work as a true color change but more of a tint. So it isn't exactly ideal, but it works for all entities without any changes to the model files.

Screenshot 2024-12-27 at 12 03 41 AM
 
Screenshot 2024-12-26 at 11 58 55 PM

### DIFF
--- a/src/components/accent-color.js
+++ b/src/components/accent-color.js
@@ -11,12 +11,10 @@ AFRAME.registerComponent('accent-color', {
   },
   init: function () {
     this.meshes = [];
-    console.log(this.el);
     // The group children aren't ready when the component is initialized, so wait a bit.
     // TODO: Find a better way to time this.
     setTimeout(() => {
       this.el.object3D.traverse((o) => o.material && this.meshes.push(o));
-      console.log(this.meshes);
       this.update();
     }, 1000);
   },
--- a/src/components/accent-color.js
+++ b/src/components/accent-color.js
@@ -10,24 +10,23 @@ AFRAME.registerComponent('accent-color', {
     }
   },
   init: function () {
-    this.meshes = [];
-    // The group children aren't ready when the component is initialized, so wait a bit.
-    // TODO: Find a better way to time this.
-    setTimeout(() => {
-      this.el.object3D.traverse((o) => o.material && this.meshes.push(o));
-      this.update();
-    }, 1000);
+    // Want the color update to apply when the mixin is loaded,
+    // which may be at a different time than the component
+    this.el.addEventListener('model-loaded', () => this.update());
   },
   update: function (oldData) {
+    // Collect meshes
+    const meshes = [];
+    this.el.object3D.traverse((o) => o.material && meshes.push(o));
     // Remove the tint on the previous mesh
     if (oldData && oldData.index !== this.data.index) {
-      const oldMesh = this.meshes.at(oldData.index);
+      const oldMesh = meshes.at(oldData.index);
       oldMesh && oldMesh.material.color.set(1, 1, 1);
     }
     // Ignore negative index
     if (this.data.index < 0) return;
     // Apply tint to selected mesh
-    const mesh = this.meshes.at(this.data.index);
+    const mesh = meshes.at(this.data.index);
     if (mesh) {
       mesh.material.color.set(this.data.color);
       mesh.material.color.multiplyScalar(255);
--- a/src/components/accent-color.js
+++ b/src/components/accent-color.js
@@ -1,0 +1,38 @@
+/* global AFRAME */
+
+AFRAME.registerComponent('accent-color', {
+  schema: {
+    color: {
+      type: 'color'
+    },
+    index: {
+      type: 'int'
+    }
+  },
+  init: function () {
+    this.meshes = [];
+    console.log(this.el);
+    // The group children aren't ready when the component is initialized, so wait a bit.
+    // TODO: Find a better way to time this.
+    setTimeout(() => {
+      this.el.object3D.traverse((o) => o.material && this.meshes.push(o));
+      console.log(this.meshes);
+      this.update();
+    }, 1000);
+  },
+  update: function (oldData) {
+    // Remove the tint on the previous mesh
+    if (oldData && oldData.index !== this.data.index) {
+      const oldMesh = this.meshes.at(oldData.index);
+      oldMesh && oldMesh.material.color.set(1, 1, 1);
+    }
+    // Ignore negative index
+    if (this.data.index < 0) return;
+    // Apply tint to selected mesh
+    const mesh = this.meshes.at(this.data.index);
+    if (mesh) {
+      mesh.material.color.set(this.data.color);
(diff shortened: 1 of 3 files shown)


src/components/accent-color.js
```diff
@@ -0,0 +1,36 @@
...
+    },
+    index: {
+      type: 'int'
+    }
+  },
+  init: function () {
+    this.meshes = [];
+    // The group children aren't ready when the component is initialized, so wait a bit.
+    // TODO: Find a better way to time this.
+    setTimeout(() => {
```
ewei2406, line 16: This is a workaround since the entity children aren't visible when the component is initialized. I'm curious which event is fired that can be used to time this correctly.

vincentfretin: `gltf-model` component fires a `model-loaded` event, so you probably want to listen to that.

```js
if (this.el.getObject3D('mesh')) {
  // do your thing
} else {
  this.el.addEventListener('model-loaded', () => {
    this.meshes.length = 0; // be sure to not keep previous gtlf meshes if we switched the model
    // do your thing
  })`
}
```

I wouldn't keep meshes in an array on the component, that's source of issues, for example if you remove the gtlf-model component, you currently keep meshes in the array that aren't garbage collected. I would just do the traverse in update, without storing an array of meshes on the component.

ewei2406: Sounds good, I tried the `model-loaded` and it works perfect!

vincentfretin: The `this.el.getObject3D('mesh')` check is needed if the accent-color component is added later, so after you got that model-loaded event. gltf-model set the loaded glb with `el.setObject3D('mesh', model)`

src/editor/components/components/AddLayerPanel/AddLayerPanel.component.jsx
```diff
@@ -145,7 +145,11 @@ const createEntity = (mixinId) => {
   }
   const newEntityObject = {
     mixin: mixinId,
-    components: {}
+    components: {
+      'accent-color': {
+        index: -1
+      }
+    }
```
vincentfretin, line 152: I wouldn't add the component by default, that's not needed unless you use it. I would create a button in the Sidebar for example "customize color" (for mixins that makes sense) that adds the component to the entity. You can do that with `AFRAME.INSPECTOR.execute('componentadd', {entity: entity, component: 'accent-color'})`

src/components/accent-color.js
```diff
@@ -0,0 +1,35 @@
...
+      const oldMesh = meshes.at(oldData.index);
+      oldMesh && oldMesh.material.color.set(1, 1, 1);
+    }
+    // Ignore negative index
+    if (this.data.index < 0) return;
+    // Apply tint to selected mesh
+    const mesh = meshes.at(this.data.index);
+    if (mesh) {
+      mesh.material.color.set(this.data.color);
+      mesh.material.color.multiplyScalar(255);
```
vincentfretin, line 32: Why do you multiply by 255?

vincentfretin:
In the UI, the index field is not really understandable by the user. I would dynamically get all material names of the model and show a select here. Not sure what are the names currently in the models, it may be nice to rename them in each model to have something clean.
It would be even better if you can make the component `multiple: true` so you can add several to change the colors of different materials. See my more generalist `material-values` component here as an example https://github.com/akbartus/A-Frame-Component-GLTF-Manipulator/issues/2

kfarr:
* In the ideal world, each vehicle has named nodes explicitly for this purpose such as "body color" and someday perhaps also an "accent color"
* We target updates specifically to this node using a method like Vincent's suggestion
* I would suggest inspecting a [few of our models in Blender](https://github.com/3DStreet/3dstreet-assets-source/tree/main/models/vehicles) and/or a tool like https://playcanvas.com/model-viewer for the glbs. In my 1 minute review it looks like the compressed GLB have not retained any specific named nodes that can be addressed in the file. In blender source files, some models like isuzu truck have body color, but many such as prius or waymo do not.
* So likely we need to modify the assets to facilitate this feature.

ewei2406:
Hmm, I agree - the use of an index is unintuitive to most users, but since many assets don't support vertex colorings nor have names for meshes the scope of this will have to include changes to the model files themselves. Then these are the steps I will take

1. Give all meshes a name by modifying their blender file
 - A quicker alternative could be to create a mapping of mesh names, like `catalog.json`, since it seems that the order of meshes within each entity is always the same. But I think this isn't a great long-term solution to have the names be in a separate file.
2. Create the component that saves the mapping of meshes and their colors (if applied)
 - On the editor side, maybe we could have a dropdown selection to pick which mesh, then a boolean for whether there is a custom color applied, and then if the boolean is true, a field for custom color. Sorta like this: 
menu

What do you think?

vincentfretin:
You're both talking about mesh names. I was talking about material names. Some models can have several meshes that uses the same material, so that doesn't make sense to me to chose a mesh. Tools like gltf-transform lose the mesh names, but keep the material name if you add `--palette false` option otherwise it merges simple materials into a single palette material, not what you want when you want the colors to be configurable.

kfarr:
Thanks for the clarification Vincent. It may make sense to have the original artist help with the conversion, perhaps we could put a list together of which files and what are the modifications we want. I can then get a quote and see if it's something we can afford in the short term or if we should do it ourselves.

Unrelated but part of the same files, we also discussed earlier removing rigging for wheels and instead using named meshes for wheel to allow rotation when animating. (rigging was overkill)

kfarr:
Another idea @ewei2406 , the Isuzu truck model in blender appears to have the named material "vehicle-body" already in a way that we might want. Perhaps you could experiment with exporting in compressed glb but in a way that retains that material name so we can target it in the color component. If it can work in this one model then we can apply the same changes to other vehicle models?

image

https://github.com/3DStreet/3dstreet-assets-source/blob/main/models/vehicles/Isuzu_truck_v01.blend

ewei2406:
Sounds good, I will try to draft up what is discussed. I will take a look at using blender to modify the meshes - I was able to transform the `prius-sedan` model to use a solid color material for the car body instead of a UV map.

gnansai:
Hi @ewei2406 
I have made the 3dStreet vehicle models in Blender. I see that you are trying to add the separate material "vehicle-body" to the vehicles in Blender. Do let me know if you need any help with blender and exporting exporting glb files from blender.

ewei2406:
See #1005

PR #993, opened 2024-12-27
Intersection crosswalk options

ewei2406:
Towards having intersections with crosswalk options, this PR changes the crosswalk design of the intersection to use a selection instead of a boolean and adds the new crosswalks requested by Marc: 
> old school 2 lines, rainbow option, piano key option, mural option, raised (in particular on a side street) with color or emphasis. 
- The available options are: no crosswalk, normal zebra stripe, double stripe, rainbow, piano keys, mural, or raised.

## Implementation
- The crosswalk type is stored in the `intersection` component as a number that is mapped to the corresponding crosswalk type in the object `CROSSWALKS` defined in `intersection.js`
- Since each crosswalk entity/mixin has different size and orientation, there is a special `CROSSWALK_TRANSFORMS` object that defines a unique transformation that each crosswalk entity can have, so that they can all be oriented correctly.
- Currently, the options for a crosswalk are: none, raised crosswalk, original crosswalk, and crosswalk-box entity.
- The textures for the new crosswalks are in the PR for the assets repo: https://github.com/3DStreet/3dstreet-assets-dist/pull/36
- Renamed the existing mixin for `street-element-crosswalk-raised` to just `crosswalk-raised` for consistency with other crosswalks.

Screenshot 2024-12-27 at 4 00 21 PM
Screenshot 2024-12-27 at 4 10 58 PM

### DIFF
--- a/src/assets.js
+++ b/src/assets.js
@@ -218,6 +218,10 @@ function buildAssetHTML(assetUrl, categories) {
         <a-mixin shadow id="street-element-traffic-island" scale="1.5 1.5 1.5" rotation="0 0 0" gltf-model="url(${assetUrl}sets/uoregon/gltf-exports/draco/curb-traffic-island.glb)"></a-mixin>
         <a-mixin shadow id="street-element-speed-hump" scale="1.5 1.5 1.5" rotation="0 0 0" gltf-model="url(${assetUrl}sets/uoregon/gltf-exports/draco/speed-hump.glb)"></a-mixin>
         <a-mixin shadow id="crosswalk-zebra-box" geometry="primitive: box; height: 0.1; width: 2; depth: 10" material="src: url(${assetUrl}materials/markings-crosswalk.png)"></a-mixin>
+        <a-mixin shadow id="crosswalk-rainbow" geometry="primitive: plane; width:2; height:12; skipCache: true;" material="src:${assetUrl}materials/crosswalk-rainbow.png; transparent: true;"></a-mixin>
+        <a-mixin shadow id="crosswalk-double" geometry="primitive: plane; width:2; height:12; skipCache: true;" material="src:${assetUrl}materials/crosswalk-double.png; transparent: true;"></a-mixin>
+        <a-mixin shadow id="crosswalk-mural" geometry="primitive: plane; width:2; height:12; skipCache: true;" material="src:${assetUrl}materials/crosswalk-mural.png; transparent: true;"></a-mixin>
+        <a-mixin shadow id="crosswalk-piano" geometry="primitive: plane; width:2; height:12; skipCache: true;" material="src:${assetUrl}materials/crosswalk-piano.png; transparent: true;"></a-mixin>
         <a-mixin shadow id="traffic-calming-bumps" gltf-model="url(${assetUrl}sets/uoregon/gltf-exports/draco/traffic-calming-bumps.glb)"></a-mixin>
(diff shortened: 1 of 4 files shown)


ewei2406:
While working on this PR, I noticed a bug with the sidewalk width parameter where the sidewalk does not correctly align with the intersection if the intersection width/length is not the default value; I'll draft a fix in a separate PR.

src/components/intersection.js
```diff
@@ -1,5 +1,40 @@
...
+  'crosswalk-piano': 6
+};
+export const CROSSWALKS_REV = {};
+Object.keys(CROSSWALKS).forEach((key) => {
+  CROSSWALKS_REV[CROSSWALKS[key]] = key;
+});
+
+const transformImageCrosswalk = (entity, length, rotZ) => {
+  entity.setAttribute('rotation', { z: rotZ });
+  entity.setAttribute('scale', { y: length / 12, x: 1.5 });
```
vincentfretin, line 19: Please define all three x y z. Although that somehow works on aframe 1.5.0, setting a partial vec3 is not a valid usage and sometimes won't work properly (setting missing values to 0 and not 1 as you expect for scale) on aframe 1.6.0 or later depending on if you insert the entity in DOM first.

src/components/intersection.js
```diff
@@ -1,5 +1,40 @@
...
+
+const CROSSWALK_TRANSFORMS = {
+  none: function () {},
+  'crosswalk-zebra': (entity, length, rotZ) => {
+    entity.setAttribute('rotation', { z: rotZ });
+    entity.setAttribute('scale', { y: length / 12 });
+  },
+  'crosswalk-raised': (entity, length, rotX) => {
+    entity.setAttribute('rotation', { x: rotX, y: 90, z: 90 });
+    entity.setAttribute('scale', { x: length / 7, z: 1.5 });
```
vincentfretin, line 30: Same here please define full the vec3 with y: 1

src/components/intersection.js
```diff
@@ -1,5 +1,40 @@
...
+const transformImageCrosswalk = (entity, length, rotZ) => {
+  entity.setAttribute('rotation', { z: rotZ });
+  entity.setAttribute('scale', { y: length / 12, x: 1.5 });
+};
+
+const CROSSWALK_TRANSFORMS = {
+  none: function () {},
+  'crosswalk-zebra': (entity, length, rotZ) => {
+    entity.setAttribute('rotation', { z: rotZ });
+    entity.setAttribute('scale', { y: length / 12 });
```
vincentfretin, line 26: Same here please define full the vec3 with x: 1, z: 1

src/assets.js
```diff
@@ -211,13 +211,17 @@ function buildAssetHTML(assetUrl, categories) {
         <a-mixin shadow id="temporary-traffic-cone" scale="1 1 1" rotation="0 0 0" gltf-part="src: #dividers; part: traffic-cone"></a-mixin>
         <a-mixin shadow id="temporary-jersey-barrier-plastic" scale="1 1 1" rotation="0 0 0" gltf-part="src: #dividers; part: jersey-barrier-plastic"></a-mixin>
         <a-mixin shadow id="temporary-jersey-barrier-concrete" scale="1 1 1" rotation="0 0 0" gltf-part="src: #dividers; part: jersey-barrier-concrete"></a-mixin>
-        <a-mixin shadow id="street-element-crosswalk-raised" scale="1 1 1" rotation="0 0 0" gltf-model="url(${assetUrl}sets/uoregon/gltf-exports/draco/crosswalk-raised.glb)"></a-mixin>
```
kfarr, line 214: we should probably keep the original id because there may be users who have existing scenes with street-element-crosswalk-raised id

PR #994, opened 2024-12-28
Fix intersection sidewalk being in wrong position for intersections with non-default sizes

ewei2406:
fixes incorrect width and depth order.

Before:
Screenshot 2024-12-27 at 4 30 09 PM

After:
Screenshot 2024-12-27 at 4 38 00 PM


vincentfretin:
zebra are missing in your after screenshot, is is intended?

ewei2406:
Good catch! These are separate scenes and I forgot to set the markings to be the same. This won't affect the stripes appearing.

PR #995, opened 2024-12-28
Add multi command to be able to undo a sequence of actions in one step

vincentfretin:
Usage:
```js
const commands: Commands = [
  ["entitycreate", {position: "0 0 0", components: {"gltf-model": "model.glb"}],
  ["entitycreate", {position: "1 0 0", components: {"gltf-model": "model.glb"}],
];
AFRAME.INSPECTOR.execute("multi", commands)
```

To create a parent with children in one go, passing parentEl the id of the parent:

```js
const parentEntityId = createUniqueId();
const definition = {
  id: parentEntityId,
  components: {
    position: "0 0 0",
    "gltf-model": "parent.glb",
  },
};
const commands: Commands = [];
commands.push(["entitycreate", definition]);
commands.push(["entitycreate", {parentEl: parentEntityId, components: {position: "0 1 0", "gltf-model": "child.glb"}]]);
const afterCreate = (entity: Entity | null) => {
  if (!entity) return; // actually not possible, but typescript doesn't know that
  textureSelectorForceUpdate();
};
AFRAME.INSPECTOR.execute("multi", commands, undefined, afterCreate);
```

Types:
```js
type ComponentAddCommand = [
  "componentadd",
  {
    entity: Entity;
    component: string;
    value: string;
  },
];

type ComponentRemoveCommand = [
  "componentremove",
  {
    entity: Entity;
    component: string;
  },
];

type EntityCreateCommand = ["entitycreate", EntityObject] | ["entitycreate", EntityObject, (el: Entity) => void];

type EntityRemoveCommand = ["entityremove", Entity];

type EntityUpdateCommand = [
  "entityupdate",
  {
    entity: Entity;
    component: string;
    property?: string | undefined;
    value: string;
  },
];

type Commands = (
  | ComponentAddCommand
  | ComponentRemoveCommand
  | EntityCreateCommand
  | EntityRemoveCommand
  | EntityUpdateCommand
)[];
```

PR #996, opened 2024-12-28
Detach with multi command

kfarr:
based on this: https://github.com/3DStreet/3dstreet/pull/995

then add detaching ability as described in this conversation: https://github.com/3DStreet/3dstreet/pull/991/files#r1898396041

this could be a function stored in sidebar for now, triggered upon convert to manual button

```
const commands = [];
commands.push([
  'componentremove',
  { entity: this.el, component: this.attrName }
]);
let entityObjToPushAtTheEnd = null; // so that the entity is selected after executing the multi command
this.createdEntities.forEach((entity) => {
  const position = entity.getAttribute('position');
  const rotation = entity.getAttribute('rotation');
  const entityObj = {
    parentEl: this.el, // you can also put this.el.id here that way the command is fully json serializable but el currently doesn't have an id
    mixin: entity.getAttribute('mixin'),
    'data-layer-name': entity
      .getAttribute('data-layer-name')
      .replace('Cloned Model', 'Detached Model'),
    components: {
      position: { x: position.x, y: position.y, z: position.z },
      rotation: { x: rotation.x, y: rotation.y, z: rotation.z },
    }
  };
  if (AFRAME.INSPECTOR?.selectedEntity === entity) {
    entityObjToPushAtTheEnd = entityObj;
  }
  if (entityObjToPushAtTheEnd == null) {
    commands.push(['entitycreate', entityObj]);
  }
});
if (entityObjToPushAtTheEnd !== null) {
  commands.push(['entitycreate', entityObjToPushAtTheEnd]);
}
AFRAME.INSPECTOR.execute('multi', commands);
```


vincentfretin:
Oh nice it worked on the first try, I edited it 2 or 3 times but I didn't even test the code in the description. :D

PR #998, opened 2024-12-29
striping first cut

kfarr:
old / new

image

### DIFF
--- a/src/components/street-segment.js
+++ b/src/components/street-segment.js
@@ -252,6 +252,19 @@ AFRAME.registerComponent('street-segment', {
         });
       });
     }
+
+    if (componentsToGenerate?.striping?.length > 0) {
+      componentsToGenerate.striping.forEach((stripe, index) => {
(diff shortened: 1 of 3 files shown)


src/components/street-segment.js
```diff
@@ -252,6 +252,19 @@ AFRAME.registerComponent('street-segment', {
...
       });
     }
+
+    if (componentsToGenerate?.striping?.length > 0) {
+      componentsToGenerate.striping.forEach((stripe, index) => {
+        this.el.setAttribute(`street-generated-striping__${index}`, {
+          striping: stripe.striping,
+          segmentWidth: this.data.width,
+          length: this.data.length,
+          positionY: stripe.positionY || 0.05, // Default to 0.05 if not specified
```
vincentfretin, line 262: You'll have a bug if you want positionY 0 here. For default value use the `??` operator instead of `||`.

kfarr: thanks

PR #999, opened 2024-12-31
vertex ai barely working

kfarr:
output is json but is not parsed, only returns text

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "classnames": "^2.3.2",
         "clipboard": "^2.0.11",
         "date-fns": "^2.30.0",
-        "firebase": "^10.9.0",
+        "firebase": "^11.1.0",
         "firebase-admin": "^12.1.1",
         "firebase-functions": "^5.0.1",
         "lodash-es": "^4.17.21",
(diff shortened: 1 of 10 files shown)


src/editor/components/scenegraph/AIChatPanel.js
```diff
@@ -0,0 +1,374 @@
...
+
+      Please analyze the request and provide one of the following:
+      1. If the user is asking about the scene, provide a natural language explanation
+      2. If the user is asking to modify the scene, use the entityUpdate function
+      3. If the user needs help, provide relevant guidance about the 3DStreet editor
+
+      In the scene state, units for length are in meters, and rotations are in degrees.
+
+      The orientation of axis to cardinal directions is as follows: x+ (positive) is north; x- (negative) is south; y+ (positive) is up; y- (negative) is down; z- (negative) is west; z+ (positive) is east;
+      Models face z- (east) when at 0º Y rotation.
```
vincentfretin: z+

src/editor/components/scenegraph/AIChatPanel.js
```diff
@@ -0,0 +1,374 @@
...
+      The orientation of axis to cardinal directions is as follows: x+ (positive) is north; x- (negative) is south; y+ (positive) is up; y- (negative) is down; z- (negative) is west; z+ (positive) is east;
+      Models face z- (east) when at 0º Y rotation. 
+
+      Make sure you convert everything to the appropriate units, even if the user uses different units.
+
+      IMPORTANT: When you need to calculate a value (like "5 - 2"), return it as a string expression ("5 - 2") in a parameter named "expression-for-value"
+
+      When changing a model, use the "entityupdate" command with the following payload:
+      {
+        "entityId": "n9eLgB9C635T_edXuXIgz",
```
vincentfretin, line 251: entity-id

src/editor/lib/entity.js
```diff
@@ -22,20 +22,37 @@ import {
...
+  // Special handling for vector components (position, rotation, scale)
+  const vectorComponents = ['position', 'rotation', 'scale'];
+
+  if (property && vectorComponents.includes(component)) {
+    // Get current values
+    const currentValues = entity.getAttribute(component) || {};
+
+    if (value === null || value === undefined) {
+      // Remove property
+      delete currentValues[property];
```
vincentfretin, line 34: You will have issue again if you set a partial xyz in aframe.
If you want to use value null or undefined to reset a coordinate, then set it to 0 (position and rotation) or 1 (scale) instead of removing.

src/editor/components/scenegraph/AIChatPanel.js
vincentfretin: It's better not to confuse the AI :D

kfarr: aca44d9ac6f302863352e6be5fd1b78b86d11337

src/editor/lib/entity.js
kfarr: decided to remove and not handle that case as it's not supposed to happen, better to throw an error
7f4cad1e20c457ec30591dd0d5eea1ae47bd988c

src/editor/components/scenegraph/AIChatPanel.js
kfarr: 7f4cad1e20c457ec30591dd0d5eea1ae47bd988c

kfarr:
As a user who wishes to modify their scene using AI help, I'd like to modify the following
- I'd like to adjust the street layout, length, segments, perhaps other properties
- i'd like to edit geospatial location, provider, other settings
- i'd like to edit environment component

PR #1000, opened 2025-01-01
mvp polygon offset

kfarr:
allows us to use logdepthbuffer = false, therefore splats in production!

other things to consider later:
- intersection striping also needs polygon offset
- vehicles or building models with excessive z fighting need to be fixed separately
- should this be changed to a component instead of manually applied? therefore this would be able to persist even if the object is detached from the generation component?


kfarr:
see https://github.com/3DStreet/3dstreet/pull/1004

PR #1004, opened 2025-01-02
mvp polygon offset

kfarr:
allows us to use logdepthbuffer = false, therefore splats in production!

other things to consider later:
- intersection striping (crosswalk) also needs polygon offset
- vehicles or building models with excessive z fighting need to be fixed separately
- should this be changed to a component instead of manually applied? therefore this would be able to persist even if the object is detached from the generation component?


kfarr:
testing with existing scenes -- this won't work very well. As a user I may have placed many manual objects and I expect them to show up correctly.

instead -- there may need to be an explicit flag such as ?splat=true that will turn on standard depth buffer

this pull will help improve visual quality for splat cases

PR #1008, opened 2025-01-07
labels with manage-street event dispatcher

kfarr:
some questions / thoughts before merging:
- should this respect alignment?
- what options should a user be able to set relating to label design? should there be an option for labels to be high up and also below?
- how are labels enabled / disabled by user? as an option on managed-street? or by adding/removing the component?

image


kfarr:
takes some fiddling with rotation x and z offset, but you can make nice labels for ortho (plan view) mode too
image

kfarr:
other requested features:
- imperial / metric units switching -- is this at the scene level? user level? street level?
- how does a user enable/disable? enabled flag on component, remove property on managed-street?

PR #1012, opened 2025-01-08
edit 3d tiles opacity and blending mode

kfarr:


src/components/blending-opacity.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+  },
+
+  tick: function () {
+    // Only update materials if there are changes
+    // if (this.el.getAttribute('tiles-material-change').needsUpdate) {
+    //   this.updateMaterials();
+    //   this.el.getAttribute('tiles-material-change').needsUpdate = false;
+    // }
+    this.updateMaterials();
```
vincentfretin, line 143: That's expensive to keep doing a traverse on each tick, also it's creating an array each time `[node.material]`.

kfarr: You're right, in obb-clipping this is throttled, I can do that again.

kfarr: Unfortunately the result looks very bad if throttled, materials appear to "flicker" as they wait for the throttle to be adjusted to a different blending.

For now I'll keep it, but there are 2 options to fix:
1. hook into the material post processing event in the existing 3d tiles library
2. consider switching to better maintained library instead of optimizing this one, hook into its material post processing event https://github.com/NASA-AMMOS/3DTilesRendererJS?tab=readme-ov-file#custom-material (blocked by a-frame / three.js upgrade -- needs v166)

kfarr:
the default should not apply this unless a user selects it. an idea:
- add new boolean to street-geo, blendingEnabled default false
- if false, don't add or modify any blending-opacity component values
- if set to true and was previously false, add blending-opacity component, allow modify of blending-opacity component values
- if set to false and was previously true, delete and re-add the google tiles without the component?

src/components/street-geo.js
```diff
@@ -76,13 +92,46 @@ AFRAME.registerComponent('street-geo', {
...
+          // If blending was previously enabled and now disabled, recreate the tiles
+          const currentEl = this.google3d;
+          this.el.removeChild(currentEl);
+          this.google3d = null;
+          this.google3dCreate();
+        }
+      }
+
+      // regenerate tiles if blending enabled in editor
+      const dataDiff = AFRAME.utils.diff(oldData, data);
```
vincentfretin, line 123: You already have `updatedData` at the beginning of `update()`, you can reuse it.

src/components/street-geo.js
```diff
@@ -76,13 +92,46 @@ AFRAME.registerComponent('street-geo', {
...
         this.google3d.removeAttribute('obb-clipping');
       }
+
+      // Handle blending updates
+      if (data.blendingEnabled) {
+        if (data.blendMode) {
+          this.google3d.setAttribute(
+            'blending-opacity',
+            this.returnBlendMode(data.blendMode)
+          );
```
vincentfretin, line 109: I think you can put here
```js
      if (oldData.blendingEnabled === false) {
        const currentEl = this.google3d;
        this.el.removeChild(currentEl);
        this.google3d = null;
        this.google3dCreate();
      }
```
for the case where you changed `blendingEnabled` from false to true instead of doing your check `changedProps.length === 1 && changedProps.includes('blendingEnabled')` below.

src/components/blending-opacity.js
```diff
@@ -0,0 +1,144 @@
...
+          });
+        } else {
+          this.updateMaterial(obj.material);
+        }
+      }
+    });
+  },
+
+  updateMaterial: function (material) {
+    if (!material) return;
```
vincentfretin, line 120: As far as I know, it's mandatory to have a material on a mesh, so I don't think this is needed. Did you have a undefined here?

PR #1014, opened 2025-01-11
Fix scene save right after creation

### LINKED ISSUE #1013: scenes autosaving from streetmix result in blank screenshot
As a user importing a street from streetmix and the street is automatically saved (per [previous PR](https://github.com/3DStreet/3dstreet/pull/957): specific file is `src/index.js`) then I see an error

image

This attempted fix https://github.com/3DStreet/3dstreet/pull/1009 removed the error from user and properly saves the scene as expected, however now the screenshot is not created and results in a black box being saved as the thumbnail for most scenes

image

vincentfretin:
This closes #1013 fixing several issues: duplicate scene saved, title not saved, positions not saved, blank screenshot

### DIFF
--- a/src/components/streetplan-loader.js
+++ b/src/components/streetplan-loader.js
@@ -2,8 +2,6 @@
 import useStore from '../store.js';
 var streetplanUtils = require('../streetplan/streetplan-utils.js');
 
-const state = useStore.getState();
-
 AFRAME.registerComponent('streetplan-loader', {
   dependencies: ['street'],
   schema: {
@@ -34,6 +32,7 @@ AFRAME.registerComponent('streetplan-loader', {
       // Update layer name with project and street names
       el.setAttribute('data-layer-name', `StreetPlan • ${streetplanName}`);
 
+      const state = useStore.getState();
       if (!state.sceneTitle) {
         state.setSceneTitle(streetplanName);
       }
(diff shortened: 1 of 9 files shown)


src/components/streetplan-loader.js
```diff
@@ -2,8 +2,6 @@
 import useStore from '../store.js';
 var streetplanUtils = require('../streetplan/streetplan-utils.js');
 
-const state = useStore.getState();
```
vincentfretin, line 5: Be careful `useStore.getState()` returns a snapshot of the state.

src/editor/lib/SceneUtils.js
```diff
@@ -171,7 +189,9 @@ export async function saveScene(currentUser, doSaveAs) {
 export async function saveSceneWithScreenshot(currentUser, doSaveAs) {
   const currentSceneId = await saveScene(currentUser, doSaveAs);
   if (currentSceneId) {
-    makeScreenshot();
+    // wait a bit for models to be loaded, may not be enough...
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
vincentfretin, line 193: I didn't have any issue locally, but I added that to wait a bit for models to be all loaded before taking the screenshot.

PR #1017, opened 2025-01-14
Manage street - simpler UI

kfarr:


vincentfretin (commented):
I think you need to also remove managed-street width property?

PR #1022, opened 2025-01-14
Fix clone nan

kfarr:


vincentfretin:
I did additional fixes while comparing the lines we changed in our first attempt in #842
If all is working for you, you can merge.

PR #1024, opened 2025-01-15
show calculated width

kfarr:
doesn't work, the value is not updated

### DIFF
--- a/src/editor/components/components/ManagedStreetSidebar.js
+++ b/src/editor/components/components/ManagedStreetSidebar.js
@@ -29,6 +29,10 @@ const ManagedStreetSidebar = ({ entity }) => {
                   isSingle={false}
                   entity={entity}
                 />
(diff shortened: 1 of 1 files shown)


vincentfretin:
yeah that part of the react ui updates only if there is an entityupdate, componentremove, or componentadd event, or selected entity changed
I guess you could run
```js
const selectedEntity = AFRAME.INSPECTOR.selectedEntity;
if (selectedEntity === this.el) {
  AFRAME.INSPECTOR.selectEntity(null)
  AFRAME.INSPECTOR.selectEntity(selectedEntity);
}
```
from manage-street component to force an update, that's not super clean but meh

vincentfretin:
Or reintroduce an refreshsidebar event we had previously I think and run `Events.emit('refreshsidebar')` from the component.

PR #1026, opened 2025-01-15
Style updates

kfarr:
- add icons to primary layers
- update geospatial layer styles
- set same height all scene graph entity panels
- remove qr code from geo settings
- update manage street icons
- make property rows 1 line for model, position, rotation, scale


kfarr:
street segment panel before / after
image

layers panel before / after
image

PR #1036, opened 2025-01-23
streetplan building support wip

kfarr:


kfarr:
already supported in different PR

PR #1037, opened 2025-01-24
Sidebar UI for street segment components

kfarr:


kfarr:
testing things:
- if a user changes the values of multi-select modelsArray via UI, then force update distribution of random autocreated children

PR #1038, opened 2025-01-24
Add support for editing a property of type array

vincentfretin:
Add support for editing a property of type array
and fix react-select usage everywhere, custom no results message is now properly working.


kfarr:
thanks, I have some conflicting work in progress https://github.com/3DStreet/3dstreet/pull/1037 which I might finish first before merging this

PR #1040, opened 2025-01-24
attempt to fix uniqueID issue

kfarr:


src/editor/lib/commands/EntityCloneCommand.js
```diff
@@ -10,17 +10,17 @@ export class EntityCloneCommand extends Command {
...
+    this.entityIdToClone = entity.id; // save the id of the entity to clone
+    this.entityId = null; // this will be the id of the newly cloned entity
   }
 
   execute(nextCommandCallback) {
     const entityToClone = document.getElementById(this.entityIdToClone);
     if (entityToClone) {
-      const clone = cloneEntityImpl(entityToClone, this.entityId);
-      this.entityId = clone.id;
+      const clone = cloneEntityImpl(entityToClone, this.entityId); // why is this.entityId passed? will this always be null?
```
vincentfretin, line 22: In the case of doing entityclone followed by entityupdate, then undo, undo, redo entityclone (we need the same id as the id that is used by entityupdate), redo entityupdate.

src/editor/lib/entity.js
```diff
@@ -155,21 +155,25 @@ export function cloneEntityImpl(entity, newId = undefined) {
...
-    if (entity.id) {
-      if (entity.id.length === 21) {
-        // nanoid generated id, create a new one
-        clone.id = createUniqueId();
-      } else {
-        // Get a valid unique ID for the entity
-        clone.id = getUniqueId(entity.id);
-      }
-    } else {
-      entity.id = createUniqueId();
```
vincentfretin, line 170: huh, you're right, this should have been `clone` instead of `entity` here, bad code on my part :) We currently never go to this branch because we set an id in `EntityCloneCommand` constructor.

src/editor/lib/entity.js
```diff
@@ -155,21 +155,25 @@ export function cloneEntityImpl(entity, newId = undefined) {
...
+  }
+
+  // Add recursive ID generation for children
+  function regenerateIDs(element) {
+    if (element.id) {
+      element.id = createUniqueId();
     }
+    Array.from(element.children).forEach(regenerateIDs);
   }
+  regenerateIDs(clone);
```
vincentfretin, line 175: We need to have children ids to be the same if you do undo entityupdate of child, undo entityclone, redo entityclone, redo entityupdate of child.

vincentfretin:
I fixed the implementation, keeping the detached clone that is never attached, so a pristine copy, to be reused when we redo the action so that other redo actions on a child is working properly because we recreated again with the same ids.

PR #1043, opened 2025-01-27
Backport fixes from inspector 2

vincentfretin:
You don't currently need it, but you may want it in the future: 
- https://github.com/aframevr/aframe-inspector/pull/797
- https://github.com/aframevr/aframe-inspector/pull/798
- https://github.com/aframevr/aframe-inspector/pull/799


vincentfretin:
Yes, you have that in material, geometry and light. I don't think this is even documented. From my understanding this is not even used, it doesn't trigger an update of the schema, like adding shader attributes to the material schema. But that's a useful information to modify the properties shown in the UI.

vincentfretin:
Nice, I'm glad you got an usage for the if condition.

PR #1052, opened 2025-02-02
barely working delete component

kfarr:
needs new icon


vincentfretin (commented):
That's great. There are indeed some styling issues to fix on the add component select. I ended up fixing the original aframe inspector code ahah.

Issue #1060, opened 2025-02-07
longstanding editor camera annoyances


kfarr:
things eating away at me w/r/t interface usability

1. "close up need to zoom a lot to move" -- as a user when i zoom in to an object and wish to keep zooming in more, i want the camera to continue moving forward therefore the rotation point of the viewport moves forward
 	- idea: when below certain level of zoom, the z just starts moving forward instead of asymptotically approaching a limit. i think this is the sketchfab method?
 
2. click on initial load can, in rare cases, mistakingly move an entity, and for some reason undo doesn't work only in this case
 	- perhaps remove the diamond transform gizmo of death as a starting point
 
3. the first click to rotate the camera view usually results in a small "jump" as though the pivot point is suddenly changed -- why and how to fix?
	- should we make rotation pivot point clearer in the ui? compare with blender which is very smooth in this regard

vincentfretin:
point 3 fixed in #1061

kfarr:
Yes the fix for 3 looks good, just merged!

For 1, it looks like if the distance between the requested zoom-in distance (`delta`) is greater than the distance between the camera and the selected object (`distance`) the zooming action is canceled in line 140: 
https://github.com/3DStreet/3dstreet/blob/main/src/editor/lib/EditorControls.js#L135

vincentfretin:
PR #1067 is my take for point 1, what do you think? This includes some of things you said.

kfarr:
This was/is great, thanks Vincent! I'll close this and use a posthog recording the next time I encounter point 2 and then make a new ticket

PR #1067, opened 2025-02-10
Move orbit center and keep minimum speed for zoom and pan in 3D View

vincentfretin:
This fixes point 1 of #1060

Move orbit center if we go past it and keep minimum speed for zoom and pan in 3D View.
This way we can actually use that zoom feature to move in the same direction indefinitely at the same speed.


kfarr:
love it! big improvement @vincentfretin

PR #1070, opened 2025-02-11
Custom focus animation position

kfarr:
seems to work if you do something like this in the console:

`document.querySelector('[mixin="bus-stop"]').setAttribute('focus-camera-pose', 'relativePosition: 1 5 2')`

and then double click to focus on that item, you'll notice the custom focus position

next is to add a function that can be called from the UI to set the `relativePosition` value of focus-camera-pose for a given selected object, care to take a try at that @vincentfretin ?

(sorry ignore unrelated UI changes in index.html and Sidebar.js)


kfarr:
Ok this sort of works, sort of not. In extreme cases the position of the relative position will be such to give a very different viewpoint as compared to when the user long-presses the button.

Not sure if I want to release as is or remove the long-press to set custom focus camera pose

PR #1072, opened 2025-02-12
wip remove length from `street-generated-*` components

kfarr:
why:
- length shouldn't be stored separately in components; it should always come from street-segment
- component should update if street-segment length/width updated

issues to resolve:
- when loading from a scene, the length isn't available until a bit after the component is initialized -- how to handle (answer: handle 0 / undefined length value without error, update to proper value when loaded)
- calcs depending on length and width should update in a uniform manner when street-segment length or width is updated. If street-segment value is changed, then to inform the children of a new length and width value through an event
- the children should listen for the event and update as needed


kfarr:
fixed here instead: https://github.com/3DStreet/3dstreet/pull/1111

PR #1077, opened 2025-02-14
barely working 3dtiles new

kfarr:
requires going to viewer mode

### DIFF
--- a/src/components/google-maps-aerial.js
+++ b/src/components/google-maps-aerial.js
@@ -1,0 +1,113 @@
+import { TilesRenderer } from '3d-tiles-renderer';
+import {
+  TilesFadePlugin,
+  TileCompressionPlugin,
+  GLTFExtensionsPlugin,
+  GoogleCloudAuthPlugin
+} from '3d-tiles-renderer/plugins';
+
+console.log('3d-tiles-renderer', TilesRenderer);
+const MathUtils = AFRAME.THREE.MathUtils;
+const DRACOLoader = AFRAME.THREE.DRACOLoader;
+
+if (typeof AFRAME === 'undefined') {
+  throw new Error(
+    'Component attempted to register before AFRAME was available.'
+  );
+}
+
+AFRAME.registerComponent('google-maps-aerial', {
+  schema: {
+    apiToken: { type: 'string', default: '' },
+    latitude: { type: 'number', default: 35.6586 }, // Tokyo Tower
+    longitude: { type: 'number', default: 139.7454 },
+    minDistance: { type: 'number', default: 500 },
+    maxDistance: { type: 'number', default: 20000 },
+    ellipsoidalHeight: { type: 'number', default: 0 }
+  },
+
+  init: function () {
+    this.initialized = false;
+    console.log('google-maps-aerial init');
+    // Initialize tiles
+    this.tiles = new TilesRenderer(
+      'https://tile.googleapis.com/v1/3dtiles/root.json'
+    );
+
+    // Register plugins
+    this.tiles.registerPlugin(
+      new GoogleCloudAuthPlugin({ apiToken: this.data.apiToken })
+    );
+    this.tiles.registerPlugin(new TileCompressionPlugin());
+    this.tiles.registerPlugin(new TilesFadePlugin());
+    this.tiles.registerPlugin(
+      new GLTFExtensionsPlugin({
+        dracoLoader: new DRACOLoader().setDecoderPath(
+          'https://unpkg.com/three@0.153.0/examples/jsm/libs/draco/gltf/'
+        )
+      })
+    );
+
+    // Set location
+    this.tiles.setLatLonToYUp(
+      this.data.latitude * MathUtils.DEG2RAD,
+      this.data.longitude * MathUtils.DEG2RAD
+    );
+
+    // Create a child entity for the height offset
+    const offsetEl = document.createElement('a-entity');
+    offsetEl.object3D.position.y = -this.data.ellipsoidalHeight;
+    offsetEl.object3D.add(this.tiles.group);
+    this.el.appendChild(offsetEl);
+    this.offsetEl = offsetEl;
+
+    // Wait for camera and scene to be ready
+    this.el.sceneEl.addEventListener('loaded', () => {
+      this.camera = this.el.sceneEl.camera;
+      this.renderer = this.el.sceneEl.renderer;
+
+      if (this.camera && this.renderer) {
+        this.initialized = true;
+        this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+        this.tiles.setCamera(this.camera);
+      }
+    });
+  },
+
+  tick: function () {
+    // only run this function 10 times total
+    if (this.tickCount >= 10) return;
+    if (this.initialized && this.tiles && this.camera) {
+      // Ensure camera is set on each tick
+      this.tiles.setCamera(this.camera);
+      this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+      this.tiles.update();
+    }
+    this.tickCount++;
+  },
+
+  remove: function () {
+    if (this.tiles) {
+      if (this.offsetEl) {
+        this.offsetEl.removeFromParent();
+        this.offsetEl = null;
+      }
+      this.tiles.dispose();
+      this.tiles = null;
+    }
+  },
+
+  update: function (oldData) {
+    // Handle property updates
+    if (
+      this.tiles &&
+      (oldData.latitude !== this.data.latitude ||
+        oldData.longitude !== this.data.longitude)
+    ) {
+      this.tiles.setLatLonToYUp(
+        this.data.latitude * MathUtils.DEG2RAD,
+        this.data.longitude * MathUtils.DEG2RAD
+      );
+    }
+  }
+});
--- a/index.html
+++ b/index.html
@@ -88,6 +88,14 @@
 
     <a-entity id="street-container" data-layer-name="User Layers" data-no-transform>
       <a-entity id="default-street" street streetmix-loader set-loader-from-hash></a-entity>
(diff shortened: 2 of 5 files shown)


src/editor/index.js
```diff
@@ -330,9 +336,12 @@ Inspector.prototype = {
     document.activeElement.blur();
 
     // quick solution to change 3d tiles camera
-    const tilesElem = document.querySelector('a-entity[loader-3dtiles]');
+    const tilesElem = document.querySelector('a-entity[google-maps-aerial]');
     if (tilesElem) {
-      tilesElem.emit('cameraChange', AFRAME.scenes[0].camera);
+      tilesElem.emit(
+        'cameraChange',
+        document.querySelector('#camera').object3D.children[0]
```
vincentfretin, line 343:
```suggestion
        this.cameras.original.getObject3D('camera')
```

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    latitude: { type: 'number', default: 37.795 }, // SF Ferry Building
+    longitude: { type: 'number', default: -122.394 },
+    minDistance: { type: 'number', default: 500 },
+    maxDistance: { type: 'number', default: 20000 },
+    ellipsoidalHeight: { type: 'number', default: 0 },
+    copyrightEl: { type: 'selector' }
+  },
+
+  init: function () {
+    this.tick = AFRAME.utils.throttleTick(this.tick, 10, this);
```
vincentfretin, line 31: It doesn't make sense. 10 is lower than 16ms. Did you mean 100ms?

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    const offsetEl = document.createElement('a-entity');
+    offsetEl.object3D.position.y = -this.data.ellipsoidalHeight;
+    offsetEl.object3D.add(this.tiles.group);
+    this.el.appendChild(offsetEl);
+    this.offsetEl = offsetEl;
+
+    // Get camera and renderer
+    this.camera =
+      this.data.cameraEl?.object3D.children[0] ??
+      document.querySelector('a-scene').camera;
```
vincentfretin, line 68: there is no cameraEl property on your component
so just use
`this.camera = this.el.sceneEl.camera`

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    this.renderer = this.el.sceneEl.renderer;
+    // if (!this.camera) {
+    //   throw new Error('3D Tiles: Please add an active camera or specify the target camera via the cameraEl property');
+    // }
+    this.originalCamera = this.camera;
```
vincentfretin, line 73: no need to store `this.originalCamera` you don't use it

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+  },
+
+  tick: function () {
+    if (this.initialized && this.tiles && this.camera) {
+      // Ensure camera is set on each tick
+      this.tiles.setCamera(this.camera);
+      this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+      this.tiles.update();
+    }
```
vincentfretin, line 111: If you do that in tick, then you can remove all the code related to `cameraChange` event and just use 
`this.el.sceneEl.camera` instead of `this.camera` I think.

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+      new GoogleCloudAuthPlugin({ apiToken: this.data.apiToken })
+    );
+    this.tiles.registerPlugin(new TileCompressionPlugin());
+    this.tiles.registerPlugin(new TilesFadePlugin());
+    this.tiles.registerPlugin(
+      new GLTFExtensionsPlugin({
+        dracoLoader: new DRACOLoader().setDecoderPath(
+          'https://unpkg.com/three@0.153.0/examples/jsm/libs/draco/gltf/'
+        )
+      })
```
vincentfretin, line 49: You should use
`this.el.sceneEl.systems["gltf-model"].getDRACOLoader();`
so you share the draco loader between the gltf-model component and this TilesRenderer.
That will use latest draco decoder 1.5.7 by default on aframe 1.7

src/components/google-maps-aerial.js
vincentfretin: Instead of tick, did you try `Render On Change`?
 https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/254774c7e0d3849a4a964633ed6376b5fbc2414c?tab=readme-ov-file#render-on-change

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    if (this.initialized && this.tiles && this.camera) {
+      // Ensure camera is set on each tick
+      this.tiles.setCamera(this.camera);
+      this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+      this.tiles.update();
+    }
+    if (this.data.copyrightEl) {
+      this.data.copyrightEl.innerHTML =
+        this.tiles.getAttributions()[0]?.value || '';
+    }
```
vincentfretin, line 115: This shouldn't be in tick.

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+        // Set new camera and update resolution
+        this.tiles.setCamera(this.camera);
+        this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+        this.tiles.update();
+      }
+    });
+
+    if (AFRAME.INSPECTOR && AFRAME.INSPECTOR.opened) {
+      // set active inspector camera
+      this.camera = AFRAME.INSPECTOR.camera;
```
vincentfretin, line 99: That line shouldn't be needed. We set this.camera above already. But you need to keep the play() here.

src/components/google-maps-aerial.js
```diff
@@ -0,0 +1,144 @@
...
+    this.renderer = this.el.sceneEl.renderer;
+    // if (!this.camera) {
+    //   throw new Error('3D Tiles: Please add an active camera or specify the target camera via the cameraEl property');
+    // }
+    this.originalCamera = this.camera;
+
+    this.tiles.setResolutionFromRenderer(this.camera, this.renderer);
+    this.tiles.setCamera(this.camera);
+    this.tiles.update();
+    this.initialized = true;
```
vincentfretin, line 78: It seems that `initialized` flag is completely useless. You can remove it and in all conditions.

src/components/street-geo.js
```diff
@@ -168,74 +168,52 @@ AFRAME.registerComponent('street-geo', {
...
-        maximumMem: 400,
-        cameraEl: '#camera',
-        copyrightEl: '#map-copyright',
-        distanceScale: 0.5,
-        emitPostProcess: true
-      });
-      google3dElement.classList.add('autocreated');
+    google3dElement.setAttribute('data-layer-name', 'Google 3D Tiles');
+    google3dElement.setAttribute('data-no-transform', '');
+    console.log(firebaseConfig.apiKey);
```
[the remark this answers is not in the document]
kfarr: 7d9978357830b2ab1154e915754af870ed32383f

src/components/google-maps-aerial.js
kfarr: 7d9978357830b2ab1154e915754af870ed32383f

src/components/google-maps-aerial.js
kfarr: 7d9978357830b2ab1154e915754af870ed32383f

src/components/google-maps-aerial.js
kfarr: c9c1bdf95de6fd6e7f75a5a9f9fe71dbabffe97d

src/components/google-maps-aerial.js
kfarr: 8ef2dd644a2fd481fbb42fb085e08b420606688c

src/components/google-maps-aerial.js
kfarr: you're right the tick was useless bc3c45efbceec4037081297d4ae1cba4dc4fbf7b
i'll take a look at render on change

src/components/google-maps-aerial.js
kfarr: bc3c45efbceec4037081297d4ae1cba4dc4fbf7b

src/components/google-maps-aerial.js
kfarr: This needs to be updated regularly as the attribution values change over time, but you're right that we could probably leverage other events such as `load-model`

src/editor/index.js
kfarr: a3944fe82da101916dc995296e488e4dbdb5d3b5

src/components/google-maps-aerial.js
kfarr: I tried render on change and it wasn't working as expected. i'm going to leave as is for now, the most expensive operation to fix next is applying blending and clipping, the tiles update on each tick is not worse than before

The issue: load-model event only seems to fire when running tiles.update() in each tick, otherwise that event wouldn't fire.

src/components/google-maps-aerial.js
kfarr: I tried only calling tiles.update() when load-model event is fired, but load-model only seems to fire when tiles.update() is called every tick so it doesn't work as expected. It's ok to leave like this for now imho

kfarr:
good enough to merge, followup tickets:
- https://github.com/3DStreet/3dstreet/issues/1122
- https://github.com/3DStreet/3dstreet/issues/1123

vincentfretin:
You can also remove the `tilesElem.emit('cameraChange'` in `src/editor/index.js` and `src/editor/lib/viewport.js`

PR #1078, opened 2025-02-16
Screenshot title customization

kfarr:
- [x] MUST - increase vertical size of modal to 90% VH
- [x] MUST - don't scroll image, only scroll the right-hand panel
- [x] MUST - check if pro account for overlay customization options
- [x] MUST - if not pro account, show upgrade teaser for overlay customization options
- [ ] SHOULD - separate canvas for auto-preview thumbnails upon save vs. user download
- [ ] NICE - save title settings in viewer stub - support for saving / loading
- [ ] NICE - collapsible "component header" section "Font Options" with T style icon
- [ ] NICE - separate color for dropdown options (dark bg) vs. button (purple bg) on download dropdown
- [ ] NICE - ensure that "download as..." still shows after a click, not the previously selected menu option


src/components/managed-street.js
```diff
@@ -54,6 +55,8 @@ AFRAME.registerComponent('managed-street', {
     }
   },
   init: function () {
+    this.zustandState = useStore.getState();
```
vincentfretin, line 58: Don't do this, you get a snapshot of the state and it may be stale when you'll then use that, checking an old state instead of the current one. Use `const zustandState = useStore.getState()` just before you use it.

src/editor/components/modals/ScreenshotModal/ScreenshotModal.component.jsx
```diff
@@ -41,23 +42,16 @@ const getSceneName = (scene) => {
 };
 
 function ScreenshotModal() {
+  // Get the entity that has the screentock component
+  const getScreentockEntity = () => {
+    const screenshotEl = document.getElementById('screenshot');
+    screenshotEl.play();
```
vincentfretin, line 48: Maybe call play only if paused, otherwise you may register listeners again and have issues.
```js
if (!screenshotEl.isPlaying) {
  screenshotEl.play();
}
```

src/components/managed-street.js
kfarr: 086c2eaa42e65246af9ea69f206234cf14cd9c91

src/editor/components/modals/ScreenshotModal/ScreenshotModal.component.jsx
kfarr: 509990df0bb8ba5ca6fea2c7794d411669ce7094

vincentfretin (commented):
The properties works great. The horizontal scroll was showing on my machine just to scroll about 10px. I dug a bit but didn't understand why, so I added overflow-x hidden to hide it.
I did some changes to hide the two stroke options if stroke is not enabled, hope this is okay.

PR #1080, opened 2025-02-19
Ruler tools

kfarr:
Must fix issues:
- [x] display the ruler value in the viewport when creating and when selected - via canvas? via three.js text?
- [x] redo (after undo) of measure-line segment doesn't work, not sure why
- [x] while in ruler mode, I cannot zoom or pan
- [x] ruler pointer - should be pointer finger instead
- [x] hide helper object from scene graph
- [x] proper removal of measurement css2d after removal

Other issues:
- [x] - the css2d labels are in front of other elements on the screen such as panels
- [x] metric / imperial switch

nice to have / to fix later:
(moved to https://github.com/3DStreet/3dstreet/issues/1081)

wait until people ask for it:
- exporting an image doesn't show css2d labels
- should this be hidden for gltf export?


vincentfretin:
For label, you can use CSS2DObject/CSS2DRenderer
https://threejs.org/examples/?q=label#css2d_label

vincentfretin:
nice labels
You maybe want to remove those measure-lines entities from the glb export?

PR #1093, opened 2025-02-26
barely working bloom

kfarr:


kfarr:
the most recent commit was an attempt to get ssao working based on this three.js example:
https://threejs.org/examples/webgl_postprocessing_ssao.html

however even cranking the settings up to the max, and switching to standard depth buffer, only produces a very faint ssao effect which isn't very visible. [2]

Maybe it's time to look up some other ssao effects like
https://github.com/N8python/n8ao
or
https://threejs.org/examples/?q=ambient#webgl_postprocessing_sao
or
https://threejs.org/examples/?q=ambient#webgl_postprocessing_gtao

[2]


image

PR #1113, opened 2025-04-05
Check explicitly for canvas element instead of doing typeof string

vincentfretin:
Context: https://github.com/aframevr/aframe-inspector/pull/806
You already had a fix with https://github.com/3DStreet/3dstreet-editor/commit/e5ee022e7a5ab2625f3c68d9be6a97023ac4e159
if your issue was also with a canvas texture, not sure if you remember...
Here the fix is different and is also removing "URL: undefined" from the tooltip.

Do you have a case where you have a canvas texture for material src?


kfarr:
it looks like it was related to a fix when clicking on entity with one of the mapbox options:
https://github.com/3DStreet/3dstreet-editor/issues/118

so i can just test this pr on all of the mapbox options to make sure it works ok? then I think it's good? we also use canvas for the street labels.

kfarr:
works great, thanks @vincentfretin for keeping things updated with the inspector.

PR #1116, opened 2025-04-09
Use error boundary for better react error handling

kfarr:
Power users sometimes manage to get in a state where all of the Editor UI will disappear which is likely a fatal react error that unmounts components. It is difficult to diagnose and users are not familiar with how to open and capture contents of javascript console. Therefore this is an attempt to use error boundaries to display a UI when there is a fatal error so that the user can choose to report the issue with enough info for us to debug in the future.

This PR is mostly from Claude. I'm not sure if it would actually catch anything, I tried to find a "real error" but all I was able to do was  having Claude also make me a react component that triggers the error to confirm the ui is ok. :/

### DIFF
--- a/src/editor/components/ErrorBoundary.js
+++ b/src/editor/components/ErrorBoundary.js
@@ -1,0 +1,193 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * ErrorBoundary component for 3DStreet that catches React errors while allowing
+ * the underlying A-Frame/Three.js canvas to continue running.
+ */
+class ErrorBoundary extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      hasError: false,
+      error: null,
+      errorInfo: null,
+      isDetailsExpanded: false
+    };
+    this.copyErrorDetailsRef = React.createRef();
+  }
+
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI
+    return { hasError: true, error };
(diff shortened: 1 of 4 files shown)


src/editor/components/ErrorBoundary.js
```diff
@@ -0,0 +1,193 @@
...
+          setTimeout(() => {
+            copyButton.textContent = originalText;
+          }, 2000);
+        }
+      } catch (e) {
+        console.error('Fallback copy failed:', e);
+      }
+
+      document.body.removeChild(textarea);
+    }
```
vincentfretin, line 100: The fallback is used for non-https and IE11.
navigator.clipboard.writeText should be implemented in all evergreen browsers.
https://caniuse.com/mdn-api_clipboard_writetext
Note that in the code base we use clipboard dependency to do the clipboard.writeText with fallback used in  
Component.js CommonComponents.js StreetSegmentComponent.js and I replaced that dependency by clipboard-copy that is more lightweight in aframe-inspector. https://www.npmjs.com/package/clipboard-copy
We also just use clipboard.writeText without fallback in AIChatPanel.js ScreenshotModal.component.jsx
For consistency, I would remove the fallback and remove the clipboard dependency, or use the clipboard-copy dependency everywhere.

vincentfretin (commented):
That indeed works and is implemented the way it's supposed to according to the documentation
https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

Issue #1120, opened 2025-04-24
POC for report generator


kfarr:
As a user I'd like to create a report for my team proposing a street change using 3DStreet.

I'd like to identify the location, problem statements, etc.

Using this information I expect 3DStreet to create an outline and draft of a street improvement plan

I'd like each section to have an illustration.

I'd like the ability to copy/paste the text as rich text that can be pasted into a google doc

PR #1121, opened 2025-04-24
report generator experiment

kfarr:
https://github.com/3DStreet/3dstreet/issues/1120

Things left to achieve this:
- [x] - add ability to run a function that will take a snapshot
- [x] - update styling of report modal - make it wider
- [x] - add location chooser for specific area in modal
- [x] - function: takeScreenshot; focus: target;
- [x] fix confusion of llm not generating report along with screenshot
- [x] let's reduce the complexity of the aichatpanel --> first break out into separate prompt file(s)? also break out functions into their own file(s).
- [x] - clean up unnecessary code that ai generated
- [x] - copy / download icon buttons for inline chat img snapshots
- [x] update AI ui for first message, have new templatized buttons / update styling to match styling from popular gpt-style interfaces
- [x] - move default messages location to top of window but beneath menu bar (as to not collide with ai panel)
- [x] - copy/paste of rich text should show the rich text in the editor to which I'm pasting, not the raw markdown text
- [x] btw why is the markdown text only with ** instead of using # for headings
- [x] have an action button with a prefilled prompt to take 3 pictures of the scene: Prompt: Take 3 different snapshots, one is birdseye view (the street position located 50 meters up looking straight down); one is Straight on (+z 20 m from the street looking forward), variety – get a photo oriented around a specific object
- [x] create an action button for a template to create a new managed street]
- [x] fix imperial / metric text size --> where is that defined - actionbar?
- [x] payment modal -- should block header with x close button
- [x] fix imperial / metric text size
- [x] Fix styling in report modal - make text area formatting match input text formatting. IE in input area there is gray background and purple rounded border 1px
- [x] report builder validation - the following fields are required: project description; problem statement; geo location set (but not project location description, that is still optional); there should be a validation error shown if the fields aren't filled in; required fields should somehow indicate they're required
- [x] ensure that opening geomodal returns to report modal when closing - right now when clicking "set location" from the report modal, it correctly opens the geomodal. However, after clicking "Update Scene Location" in geomodal it does NOT return to the report modal as expected. There may be outdate code to attempt to make this work, but it doesn't work. 
- [x] - DONE - validation reset when click cancel
- [x]DONE question -from report modal, when typing project description, clicking to set location, and returning, is the existing content saved? it should be but not trigger validation and don't trigger actual report generation
- [x] NEXT reset aichatpanel upon new load
- [x] Then this is "good enough" to ship and start marketing around
- [x] allow LLM to adjust geo location
- [x] Create good enough basic AI docs
- [x] LLM does not have up-to-date list of models; instead of being in prompt this model list should be loaded in the same way as elsewhere in the application (probably the drop down menu options for the sidebar)

Nice to have:
- [ ] enhance geolocation with metadata from google / geonames
- [ ] not all objects have uuid when passing to LLM --- instead, all objects should have uuid
- [ ] managed street autocreated entities are not passed for llm scene graph state

LATER
- [ ] create showButton function that will accept a parameter for the button name that will be displayed. 2 examples: open report modal and set / edit location; 
- [ ] create better ui for memory / display-only values with button for editing report
- [ ] making a snapshot actually makes a saved snapshot that can be viewed as an entity; the focused entity can also be selected when the snapshot is selected via a custom snapshotsidepanel option
- [ ] - function: create a polygon that is a highlight of a street corridor --> fetches a way from mapbox for example and creates a line that represents it; or creates a geometry that acts as a highlight ...
- [ ] LATER should use a function to generate report instead? create as a new entry in child of report class parent under memory entity


src/editor/components/scenegraph/AIChatPanel.js
```diff
@@ -1218,6 +1238,52 @@ const AIChatPanel = () => {
...
+                if (!projectInfoEntity) {
+                  console.log('Creating project info entity');
+                  projectInfoEntity = document.createElement('a-entity');
+                  projectInfoEntity.setAttribute('id', 'project-info-entity');
+                  projectInfoEntity.setAttribute('project-info', {});
+
+                  // Add to the scene under user layers
+                  const userLayers =
+                    document.querySelector('#user-layers') ||
+                    document.querySelector('a-scene');
```
vincentfretin, line 1260: Why do you have a fallback to a-scene here? #user-layers should always exists right?

src/editor/lib/cameras.js
```diff
@@ -36,7 +36,7 @@ export function initCameras(inspector) {
 
   // Create Inspector camera.
   const perspectiveCamera = (inspector.camera = new THREE.PerspectiveCamera());
-  perspectiveCamera.far = 20000; // Changed from 10000 to 20000
+  perspectiveCamera.far = 100000000; // Changed from 10000 to 20000
```
vincentfretin, line 39: Update the comment as well, just keep `// Changed from 10000` so you won't need to update the comment next time.

src/editor/components/scenegraph/AIChatPanel.js
kfarr: yeah it's actually a mistake should be '#street-container' only, thanks for catching

src/editor/lib/cameras.js
kfarr: good idea thanks

src/editor/lib/EditorControls.js
```diff
@@ -159,6 +159,8 @@ THREE.EditorControls = function (_object, domElement) {
 
   this.zoom = function (delta) {
     var distance = object.position.distanceTo(center);
+    object.far = Math.min(100000000, Math.max(20000, distance * 10));
+    object.updateProjectionMatrix();
```
kfarr, line 163: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1110

kfarr:
@vincentfretin would love feedback in a timeboxed review if you have time; there is probably too much here so I would focus your attention on the a-frame adjacent things such as ensuring robustness of the memory entity concept (or perhaps the memory entity "hack" since it's kinda weird) and the new snapshot tool and how it may interface with camera control, etc.

src/street-utils.js
```diff
@@ -52,6 +52,21 @@ export function newScene(
...
+  }
+
+  // reset AI Chat
+  if (
+    window.aiChatPanelRef &&
+    typeof window.aiChatPanelRef.resetConversation === 'function'
+  ) {
+    window.aiChatPanelRef.resetConversation();
+  }
+
```
vincentfretin, line 69: Wouldn't it better to use the newScene event to reset the conversion from the AIChatPanel component? Similar to what we do in src/editor/contexts/Geo.context.js and src/editor/index.js (search for newScene). Better to avoid any usage of global variables like this.

src/editor/components/modals/ReportModal/ReportModal.component.jsx
```diff
@@ -0,0 +1,442 @@
...
+    description: '',
+    problemStatement: '',
+    geoLocation: ''
+  });
+
+  const [touched, setTouched] = useState({
+    description: false,
+    problemStatement: false,
+    geoLocation: false
+  });
```
vincentfretin, line 33: You're starting to write your own form library at this point.
Did you look at using https://react-hook-form.com for example that is quite popular? react-hook-form is the library used by shadcn https://ui.shadcn.com/docs/components/form with the zod validation library https://zod.dev

kfarr:
@vincentfretin Thanks for all the feedback. There are more than a few bad decisions Claude and I have made. I think the worst of all seems to be saving project-info memory as entity/component -- that's also hard to reverse if users save a bunch of stuff in that format, so perhaps it is worth spending a little bit of time to do it the "right" way using zustand and saving/load as specific json key.

src/editor/components/modals/ReportModal/ReportModal.component.jsx
kfarr: Thanks for this feedback. I can fix this later, it's good enough to keep as-is for now and then can replace in the future without affecting core flow or data

src/street-utils.js
kfarr: Thanks good idea https://github.com/3DStreet/3dstreet/pull/1121/commits/ef7a077aca497a901b0ee0453aeaf3da7cbd332e

src/editor/components/Main.js
```diff
@@ -26,6 +27,16 @@ THREE.ImageUtils.crossOrigin = '';
...
+  // Create a ref for the AIChatPanel component
+  const aiChatPanelRef = useRef(null);
+
+  // Expose the ref globally for other components to access
+  useEffect(() => {
+    window.aiChatPanelRef = aiChatPanelRef.current;
+    return () => {
+      window.aiChatPanelRef = null;
+    };
+  }, []);
```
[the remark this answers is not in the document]
kfarr: Good feedback to consider using zustand for this. Unlike memory which had an irreversible downstream effect of saving user data in a bad way, this can be fixed later without affecting data storage, so I'll leave it for now since it works.

src/editor/components/scenegraph/AIChatPrompt.js
```diff
@@ -0,0 +1,250 @@
...
+      
+      Always look for opportunities to update the project information when users provide relevant details, even if they don't explicitly ask you to update it.
+
+      IMPORTANT: Always respond with a text message, even if the user is asking for a function call.
+
+      You can invoke more than 1 function at a time.
+
+      Most of the time you're polite but serious. Sometimes you can be a little bit more fun and creative. You're a friendly AI assistant that can help you with your 3D scenes. You can also help with other tasks, such as thinking about safe streets, morality of spatial equity, dad jokes, or general moral humanistic quandaries a la Claude from Anthropic. But mostly you're there to help modify 3DStreet scenes.
+
+      Sometimes the user may ask that you update the scene location. You can do this by updating the geospatial latidude and longitude. Try your best to estimate the lat/lon based on the user's description and tell them you've found the general area, and instruct them to click "Edit Precise Location" button to find the precise desired location.
```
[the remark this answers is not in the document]
kfarr: thanks dude!

PR #1128, opened 2025-04-30
Allow drag and drop entity DOM ordering

kfarr:
Sort of works

Still to do:
- Replace row-based dragging with a dedicated grab handle in the left margin area (in same area as leftmost checkmark, so this will only show in entities below user layers); the grab handle should only show on hover  
- Improve visual feedback with a line indicator showing precise drop location (above/below based on drag direction) 

Known Issues:
- moving a managed street doesn't seem to work reliably -- sometimes it would disappear
- sometimes moving a crosswalk box would make it disappear
- moving a segment within a managed street doesn't work -- the surface is not redrawn in its new location. reloading works

Future:
- reordering should eventually be implemented as an undoable command
- Enable cross-branch dragging (moving entities between different parent containers)

### DIFF
--- a/src/editor/components/scenegraph/Entity.js
+++ b/src/editor/components/scenegraph/Entity.js
@@ -51,7 +51,19 @@ export default class Entity extends React.Component {
       entity.id === 'reference-layers' ||
       entity.id === 'environment';
 
-    if (tagName === 'a-scene' || isContainer) {
+    // Check if the entity is within the user layers (street-container)
+    let isInUserLayers = false;
+    let parent = entity.parentElement;
+    while (parent && parent.isEntity) {
+      if (parent.id === 'street-container') {
+        isInUserLayers = true;
+        break;
+      }
+      parent = parent.parentElement;
+    }
+
+    // Prevent dragging if it's a scene, container, or not in user layers
+    if (tagName === 'a-scene' || isContainer || !isInUserLayers) {
       e.preventDefault();
       return false;
     }


src/editor/components/scenegraph/SceneGraph.js
```diff
@@ -302,6 +305,79 @@ export default class SceneGraph extends React.Component {
...
+      newIndex: targetIndex
+    });
+
+    // Rebuild the entity list to reflect the new order
+    this.rebuildEntityOptions();
+
+    // Trigger autosave
+    // NOTE: In a future PR, this should be implemented as an undoable command
+    // so that undo/redo works properly and autosave is triggered through the history system
+    const useStore = require('@/store').default;
```
vincentfretin, line 373: Why not just an import statement at the beginning of the file?
Avoid mixing import and require in the same file if possible. Although webpack and bun support that, that's not valid for an ES module.

vincentfretin:
Don't we have issues with `insertBefore` and entity reparenting in general? https://github.com/aframevr/aframe/issues/2425
You should probably do `entity.flushToDom();` before the `insertBefore`.

I'm curious if you tried the newer api `moveBefore`
https://developer.mozilla.org/en-US/docs/Web/API/Element/moveBefore

vincentfretin:
For moveBefore, that's not implemented in Firefox and Safari, so we may not want to use it. Also it has the same effect as insertBefore currently, just tested it on my project, the model indeed disappear. We probably need to implement the connectedMoveCallback on AEntity to move the threejs object3d to the new parent as well, and maybe other changes. I won't pursue that lead. I'll go with serializing the entity, remove and recreate.

PR #1135, opened 2025-05-28
Rename components directories

kfarr:
there are 3 directories named "components" which is confusing


kfarr:
This seems to work and it's super basic to rename directories but it is scaring me for some reason ... @vincentfretin does this make sense to you and seem ok?

vincentfretin (commented):
Having /components with the aframe components and /editor/components with react components really don't bother me. /editor/components/components was indeed a bit redundant, but not sure if renaming that folder to elements is the solution here, I would have probably just moved all those single components to the parent directory. That's just my opinion, go ahead if you find this better.

kfarr:
@vincentfretin thanks fair points. I still kinda like separating out the little "bits" in elements and having the bigger chunks one level higher and I have sunk cost fallacy. But you're right it would make just as much sense to follow your suggestion.

PR #1138, opened 2025-06-01
Ar webxr viewing mode

kfarr:


src/aframe-components/viewer-mode.js
```diff
@@ -75,15 +89,83 @@ AFRAME.registerComponent('viewer-mode', {
     this.cameraRig.setAttribute('cursor-teleport', 'enabled: false');
     this.camera.setAttribute('look-controls', 'enabled: false');
 
+    // Save and remove blink-controls from hands
+    if (this.leftHand && this.leftHand.hasAttribute('blink-controls')) {
+      // Store current settings before removing
+      this.blinkControlsSettings.leftHand =
+        this.leftHand.getAttribute('blink-controls');
```
vincentfretin, line 96: Important you need to make a copy, component.data will be recycled for another component when the component is removed.
```js
      this.blinkControlsSettings.leftHand =
        {...this.leftHand.getAttribute('blink-controls')};
```

kfarr: thanks for helping avoid another landmine
871efbdb76bb50d8e9548083c9fde0af61376b58

PR #1151, opened 2025-06-10
Measure line for camera path

kfarr:


src/aframe-components/viewer-mode.js
```diff
@@ -19,9 +19,14 @@ AFRAME.registerComponent('viewer-mode', {
...
       type: 'string',
       default: 'circle',
-      oneOf: ['circle', 'forward', 'strafe'],
+      oneOf: ['circle', 'forward', 'strafe', 'measure-line'],
       if: { preset: 'camera-path' }
     },
+    measureLineEntity: {
+      type: 'string',
+      default: '',
+      if: { preset: 'camera-path', cameraPath: 'measure-line' }
```
vincentfretin, line 28: Nice, I think that the first time we use two properties for a condition here. When I wrote the condition logic, I thought about it and implied a AND if there were more than one property.

PR #1156, opened 2025-06-12
non-owner cannot create AR scene

kfarr:


kfarr:
Using STREET global is not good for this but it works for now

https://github.com/3DStreet/3dstreet/issues/1157

PR #1161, opened 2025-06-12
update 3DTilesRendererJS to v0.4.10

kfarr:


kfarr:
see https://github.com/3DStreet/3dstreet/pull/1164

PR #1164, opened 2025-06-14
Tiles flattening experiment

kfarr:


src/editor/components/elements/GeoSidebar.js
```diff
@@ -5,11 +5,168 @@ import AdvancedComponents from './AdvancedComponents';
...
+
+    // Generate a unique ID
+    const shapeId = 'flattening-shape-' + Date.now();
+
+    // Create the default flattening shape using inspector command
+    const definition = {
+      id: shapeId,
+      element: 'a-box',
+      components: {
+        'data-layer-name': 'Geo Flattening Shape',
```
vincentfretin, line 63: That data-layer-name shouldn't be in components but at the same level as components. Does your code work? I see you're doing above
```js
const existingShape = document.querySelector(
      '[data-layer-name="Geo Flattening Shape"]'
    );
```

vincentfretin: uh and also class #1170

Issue #1166, opened 2025-06-17
allow drag and drop to reorder / reparent


kfarr:

vincentfretin:
The work started on #1128 I'll continue it, I just implemented it in my solidjs/tailwind project with a new entityreparent command that supports reordering in the same parent or reparent.

vincentfretin:
#1169 implements it, but I disabled for now moving or dropping on managed-street and street-segment. We probably need a few more hours to investigate some weird issues with those components.

PR #1169, opened 2025-06-20
Add new entityreparent command and ui to reorder / reparent an entity

vincentfretin:
This is implementing #1166 and replaces #1128
But we still have the some weird issues:
- moving a street segment will make the adjacent segment disappear
- probably others... I didn't do much test

moving something not related to managed-street or street-segment works fine, so the remaining issue are related to how those components behave.
I disabled for now moving or dropping on managed-street and street-segment.
So in the current state this is quite limited, you can reparent a bicycle to another bicycle, or reordering inside a custom-group. There is no way to create your own group I think currently, we disabled creating an empty entity.


vincentfretin:
Which issue did you find with other types of objects? specific to reparenting or do you have an issue with just reordering?

Not sure if just reordering gives enough value here. The main use case I see is really reparenting to create a group so you can move the whole group.

This folder entity would be just an entity without mesh really, and you can disallow reparenting an entity with mesh on another entity with mesh.

PR #1230, opened 2025-07-14
Select street-segment first then a clone

vincentfretin:
xplosneer Request: When selecting a street, I'm only very rarely trying to modify a model in the street itself. I'm usually trying to adjust length, rotation, or position. Make it so when I click over a segment, the default is to select the entire segment. If I select again while the street is already selected, then select the lane or clone.

https://github.com/user-attachments/assets/e9e5d4a0-f5a0-4c37-8b1f-f1b4708e2b43


kfarr:
works great! thanks @vincentfretin

PR #1298, opened 2025-08-16
[IGNORE] Entity color options revisit

ewei2406:
Some downstream repo syncing problems, ignore


kfarr:
As you noted in the assets-dist repo, there is a file size impact of vertex colored models; the file size itself is probably ok as the objects can easily be cached. I also tested this out with complicated scenes and it seems to work ok, for example https://dev-3dstreet.web.app/#/scenes/848d4f7f-d04b-491c-b531-216ed8b9b593

checklist before merging

- [ ] - when loading a saved scene with a managed street where a user chose "neutral" colors, the materials of the segment ground (roadway) appear white
- [ ] - when loading a saved scene with managed street where user chose "neutral" colors, the colors of the loaded vehicles do not match the colors when the scene was saved (ie colors for random or neutral should be deterministic. could use a seed like we do for cloned items)
- [ ] - the choice of random, neutral, "reset" -- is this currently stored somewhere? how does the user know which option they selected? this could use more code review from kieran
- [ ] - still to test / review, editing ui of an individual vehicle or object

PR #1364, opened 2025-10-05
update_streetplan_mapping

atmobi998:
update streetplan mapping


kfarr:
@atmobi998 can you be a bit more descriptive in the pr name and/or description about this pr beyond mapping since seeing a bunch of PRs with the same name might be confusing for those looking at project history?

for example is it... update building mappings? another type of segment?

PR #1370, opened 2025-10-12
streetplan mapping update

atmobi998:
streetplan mapping update


atmobi998:
> mapping

change building mode from random to fit, fit mode working fine

PR #1458, opened 2026-02-26
streetplan_change_segment_name

atmobi998:
streetplan change segment's name


atmobi998:
changes as Claude suggested !
Thanks

PR #1460, opened 2026-02-26
⏳ Use InstancedMesh for generated clones and pedestrians

kfarr:
🧑‍🦲 Status:
- pending user feedback, temporarily deployed to dev server
- issue: some 3d objects when instanced appear as oversized and distorted versions, possibly due to complicated transforms in the affected models
- since instanced objects aren't selectable (which is ok, kinda by design), there is no detach button visible; instead the managed street segment should have a detach option (or the street itself)

🤖  Summary

- Replace individual A-Frame entity creation with `THREE.InstancedMesh` in `street-generated-clones` and `street-generated-pedestrians` components
- Each unique model is now rendered as a single instanced draw call instead of one entity per clone
- `detach()` still works — stored placement specs are used to recreate individual entities on demand
- New shared utility `src/lib/instanced-mesh-helper.js` handles model loading and instanced mesh creation

## Performance Results (East Ave scene)

| Metric | Production (before) | With Instancing (after) | Improvement |
|---|---|---|---|
| **Draw calls** | 2,397 | 1,954 | -18% |
| **Entities** | 5,525 | 2,112 | **-62%** |
| **Autocreated entities** | 5,083 | 1,670 | **-67%** |
| **Three.js Objects** | 39,274 | 10,036 | **-74%** |
| **JS Heap** | 2,466 MB | 1,708 MB | **-31%** |
| **Triangles** | 5.68M | 5.57M | ~same |
| **FPS (avg)** | 1.3 | 36.8 | **28x faster** |
| **Frames in 10s** | 12 | 358 | **30x more** |

The remaining 1,670 autocreated entities are all from stencil components (left unchanged per plan).

## Test plan

- [x] `npm test` — all 309 tests pass
- [x] `npm run lint` — no new errors
- [x] Performance profiling — 28x FPS improvement on freezing scene
- [ ] Visual check — models appear in correct positions/rotations - FAIL for some models such as streetplan foliage and buildings
- [ ] Editor `detach()` — select segment, detach generated-clones, verify entities created -- FAIL do not have ability to trigger since instanced objects aren't selectable which is how the current detach button is accessed by user
- [ ] Test with various street templates (Streetmix import, StreetPlan import)


vincentfretin:
Alternative with BatchedMesh:
- #1755

kfarr:
closed by #1755

PR #1531, opened 2026-04-06
changes to streetplan street's label, add rulers to labels

atmobi998:
changes to streetplan street's label, add rulers to labels

Screenshot from 2026-04-02 10-46-27

### DIFF
--- a/src/aframe-components/managed-street.js
+++ b/src/aframe-components/managed-street.js
@@ -315,6 +315,20 @@ AFRAME.registerComponent('managed-street', {
         variant: segment.variant,
         side: segment.side
       });
+
+      const streetURL = window.location.hash.substring(1);
+      if (streetURL.includes('streetplan.net/')) {
+        segmentEl.setAttribute('data-layer-rulergrpa', segment.rulergrpa);
+        segmentEl.setAttribute('data-layer-rulergrpaw', segment.rulergrpaw);
+        segmentEl.setAttribute('data-layer-rulergrpaofs', segment.rulergrpaofs);
+        segmentEl.setAttribute('data-layer-rulergrpb', segment.rulergrpb);
+        segmentEl.setAttribute('data-layer-rulergrpbw', segment.rulergrpbw);
+        segmentEl.setAttribute('data-layer-rulergrpbofs', segment.rulergrpbofs);
+        segmentEl.setAttribute('data-layer-leftsidec', segment.leftsidec);
+        segmentEl.setAttribute('data-layer-rightsidec', segment.rightsidec);
+        segmentEl.setAttribute('data-layer-price', segment.price);
+      }
+
       segmentEl.setAttribute('data-layer-name', segment.name);
       // wait for street-segment to be loaded, then generate components from segment object
       segmentEl.addEventListener('loaded', () => {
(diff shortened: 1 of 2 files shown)


kfarr:
Thanks for working on this @atmobi998 — the idea of showing grouped label bands (like "Roadside > Sidewalk > Buffer") across segments is interesting for understanding a street cross-section at a glance.

That said, I can't merge this PR as-is. It has **breaking changes that affect all 3DStreet users** (not just StreetPlan users), and there are deeper design questions we need to think through before committing to this feature.

---

## Breaking changes

**The \`totalWidth\` regression breaks all non-StreetPlan labels.** The calculation was changed from \`widthsArray\` to \`rulerawidArray\`, which is empty for non-StreetPlan scenes. This means \`totalWidth = 0\` for every Streetmix or manually-created street, **breaking label rendering for all existing users**. We have to be very careful with changes to shared components like \`street-label\` — they affect everyone.

**The \`labelHeight\` default is evaluated once at module load time** based on the URL hash. If the hash isn't set yet when the module loads (which is common), StreetPlan scenes get the wrong height. A-Frame schema defaults are baked in at registration time and can't react to later navigation.

**\`setUnitsPreference\` is called unconditionally on StreetPlan load**, silently overriding the user's existing unit preference. (Already discussed on PR #1530.)

## Other issues from code review

- **Unguarded \`rulerdata\` lookup** — \`rulerdata[...].length\` will crash with a TypeError if the key isn't found
- **\`rulerdata\` duplicated twice** (~200 lines copy-pasted) with the two copies already disagreeing: \`r1901\` is \`'Buildings'\` in one and \`'Land Use'\` in the other
- **\`window.location.hash\` checks in 4 places** across 2 files — breaks when scenes are saved and reloaded (URL changes to \`#scenes/abc123\`)
- **Arrays used as hash maps** — \`rulerawid = []\` used with string keys should be \`{}\`
- **Inconsistent unit conversion constants** — \`0.3048\`, \`3.2808\`, and \`3.28084\` all appear for the same conversion

## Bigger design concerns

Beyond the code issues, I'm concerned about the complexity of adding multiple layers of label grouping to 3DStreet. This is a deep feature that we'd have to maintain long-term, and I'm not convinced we've thought through the implications:

**These are display labels, not semantic data.** Looking at the StreetPlan JSON, \`Group1\`/\`Group2\` are categorizations for StreetPlan's 2D ruler display — they're deterministic from the segment Type and side (e.g. "Walkways" on the left is always "Sidewalk" / "L. Roadside"). They don't add new semantic information beyond what the segment type already tells us. So the question is: should 3DStreet store and render StreetPlan's display categorization, or is that better left in StreetPlan's 2D view?

**No editing UI or update mechanism.** If a user duplicates, copies, deletes, rearranges, or changes segment types in the 3DStreet editor, there's no way to update or recalculate the group assignments. The groups would immediately become stale or incorrect. How will 3DStreet users be able to manage these fields? What's the expected behavior?

**Read-only strings with no user control.** Can users see these values in the properties panel? Edit them? If not, they're a black box that silently appears on some scenes and not others.

**Maintenance burden.** Adding a second and third rendering layer to \`street-label\` means we need to maintain, test, and keep them working through future changes to segments, the scene graph, save/load, undo/redo, copy/paste, etc. That's a real ongoing cost.

## Path forward

I'm not saying no to this feature, but I want to discuss it more before we commit. Some options ranging from simple to complex:

1. **Keep grouped ruler labels in StreetPlan's 2D view only.** The 3D view focuses on spatial visualization; the detailed label hierarchy may not need to be replicated here.
2. **Show Group1 as an optional subtitle on the existing single-row labels.** Much less complexity — just one extra line of text per segment, no spanning/grouping logic, and it degrades gracefully when segments are edited.
3. **Full grouped bands (what this PR attempts).** Most feature-rich, but also the most complex to build, maintain, and keep in sync when users edit streets.

Can you share your thoughts on the above?

PR #1532, opened 2026-04-07
dashed strip between buslane and drive lane

atmobi998:
dashed strip between buslane and drive lane

### DIFF
--- a/src/aframe-components/managed-street.js
+++ b/src/aframe-components/managed-street.js
@@ -546,6 +546,18 @@ AFRAME.registerComponent('managed-street', {
         variantString = 'dashed-stripe';
       }
 
+      const streetURL = window.location.hash.substring(1);
+      if (streetURL.includes('streetplan.net/')) {
+        if (
+          (currentSegment.type === 'bus-lane' &&
+            previousSegment.type === 'drive-lane') ||
+          (currentSegment.type === 'drive-lane' &&
+            previousSegment.type === 'bus-lane')
+        ) {
+          variantString = 'dashed-stripe';
+        }
+      }
+
       // Drive lane and turn lane combination would go here if needed
     }
 


kfarr:
@atmobi998 Why is this needed? What is wrong with the default of solid stripe for travel lane adjacent to bus only lane?

If this is still needed, this is the wrong file for it, as Claude is saying. Instead it should be part of your streetplan mapping file. I would close this without merging unless you can address why this needs to be in managed-street which is global and affects all users?

atmobi998:
changes isolated to streetplan streets only:

      const streetURL = window.location.hash.substring(1);
      if (streetURL.includes('streetplan.net/')) {
        if (
          (currentSegment.type === 'bus-lane' &&
            previousSegment.type === 'drive-lane') ||
          (currentSegment.type === 'drive-lane' &&
            previousSegment.type === 'bus-lane')
        ) {
          variantString = 'dashed-stripe';
        }
      }

with the current logic, this should be applied to managed-street.js file ...

kfarr:
Thanks for splitting these out @atmobi998. It's up to you if the bus-lane adjacent to drive-lane should have dashed striping in StreetPlan imports by default, although in my experience those are usually solid stripes and I'll keep that setting as the 3DStreet default.

The issue is *where* to make it. `managed-street.js` is the core engine shared by all sources (Streetmix, StreetPlan, manual creation), so we want to avoid source-specific logic there — especially checking `window.location.hash`, which is fragile and won't work for saved/reloaded scenes.

The good news is there's already a clean way to do this. In `managed-street.js` line 321, the auto-striping logic is skipped when `segment.generated.striping` is already set. So during the StreetPlan import (in `street-mapping-streetplan.js` or the StreetPlan parsing section of `managed-street.js`), you can set the striping explicitly on the segment data *before* the segment element is created:

```js
// During StreetPlan segment processing, after determining segment types:
if (
  (segment.type === 'bus-lane' && previousSegment?.type === 'drive-lane') ||
  (segment.type === 'drive-lane' && previousSegment?.type === 'bus-lane')
) {
  segment.generated = segment.generated || {};
  segment.generated.striping = [{ striping: 'dashed-stripe', length: streetLength, segmentWidth: segment.width }];
}
```

This way the StreetPlan import controls its own striping without touching the shared logic, and it'll work correctly when scenes are saved and reloaded too.

PR #1539, opened 2026-04-15
Fix selection box not updating when moving group entities via panel or undo

vincentfretin:
The entityupdate handler required object3DMap.mesh to update the selection box, so group entities (parents without a mesh) were skipped. Removed that check and the redundant second handler's geometry-based selection box update.

Steps to reproduce
- move managed street entity
- undo
- the selection box should be updated

### DIFF
--- a/src/editor/lib/viewport.js
+++ b/src/editor/lib/viewport.js
@@ -380,10 +380,7 @@ export function Viewport(inspector) {
 
   Events.on('entityupdate', (detail) => {
     const object = detail.entity.object3D;
-    if (
-      inspector.selected === object &&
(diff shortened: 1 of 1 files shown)


kfarr (approved):
the fix works great, thanks @vincentfretin !

vincentfretin:
@claude about Edge case — empty groups. Box3.setFromObject on a group with no children (or one whose children have not loaded yet) produces a degenerate bounding box. This was already true before this PR and Three.js handles it gracefully, but it is worth keeping in mind.

Yes there is the `if (_box.isEmpty()) return;` check in the `OrientedBoxHelper.update` so we handle the case, that's also present in `BoxHelper.update`.

PR #1572, opened 2026-04-23
Support editing a property of type array without oneOf like animation startEvents

vincentfretin:
For array-type properties without oneOf (rendered via InputWidget), stringify the array for display using schema.stringify (", "-joined) and parse the comma-separated string back to an array in onChange. This ensures A-Frame always receives an array for array-type properties — avoiding arrayStringify being called on a raw string downstream (which would crash with "t.join is not a function").

Backport of https://github.com/aframevr/aframe-inspector/pull/855

### DIFF
--- a/src/editor/components/elements/PropertyRow.jsx
+++ b/src/editor/components/elements/PropertyRow.jsx
@@ -68,14 +68,29 @@ export default class PropertyRow extends React.Component {
         ? props.entity.getDOMAttribute(props.componentname)?.[props.name]
         : props.data;
 
+    const isFreeInputArray =
+      type === 'array' &&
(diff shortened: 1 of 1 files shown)


vincentfretin:
Oh you already did a similar fix in 
https://github.com/3DStreet/3dstreet/commit/2162f9e165ef77ab85253eaf2ee22815c3923f0b actually, interesting, closing for now.
Your change doesn't trim after the split(','), so it may or may not have an issue with "one, two, three".

kfarr:
thanks yeah this is one of many bugs I found and solved and forgot about

vincentfretin:
I modified it with extra goodies in #1573 to backport some changes I did in aframe-inspector/aframe-editor you may benefit in future features.

PR #1573, opened 2026-04-25
Support editing properties of type selector and selectorAll

vincentfretin:
Add support for editing a property of type selector and selectorAll types.
Also refactor the previous fix https://github.com/3DStreet/3dstreet/commit/2162f9e165ef77ab85253eaf2ee22815c3923f0b that added support for type array without oneOf.

That backports changes from
- part of https://github.com/c-frame/aframe-editor/commit/5af818ceedeebf775b2b3c7f3c98f92546f36ae1 to support resetting a property by the entityupdate command, I didn't backport the ui for it though.
- https://github.com/c-frame/aframe-editor/pull/52 except we can't edit the id in 3dstreet
- https://github.com/c-frame/aframe-editor/pull/69

### DIFF
--- a/src/editor/components/elements/PropertyRow.jsx
+++ b/src/editor/components/elements/PropertyRow.jsx
@@ -63,49 +63,57 @@ export default class PropertyRow extends React.Component {
       type = 'boolean';
     }
 
-    const value =
-      type === 'selector'
-        ? props.entity.getDOMAttribute(props.componentname)?.[props.name]
-        : props.data;
+    const isSelectorType = type === 'selector' || type === 'selectorAll';
 
+    const value = isSelectorType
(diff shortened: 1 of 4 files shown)


vincentfretin:
@claude about the e.stopPropagation() in onKeyDown, I did it similar to NumberWidget without thinking too much, you're right that it prevents the event to emit on the document and that's on purpose, but actually we also have the shouldCaptureKeyEvent in shortcuts.js that return false if we're in an input, so even without the e.stopPropagation(), the event will be discarded and won't trigger the delete confirmation of the selected entity if we press Delete inside the input. I'd say we keep it, that's fine, and ctrl+z and arrows actually works correctly with it. Esc doesn't work as you describe with or without it.

PR #1574, opened 2026-04-25
EntityReparentCommand: call updateLocalTransform synchronously

vincentfretin:
EntityReparentCommand: call updateLocalTransform synchronously and not in the loaded listener so that when historychanged event is emitted the entity already have the updated position/rotation/scale.

Backport of 
- https://github.com/c-frame/aframe-editor/pull/70
- https://github.com/c-frame/aframe-editor/pull/71

I don't think you were impacted by the issue because you debounce the save after a few seconds so the loaded event was executed before it saved the changes.
But if you queue changes to do incremental save or do collaborative broadcast right away when the command is executed on historychanged event and the user didn't update anything on the entity after reparenting, you ended up with the old position/rotation/scale saved.


vincentfretin:
@claude why do we wait for loaded event in createEntityFromObj to set the components? I couldn't find an explanation, the src/json-utils_1.1.js was renamed from src/json-utils.js before? I'm not sure. The components should be set right away in createEntityFromObj in my opinion, this is what we do in aframe-editor were the equivalent of createEntityFromObj is objectToElement https://github.com/c-frame/aframe-editor/blob/45633609038acb40624cd409ee3f19a6f9bf1cf5/src/lib/entity.js#L708-L746

vincentfretin:
@claude recreatedEntity.object3D exists right away, it's created in the constructor of the a-entity web component, see
https://github.com/aframevr/aframe/blob/d7a10b14ada468b8b149bf8d6f1b741f81f79977/src/core/a-entity.js#L34

with the current version of aframe we're using:
```js
el=document.createElement('a-entity')
el.setAttribute('position', '1 2 3')
el.setAttribute('position', '1 2 4')
el.components.position.data
{x: 1, y: 2, z: 4}
el.object3D.position
{x: 0, y: 0, z: 0}
document.querySelector('a-scene').appendChild(el)
el.object3D.position
{x: 1, y: 2, z: 4}
```

the 3 lines 
```js
    entity.object3D.position.copy(newLocalPosition);
    entity.object3D.quaternion.copy(newLocalQuaternion);
    entity.object3D.scale.copy(newLocalScale);
```
are actually useless because the transform will be changed by the components init

vincentfretin:
@kfarr do you know why the components and mixin were set in a loaded listener?
You will need to test that PR for sure for any regression.

I see other loaded listeners that set geometry height or atlas-uvs in
aframe-streetmix-parsers.js and street-generated-stencil.js
and one for segmentEl that set street-segment in AIChatTools.jsx
that may be suspicious or not of why those exist.

vincentfretin:
> @kfarr do you know why the components and mixin were set in a loaded listener? You will need to test that PR for sure for any regression.
> 
> I see other loaded listeners that set geometry height or atlas-uvs in aframe-streetmix-parsers.js and street-generated-stencil.js and one for segmentEl that set street-segment in AIChatTools.jsx that may be suspicious or not of why those exist.

We probably set the components in a loaded listener because some components accessed object3D.position/rotation before the entity was appended to a parent like I described in https://github.com/3DStreet/3dstreet/pull/1574#issuecomment-4319948362 but I think we changed that in all components already. I asked claude to search for the pattern of setting position/rotation/scale components, then direct access/write to object3D.position/rotation/quaternion/scale or the reverse on freshly created entities before appendChild, it find only the case in #1576

kfarr:
@vincentfretin in general components and mixin were set in a loaded listener for dealing with the async loading of managed street parent / segment / children. Hacks on top of hacks, so it is not ideal, and perhaps it is not necessary anymore. I will do some testing on this and let you know.

kfarr:
I get a `Cannot read properties of undefined (reading 'geometry')` when loading a scene with these changes.

http://localhost:3333/#/scenes/437d4d67-cd50-418e-b944-8fdb7cb2b2b4

I presume something is trying to do an operation on that geometry before it exists; It may be related to Managed Street which has known issues and is on the list to address.


Uncaught runtime errors:
×
ERROR
Cannot read properties of undefined (reading 'geometry')
TypeError: Cannot read properties of undefined (reading 'geometry')
    at n.update (http://localhost:3333/dist/aframe-street-component.js:42044:49)
    at n.initComponent (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1025166)
    at n.updateProperties (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1024879)
    at n.nC (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1024259)
    at new n (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1030207)
    at bC.initComponent (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1038970)
    at bC.updateComponent (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1040525)
    at bC.updateComponents (https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1040344)
    at https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1038522
    at https://cdn.jsdelivr.net/gh/aframevr/aframe@6a054e85953bbcc71854bcd64f9abde0897be649/dist/aframe-master.min.js:1:1034322

vincentfretin:
Ah indeed I didn't have atlas-uvs in the scene I tested. I tested another scene before retrieving your change and indeed I have the error. Looking at node_modules/aframe-atlas-uvs-component/index.js the atlas-uvs component has `dependencies: ['geometry']`
it correctly add geometry component first
Capture d’écran du 2026-04-30 10-36-26
but it's not initialized yet
Capture d’écran du 2026-04-30 10-42-00
there is no mesh at that time. The components are initialized later with the load event when it's appended to the DOM
Capture d’écran du 2026-04-30 10-49-46
Capture d’écran du 2026-04-30 10-54-41
and it indeed take the order defined in the mixin, so your fix is indeed correct.
With your fix:
Capture d’écran du 2026-04-30 10-56-26

vincentfretin:
See the last changes, can you please test again to see if all is okay with the geometry and atlas-uvs, related to the generated clones and stencils.

kfarr (approved):
passes all the manual QA I can throw at it!

Issue #1561, opened 2026-04-18
[generator] remove: inpaint / outpaint


kfarr:
unused feature, dedicated bfl api that we could remove

instead, we could retain inpainting by move the mask field as optional under the "advanced" dropdown in modify tab

PR #1585, opened 2026-05-01
remove bfl, inpaint and outpaint

kfarr:


kfarr:
https://github.com/3DStreet/3dstreet/issues/1561

PR #1589, opened 2026-05-02
guard atlas-uvs.update() in cloneMixinAsChildren when not yet mounted

vincentfretin:
That was missed with the manual testing in #1574 I just created a new scene importing the default example from streetmix and got the error atlas-uvs wasn't defined.

cloneMixinAsChildren is called with parentEls created via createStencilsParentElement that are not yet mounted (only appended to the segment later). The atlas-uvs component on placedObjectEl isn't initialized in that case, so calling update() on it threw TypeError on parking-lane sideways/angled variants.

Guard with optional chaining; when not yet initialized, atlas-uvs's init will read the current geometry.height once the subtree is mounted.

### DIFF
--- a/src/aframe-streetmix-parsers.js
+++ b/src/aframe-streetmix-parsers.js
@@ -22,7 +22,10 @@ function cloneMixinAsChildren({
 
     if (length) {
       placedObjectEl.setAttribute('geometry', 'height', length);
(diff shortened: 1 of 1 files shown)


vincentfretin:
the one in street-generated-stencil.js is ok, the entity is attached to the DOM

PR #1590, opened 2026-05-02
Managed street code clean up and bug fixes

vincentfretin:
In prevision to working on grouping and reparenting again, fixes issues related to leaked listeners and MutationObserver not disconnecting. See each commit description for review.

PR #1601, opened 2026-05-03
add doc how managed-street and its child segments are wired together

vincentfretin:
summary doc of my session yesterday doing #1590 that may be useful for next time Claude need to understand that part of the code base to make no mistake. Also useful for a human. :D


kfarr:
great writeup thanks!

PR #1606, opened 2026-05-06
fix: remove validation guards from SegmentAddCommand.execute, validate at caller

vincentfretin:
Validation belongs at the call site, not inside execute(). Remove all three guards from execute() (streetEl null check, segment.type check, segmentIndex range check). Add segmentIndex range validation for add-segment to managedStreetUpdateHandler where it belongs.


vincentfretin:
@claude can you rebase that branch and add back 
```js
if (!streetEl) {
  console.error(`[segmentadd] street ${this.streetId} not found`);
  return;
}
```
in reality it should never happen, the check is done before calling execute.

kfarr (approved):
thanks was able to reproduce console errors on prod that were resolved this this fix

PR #1613, opened 2026-05-11
Navigation

diarmidmackenzie:
Draft PR, just to help see scale of code diff.


kfarr:
Hi @diarmidmackenzie thanks here's a manual review:

Nice things:
- YES! Zooming into the current cursor meets ground position feels natural
- YES! Dynamic of zooming into a ground position that continues to bring camera to street level works great
	- Many users naturally will keep zooming in
- YES! WASD feels nice
	- difference in WASD on top-down mode vs. "streetview" mode
	- btw what do we call these modes?

For Improvement:
- Show centerpoint of the circle from which the camera is rotating
	- It's not always clear around which point the camera is rotating
- FOV upon zoom in to ground starts pretty zoomed in, can it start with a wider FOV?
- Ground click to move (teleport controls); upon moving FOV expands wider
- Set the camera rotation centerpoint upon cursor click
	- Currently it is not clear how the rotation centerpoint is decided; ideally it possible to set that centerpoint by clicking
- "False streetview mode":
	- When zoomed out (above certain elevation) but angle is less than certain angle, then the UI reverts into "streetview" mode
	- But the camera is really zoomed out so to the user it still feels like topdown mode, not streetview
	- This happens unintentionally when a user is rotating about an axis as during the rotation the camera position falls close to the horizon
	- Impact: zoom-in speed isn't as expected, rotation behavior isn't as expectd
	- Instead:
		- Maybe Streetview should be restricted to certain elevation near the ground? Or streetview cannot be entered from a rotation event, only from a zoom event?
	- How do we know we're in "streetview" -- the letterbox indicator is pretty helpful
- Streetview below-ground
	- In hilly areas, sometimes zooming in to an area above the elevation of the x,y,z = 0 origin the user can enter "streetview" below ground
	- Instead, Streetview should be at y = 0 or google maps surface elevation (raycaster for this, need same for snap to ground object placement)
- Regression - rotate via shift should be togglable on and off while panning is in progress (not just pressing shift before panning) restore this ability which exists in main branch (this means also that unpressing shift will stop rotating)
- Other misc:
	- No momentum, clunky and slow feeling instead of smooth and fast
	- Needs more defining of error case, but there are still zoom smoothness issues
	- In rare cases zooming out from street-view to "top-down" mode results in unexpected camera location
	- But the general direction of the movements is correct

How to know objectively this is working?
- User test, 3 example tasks:
	- Zoom into a street perspective in the middle of the intersection
	- Rotate the street scene to be similar to 45º paperboy style orientation
	- rotate the camera to point to the distant city skyline
	- (Provide the base scene file and example output image, allow user to try doing all 3 tasks with old method and with new method)
- Funnel test - confirm with Anita how this is represented

How to move forward:
- Diarmid is at 3/4 of the budgeted time for this task
	- In a perfect world, these changes are well developed / human reviewed code / architected so they don't revert previous things and only improve what we have
	- We are not there yet, 
	- It will probably take more than 5 hours of Diarmid's time to get to that point

Other to-dos before merging:
- Need to remap keyboard shortcuts for key editor tool modes
- Create video demo
- Update editor movement documentation

src/editor/components/Main.jsx
```diff
@@ -76,10 +80,12 @@ export default function Main() {
...
             <SceneGraph scene={scene} selectedEntity={state.entity} />
             <RightPanel entity={state.entity} />
-            <div className={`clickable ${styles.primaryToolbarDock}`}>
+            <div
+              className={dockClass(`clickable ${styles.primaryToolbarDock}`)}
+            >
               <PrimaryToolbar />
             </div>
-            <div className={`clickable ${styles.actionBarDock}`}>
+            <div className={dockClass(`clickable ${styles.actionBarDock}`)}>
```
vincentfretin: We use `classnames` in the code base, I think this can be written as:
```js
<div className={classnames("clickable", styles.primaryToolbarDock, {[styles.pedestalMode]: isPedestalMode})}>
```

.husky/pre-commit
```diff
@@ -1 +1 @@
-lint-staged
+npx lint-staged
```
vincentfretin, line 1: Did you have an issue with the pre-commit hook? We don't need to prefix with npx since husky 9.1.1 https://github.com/typicode/husky/releases/tag/v9.1.1 and we use husky 9.1.7 currently.

src/editor/components/Main.jsx
diarmidmackenzie: Thanks.  Code is not ready for review yet - AI-generated for UX testing purposes.  Still needs work to align with existing codebase conventions.

.husky/pre-commit
diarmidmackenzie: Yes, hit problems on Windows.  Can look into this further

diarmidmackenzie:
Thanks Kieran.  Based on this + my own notes, I've compiled the following list of open items.

## Bugs / missing features

B1 - Double-click logic improvements (phase 4)
B2 - Plan view should be a button, not a mode selection. Possibly combine with a compass indicator/control, like in Google Maps.  Needs design input, D1
B3 - Visible signal to the user when they are inside / outside a finite scene boundary.  Needs design input, D2.
B4 - Height for swoop zoom is controlled by y co-ordinate. Should be mesured above street level (for scenes where street level is not y=0)
B5 - FOV zoom lost on teleport
B6 - Should be able to transition between rotate and truck while dragging (Shift state should not latch on mouse-down)
B7 - "Zoom smoothness issues" (needs clarifiying)
B8 - "In rare cases zooming out from street-view to "top-down" mode results in unexpected camera location" - needs clarifying

## Code tasks

C1 - Review overall code for quality & consistency with existing codebase.
C2 - Use THREE.OrbitControls or custom code?  (current implementation used custom orbit code; TBC if there are good reasons not to use THREE.OrbitControls)


## Scope questions

S1 - FPS / Pointer lock mode?
S2 - Custom rotation center point (set by user)
S3 - Orthogrphic mode: retained or retired?
S4 - Migration of existing users: UI toggle to keep old controls?


## Design questions

D1 - Design for plan view / compass button
D2 - Design for indicator when user is inside/outside finite scene boundary
D3 - Design for rotation center indicator
D4 - Starting FOV when landing on ground: should it be wider?
D5 - UX for setting rotation center point (if S2 is in-scope)
D6 - Restrictions on "Streetview" mode above a certain elevation
D7 - Indicator for "Streetview" mode: letterbox OK, or something else?
D8 - Names for different modes "Maps" / "Streetview" ??
D9 - Momentum pattern for WASD controls
D10 - Remap of WASD key pre-existing functions.  So far, I picked
  - W → T (translate mode)
  - S → L (scale mode)
  - D → C (clone)

## Codebase questions

Q1 - Tests alongside code, or in top-level tests folder. README says the former, codebase actually does the latter.  Align?
Q2 - Dev Experience issues on Windows: rolldown and husky.  Fix these?


## Other tasks

O1 - User testing.  At what stage?  Early or late?  Could provide significant feedback on entire direction, so soon is good.  But prototype needs to be "good enough".
O2 - Video demo
O3 - docs updates
O4 - Deeper UX testing of the final implementation

.husky/pre-commit
vincentfretin: Don't bother to look into this further, if using npx is working for everyone, good to keep it. I just wanted to know if you had a real issue or if the AI changed that for some reason.

PR #1616, opened 2026-05-12
Dragndropglb

kfarr:
for review notes as needed


kfarr:
Thanks @vincentfretin this is a great start, super smooth drag-and-drop, immediate-value experience feels right. I still need to test quota logic but here is initial feedback:

things to make this complete:
- +1 on the drag-from-assets-panel todo, important case to support
- +1 on the thumbnail todo, also important

details modal for mesh:
- as a user I can edit additional information / settings about a model; this could accessed when clicking on the mesh in the assets panel to show the "gallery details modal for custom models"; also could be access from the entity properties sidepanel if an object is selected that has an associated cloud-stored model, the user can click a button to show the same "gallery details modal for custom models"
  - this modal allows changing the display name of the asset, it shows the preview image (thumbnail) for an asset, it shows other metadata, some of which might be editable, some not, some of which comes from catalog.json like bbox, facing direction, category, attribution, actions: place in scene, download, delete (later replace / re-optimize), nice to have embedded gltf viewer, nice but not blocking

for kieran to do:
- allow hard deletes instead of soft deletes by updating firebase storage rules. this is a little hard for me to accept because i wanted to make the UI super fast to delete things and allow user to undelete later if they make a mistake; but it probably makes sense to use hard deletes for now, and later restore undelete once we actually have time to make a full UI for that and a cloud cleanup script, etc.

other things:
- copy/paste to another scene (wait)
- still haven't tested quota stuff
- still need to test with larger / more complicated models
- still to think about how to support the streetplan use case; I agree we can wait for a future PR to do this. I am rethinking whether a user friendly "shortname" is actually needed, or if we can just use the asset UUID as a global unique identifier that they include in their mapping code instead. There's also an interesting crossover with the need to create a script to populate assets like that on dev server / prod server, and also the overlap with the existing catalog.json in our repo anyway.

vincentfretin:
About the soft delete, I can put it back. If the rule is to not count it in the quota (you don't count it in your quota-audit script) then there is no issue with it. If soft deleted models count in the quota and you can't see what takes up your quota, that's an issue. We can do a trashcan (just filter on soft deleted), a restore action and delete permanently action later like you suggest. We could also do a cleanup script that permanently remove after 30 days.

vincentfretin:
I created a mesh details modal with model-viewer in an iframe, with only the display name editable, and show read only the other metadata.
I didn't add bbox, facing direction, category, attribution, not sure how to present that and what to do about them.

The drag and drop from panel is working, the model thumbnail generated client side is working as well.
We have a minimal working version here.

vincentfretin:
So in that version, we don't save the original glb, but only the optimized glb. And there is not really a local glb mode because we optimize and upload right away on drop.
We could add the alternate flow with the modal to add locally the model, and option to not optimize (for the vehicle wheels mesh names) if we go the File->Import route?

kfarr:
good stuff:
  - thumbnails are awesome! amazing what can be done on the client side
  - drag and drop / upload experience is great! i don't mind the "flash" -- it is helpful for me since I know it is loading from the cloud that it is working; not sure if users will know that or care
  - I like the model modal with the iframe model viewer, it's magical
  - i like the current set of the model modal fields that are exposed, perhaps we can just keep those for now and expose more as needed

things to fix:
  - optimization on already optimized files / webp issue - see claude writeup below
  - the image modal has forward / back buttons / keys - not showing in model modal
  - discussion: as a user if i upload a model, then i delete it from my assets but not my scene:
	  - because of soft delete the asset will still be visible in my scene; i'm not sure if users expect this behavior;
	  - the UI in the sidepanel does not indicate that the model has been deleted in cloud storage, in fact it still says "Uploaded to cloud, Asset source: your cloud", perhaps it should be red and say deleted? and eventually (we can defer) it can be "undeleted" if it hasn't been hard deleted yet
	  - this does remind me of vincent's original idea to warn if an item is still placed in a scene
	  - this issue is compounded by soft delete which means the object may "disappear" to the user 30 days later during trash collection and surprise them
	  - Therefore: the task is at least to update the deleted state when the model is selected in a scene
  - discussion: only uploading optimized glb, not keeping original -- yes this may eventually be a problem for some users, it would be nice to be able to disable optimization; Yes one idea could be a separate upload flow with an option to disable optimization; or it could be "advanced import" modal which shows the local unmodified model in the embedded model viewer with a few options on the right hand panel including optimization toggle, and then you can choose to go ahead and import or cancel the import;

kieran to-do's:
  - should i re-upload the security rules - were they changed for soft delete or should i revert that do disallow hard delete from user?
  - we should revisit the token access scheme for the model / reminder to me if we chose the proxy method to double check access rights or if we deferred that
  - will the public/model-viewer.html need any additional setup in hosting or will it "just work" when we do firebase hosting deploy to resolve on app.domain/model-viewer
  - design of [2] isn't so hot, there's a lot of layers of text here; could use (+) symbol from new layer for Upload; could merge cloud storage % indicator with line starting "277 items"; 
  - need to have designs for free / approaching limit / at limit / paid / paid approaching limit / paid at limit upsell flows [3]
  - double check storage quotas - is 100mb too generous for free? takes a long time to fill with glb objects; this may change quickly with splats, even output from "sharp" ml from single shot image to splat can be 10mb+ per splat output


### [1] Claude's writeup -- Optimization makes already-optimized files *bigger*

Tested with two GLBs (see link to zip at the bottom) that were already Draco + WebP compressed (output of an earlier gltf-Transform v3.10.1 pass). After upload through the current pipeline (`src/editor/lib/asset-upload/optimizeGlb.js`):

| File | Source | Uploaded | Delta |
|--|--|--|--|
| `dump_truck.glb` | 284 KB | 1.18 MB | **4.1× larger** |
| `dirty_traffic_cone.glb` | ~3 MB | ~4.9 MB | **~1.6× larger** |

For the dump truck the mesh is basically unchanged (65,512 → 64,734 vertices — `simplify` did almost nothing thanks to the 0.001 error tolerance). The bloat is entirely in the **texture re-encode**: 1024×1024 WebP went from **145 KB → 1.04 MB** (7×).

Root cause: `textureCompress({ targetFormat: 'webp', resize: [2048, 2048] })` doesn't pass a `quality` option, so the canvas-fallback WebP encoder uses browser defaults that are much higher quality than the source's original encode.

Suggested fixes (in priority order):
- **Pass an explicit WebP quality** (e.g. `quality: 0.8`) to `textureCompress`. Probably the single highest-leverage change.
- **Bail if output ≥ input.** After `writeBinary`, compare byte lengths and just upload the original if optimization made it bigger. Cheap insurance.
- **Detect "already optimized"** (Draco + WebP in `extensionsUsed`) and either skip the pipeline entirely or run a lighter variant.
- **Reconsider `simplify` in the default pipeline.** It's a destructive op (ratio 0.5) that's appropriate for high-poly scans but risky for hand-authored or already-decimated assets. Right now the 0.001 error tolerance is the only thing keeping it from butchering output — that's fragile.

[cone-and-truck-optimized-models.zip](https://github.com/user-attachments/files/27787603/cone-and-truck-optimized-models.zip)

[2] design current

image

Instead, at 80%+ should indicate getting close to limit

[3] 
image

Local only / will not persist is closer to an error state than a neutral event, could be red or better designed for paid upsell for reaching storage limit

kfarr:
@vincentfretin I've made a pretty big change to make the assets into a shared component for the other "apps". I put it in a [draft pr](https://github.com/3DStreet/3dstreet/pull/1625) and I plan to merge it in. Then I will start working on these in rough priority order, feel free to provide more feedback:

- [x] * bugs
  - [x]   * re-upload the security rules on dev
  - [x]   * we don't want multiple file upload (1. Multiple files dropped into viewport — only the first is handled The file picker in AppMenu and GalleryPanel iterates all files correctly with for...of. The drop handler should too.)
  - [x]   * minor 2. Blob URL memory leak on successful retry In uploadAndPlaceAsset, createPlaceholderEntity allocates a blob URL and stores it in blobUrl. On a failed upload, this is not revoked (intentionally — needed for retry). But on a successful retry (existingEntity is passed, blobUrl = null), the original blob from the first attempt is never revoked. It outlives the session.The fix is to stash the original blob URL on the entity's Zustand slot and revoke it on retry success.
  - [x]   * minor 3. MAX plan users display as "TEAM" in the assets quota UI

- [x] 0 ability to upload without optimization - improve information display, defer option
     - [x] 0.1 fix optimization pipeline before storing anything "permanent"
          - [x] pass explicit WebP quality (0.85) to `textureCompress` — fixes 4-7× bloat on already-WebP textures
          - [x] bail-if-bigger: after `writeBinary`, compare byte lengths; if optimized ≥ original, use original bytes and flag `optimizationSkipped: true, reason: 'not_smaller'` in metadata
          - [x] detect already-optimized: inspect `extensionsUsed` for Draco + WebP before running pipeline; skip entirely with `optimizationSkipped: true, reason: 'already_optimized'`
     - [x] 0.2 move `simplify` out of default pipeline
          - [x] removed `simplify`, `palette`, `flatten`, `join` from default path; also dropped `meshoptimizer` dep
          - [x] default pipeline is now: `dedup → instance → weld → resample → prune → sparse → textureCompress (webp, q=0.85) → draco`
          - [x] `simplify` (and `flatten/join/palette`) deferred to a later opt-in "aggressive" mode settings toggle
     - [x] 0.3 store original source alongside optimized source in Firestore + Storage
          - [x] `storageUrl` / `storagePath` / `size` — always the original source file (convention documented in assetsService.js JSDoc)
          - [x] `optimizedSourceUrl` / `optimizedSourcePath` / `optimizedSourceSize` — present only when optimization ran and produced a smaller file
          - [x] `optimizationMetadata` — always present on GLB uploads: `{ optimizationSkipped, reason?, inputBytes, outputBytes }`
          - [x] scene placement prefers `optimizedSourceUrl ?? storageUrl`; `data-asset-id` saved for future re-resolve on 403/404
          - [x] `GLB_MAX_BYTES` gate now checks original file size (not optimized)
          - [ ] `pipelineVersion` field — deferred; add when pipeline stabilizes
          - [ ] `derivedFrom` back-reference on optimized doc — deferred
     - [x] 0.4 quota counts only `storageUrl` (original source)
          - [x] `optimizedSource` is a platform-derived artifact and should not count against user quota
          - [x] update quota-audit logic and display to use original `size` only
          - [x] update storage rules / metadata tagging so quota scripts can distinguish the two
     - [x] 0.5 loader fallback: retry on 403/404 using `data-asset-id` → re-resolve from asset record
          - [x] catch `model-error` on entities with `data-asset-id`, fetch asset doc, retry with `optimizedSourceUrl ?? storageUrl`
          - [x] prerequisite for safe server-side re-optimization (overwriting storage path invalidates the old token)
          - [x] log fallback attempts to console for dev visibility
     - [x] 0.6 display dual-source optimization info in mesh details modal
          - [x] "Original: X MB → Optimized: Y MB (−Z%)" from `optimizationMetadata`
          - [x] if `optimizationSkipped: true`, show reason ("Already optimized" or "Optimization didn't reduce size")
          - [x] download buttons for both original and optimized when both URLs are present
     - [x] 0.7 analytics events: optimization pipeline telemetry
          - [x] emit on every upload: `file_size_input`, `file_size_optimized`, `compression_ratio`, `optimization_made_bigger` (bool), `optimization_skipped` (bool), `skip_reason`, `already_had_draco`, `already_had_webp`, `duration_ms`
          - [x] `optimization_made_bigger` flag is the early-warning signal for pipeline regressions
          - [x] consolidate with existing item 4 analytics task
 
- [x] 1 error state handling:
  - [x] Local only, model will not persist (instead of Local only -- will ...)
  - [x] local only should be red / error and reason it's local only -- no quota left?
  - [x] this file exceeds the 50 mb limit for glb uploading - show the error on upload attempt but still allow the glb to be displayed; should also display toast to make sure user knows (ie as a user i should be able to preview larger size files on my computer even if they're not going to be uploadable, esp. for photogrammetry)
  - [x] error/retry UX when upload or optimization fails partway
  - [x] the cloud model sidebar should indicate file state -- ie if it's deleted, show red (ie soft deleted, also in the case where the model doesn't load for some reason / not found / 403)           

- [x] 1.5 QOL improvements
   - [x] background color for auto screenshot
   - [x] restore animated placement cursor
   - [x] consider keeping existing mesh until optimized mesh is actually downloaded --> noticeable on slow connections
   - [x] allow for cancellation of upload from asset panel and sidepanel upload info

- [x] 2 upload process indicator which is confusing for non-mesh or any upload in other apps besides editor + mesh which shows sidebar -- what if upload is shown as a "new entry" in the gallery; it can only be 1 - we only allow one upload at a time                                                
    - [x] as a user i may see the right panel (if i'm using scene graph and drag and ddrop) or i'll have another object selected while mine is uploading, so the left asset panel will keep going from the same source
    - [x] as a user i may be using the generator but want to upload a file that i'll use for ai generation source and the only ui i have will be in the asset panel
    
- [x] 3 optimization on already optimized files + decide on appropriateness of destructive mesh simplification (see claude notes above)
     - [x] 3.5 QA on models of various types, esp. after optimization logic changes                         
- [x] 4 analytics events on the new flows (optimization on/off, upload success/fail, time-to-complete)
- [ ] 5 docs including explaining what "optimization" does 
- [x] 6 tests around the shared-assets service since it's the breaking-change core
- [ ] 7 design polish to fix "lots of things smushed together" issue for assets                      
- [ ] 8 design / implementation of upsell                                                            

 - LATER make ticket for advanced import modal to allow a user to opt out of optimization

vincentfretin:
Like we discussed, you could add a field originalSource on the asset doc that is the original glb with just the resized textures and nothing more just to avoid uploading a big model with 4k or 8k textures. So you optimize the dropped glb twice in parallel and upload. You can later tweak the optimizations parameters to regenerate the optimized model with different ratio value for the simplify step or no simplify step at all. Others steps you may want to disable is flatten/join and palette for some use cases.

vincentfretin:
Something to be aware of is the instance optimization that may create an InstancedMesh in the glb if it detects similar meshes in an empty, so the optimized glb will be using the EXT_mesh_gpu_instancing glTF extension. You can have it in the original glb as well if you exported in blender with GPU Instances checked in the glTF export settings Scene Graph section.
If you later create a glb of the scene, and use model-viewer that auto creates the usdz for iOS Quick Look you will have issues with it, USDZExporter currently doesn't handle the EXT_mesh_gpu_instancing glTF extension. What I do currently it creating a live glb to remove usage of the extension and then it creates the usdz, but that's not viable of some heavy models because of the memory used in that process that may crash the browser tab.

vincentfretin:
As you continue the PR, you can ask Claude to use the animated placement cursor that we use in the add layer panel as well.

vincentfretin:
For webp textures, I read that client side it's using the canvas browser api to create it and server side it's using the sharp library, so you may see some differences. The quality parameter is something to test indeed.

kfarr:
These top 2 bugs from latest claude review are not actually bugs, I chased them down. This looks to be good enough to release.

Deployment checklist:
- [x] deploy latest to dev server incl. functions and updated storage rules
- [x] re-test team plan quota after dev deploy
- [x] test pro plan quota on dev server
- [x] test and merge attribution
- [x] merge
- [x] deploy to prod: hosting, functions, storage rules

Fast follows:
- [x] docs including explaining what optimization does
- [ ] video for blog post and social
- [ ] blog post to include in next newsletter

Deferring to later:
- design polish to fix "lots of things smushed together" issue for assets
- design / implementation of upsell
- server-side advanced glb optimization (incl. advanced textures, destructive geometry operations, etc.)
- ability for user to re-initiate glb optimization with user-specified optimization settings

vincentfretin:
Yes, its claim that model-loaded/model-error events on the entity are emitted with bubbles: false is completely wrong, the event is emitted with bubbles true that is the default, that's also why we need to add `if (event.target !== el) return;` in every listener to not catch the wrong model-loaded event of a child.

kfarr:
ok @claude one more (hopefully final) code review before merging

kfarr:
ok deployed to prod and a quick smoke test shows it's working

PR #1638, opened 2026-05-24
Asset upload fast follows

kfarr:


vincentfretin:
That's a good idea to transfer the Blob to the iframe, this way we can upload the capture earlier.
But it's still waiting for the optimized glb before doing the capture I see `const blobToCapture = optimizedBlob || file;`, Shouldn't we create the thumbnail right away from the original blob?
I asked ChatGPT if there is a limit transferring a large blob via postMessage. Passing a Blob does a structuredClone of it to pass it to the iframe, probably okay for <50MB. You can also transfer a blob URL.

Create an object URL:

```js
const url = URL.createObjectURL(blob)
iframe.contentWindow.postMessage({ url }, "*")
```

Then inside the iframe:
```js
window.addEventListener("message", async (e) => {
  const res = await fetch(e.data.url)
  const blob = await res.blob()
})
```

We need to be sure to revoke the url after we're done with the capture or timeout of course.
That's something to test.

kfarr:
@claude ok actually ready for a review now

PR #1673, opened 2026-06-12
Generator: 360 → Splat from raw .insv (direct dual-fisheye vid2scene path)

kfarr:
## What

Adds a **"360 → Splat (vid2scene, .insv)"** model to the generator splat tab: a raw Insta360 `.insv` recording straight off the camera goes in, and vid2scene reconstructs **directly from the two fisheye sensor streams** — no Insta360 Studio stitch, and full ground/sky coverage (the equirectangular path's known weak spot: the ER stitch smears the nadir across the bottom pixel row, and the ER virtual rig never looks below −80°).

Single research-preview tier priced at Max (60 tokens) and **gated to the Max plan** while the path is calibrated on real footage — raw .insv runs ~900 MB/min, so uploads ride the Max tier's 5 GB per-file cap, and the processing carries Max-tier COGS anyway.

## Changes

- **`public/functions/replicate-models.js`** — `vid2scene-360` entry with a new `requiresPlan: 'MAX'` knob. The `insv_fisheye: true` knob rides the Modal enqueue body like every other pipeline knob.
- **`public/functions/replicate.js`** — `generateReplicateSplat` enforces `requiresPlan` via `resolvePlanForUser` (fresh custom claims; the needed secret was already on the function) and rejects with `permission-denied`. The flag must be explicit: the Modal worker stages the downloaded input as an extensionless file, so the pipeline's `.insv` auto-detection by extension never fires through this route.
- **`src/generator/splat.js`** — standalone dropdown entry plus three pre-sanded rough edges:
  - per-model `accept=".insv"` — `.insv` has no registered MIME type, so without an extension filter the file picker greys the files out;
  - per-model 5 GB upload ceiling — decimal 5 GB, exactly the MAX plan's per-file cap and the Storage-rules backstop (which already allows `application/octet-stream`); ≈5 min of X4/X5 footage;
  - a Max-plan pre-flight (via `getUploadQuota`) before the multi-GB upload, so a non-Max user fails in seconds rather than after the upload; fails open — the server gate is the backstop;
  - an extension pre-check at file-pick and at submit, so a wrong file fails before the long upload rather than 20 minutes into a job;
  - honest `application/octet-stream` contentType for `.insv` uploads instead of mislabeling them `video/mp4` (informational only — the pipeline sniffs the container with ffprobe).

## Failure handling (research preview)

The pipeline **fails .insv jobs early** when the factory lens calibration cannot be parsed (vid2scene `253e52c`) instead of silently degrading to the idealized lens model — real X4/X5 sensors are portrait-mounted (~90° roll), so a degraded run burns a full SfM+training pass on a rig that cannot register. On the app side, every failure finalization path now:

- **retains the staged source** (flipped back to private) instead of deleting it, so the failure can be re-run against the exact input — a `retainedSplatSources` doc drives a GC pass on the 10-min reconciler sweep (7-day retention);
- **posts to the existing Discord webhook** with the error, job/user ids, and retained path (`DISCORD_WEBHOOK_URL` added to the finalizing functions' secrets).

Success and user-cancel clean up immediately, as before.

## Backend prerequisites (deploy in this order)

Each layer rejects the new flag until the layer below it knows it:

1. **Cog image rebuild** from [`vid2scene-cog#insv-fisheye`](https://github.com/3DStreet/vid2scene-cog/tree/insv-fisheye) (pins upstream at `cog-phase2` = the [direct-.insv ingestion PR](https://github.com/3DStreet/vid2scene/pull/1) head) → push to `r8.im/kfarr/vid2scene`.
2. **`modal deploy modal_app.py`** from that branch.
3. **Functions deploy**, then this UI.

Smoke-test the backend before the UI: `modal run --detach modal_app.py --video-url <signed .insv URL> --insv --split` exercises the identical path with full pipeline logs.

## Notes

- This branch is the single UI/config commit rebased onto `main`; the earlier `insv-360-splat` branch carried the already-squash-merged `modal-splat-backend` history and can be deleted.
- The insv path internally floors its image budget at ~800 frame pairs × 18 virtual views ≈ 1,600 SfM images (Max-like COGS); repricing awaits a measured calibration run.

### DIFF
--- a/public/functions/replicate-models.js
+++ b/public/functions/replicate-models.js
@@ -201,13 +201,17 @@ const REPLICATE_MODELS = {
   // images reject the unknown kwarg and the job fails at the SfM stage.
   // RESEARCH-PREVIEW TIER: priced at Max (the insv path renders ~800 frame
   // pairs x 18 virtual views ≈ 1,600 SfM images — Max-like COGS until a
-  // calibration run says otherwise).
+  // calibration run says otherwise). MAX-PLAN ONLY: raw .insv files run
+  // ~900 MB/min, so the upload rides the MAX tier's 5 GB per-file cap, and
+  // the processing cost matches the top plan anyway. Enforced server-side in
+  // generateReplicateSplat via resolvePlanForUser (fresh custom claims).
   'vid2scene-360': {
     name: 'vid2scene 360 (.insv to Splat)',
     modelName: 'kfarr/vid2scene',
     type: 'splat',
     inputKind: 'video',
     provider: 'modal',
+    requiresPlan: 'MAX',
     assetSlug: 'vid2scene-splat',
     assetLabel: 'vid2scene Splat',
     attribution: {
--- a/public/functions/replicate-models.js
+++ b/public/functions/replicate-models.js
@@ -191,6 +191,39 @@ const REPLICATE_MODELS = {
       resolution: 1920
     },
     tokenCost: 60
+  },
+  // 360 → Splat: a raw Insta360 .insv recording in, reconstructed DIRECTLY
+  // from its two fisheye sensor streams (no Insta360 Studio stitch). The
+  // `insv_fisheye` knob rides the enqueue body like every other pipeline knob;
+  // it must be explicit because the staged Modal input file has no extension,
+  // so the pipeline's .insv auto-detection never fires. Requires the cog image
+  // built from vid2scene-cog#insv-fisheye (upstream pin cog-phase2) — older
+  // images reject the unknown kwarg and the job fails at the SfM stage.
+  // RESEARCH-PREVIEW TIER: priced at Max (the insv path renders ~800 frame
+  // pairs x 18 virtual views ≈ 1,600 SfM images — Max-like COGS until a
+  // calibration run says otherwise).
+  'vid2scene-360': {
+    name: 'vid2scene 360 (.insv to Splat)',
+    modelName: 'kfarr/vid2scene',
+    type: 'splat',
+    inputKind: 'video',
+    provider: 'modal',
+    assetSlug: 'vid2scene-splat',
+    assetLabel: 'vid2scene Splat',
+    attribution: {
+      model: 'samuelm2/vid2scene',
+      modelName: 'vid2scene 360 (.insv to Splat)',
+      sourceType: 'video'
+    },
+    pipeline: {
+      // The pipeline floors the .insv image budget at 800 anyway; state it.
+      target_framecount: 800,
+      training_num_steps: 30000,
+      training_max_num_gaussians: 2000000,
+      resolution: 1920,
+      insv_fisheye: true
+    },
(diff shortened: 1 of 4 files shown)


kfarr:
## Cloud validation findings + next steps (2026-06-13)

Ran the `.insv` path end-to-end on Modal (dev env, isolated image `kfarr/vid2scene-dev` — production untouched). Recording what we learned and what's left. **Not doing these now; this is a parking note.**

### ✅ Validated on real footage
- **Calibration parses correctly in-container** — factory lens calibration (X4 *and* X5 trailers) loads on the Modal worker, not just in local unit tests.
- **Rig geometry reconstructs cleanly** — on a real X5 clip, 63/85 rig frames registered with ~1,000–1,300 points/image before the run was cut off. The direct dual-fisheye method works.

### ⛔ Blocker: SfM exceeds the 2h Modal task timeout
- `TIMEOUT_S=7200` (2h). A real job spent ~90 min in feature matching (~1,530 images at `overlap=30`) + COLMAP *incremental* mapper (hardcoded `mapper="colmap"`, repeated global BA), and hit the 2h wall mid-BA — killed, no output. **Not a crash/bug — a timeout.** ~$2.70/run (SfM only, no training reached).

### Short-term — get ONE completed run (validate SfM → training → .ply)
We've never seen the back half finish. Temporarily either lower the image-budget floor (`vid2scene.py:622`, `max(target_framecount, 800)` → ~400) **or** bump `TIMEOUT_S` — just to confirm training + .ply export work. Test-only; keep separate from the production config.

### Long-term — Very High-level quality (12h+ jobs)
Quality is the goal, so **do NOT cut the image budget for production.** Instead:
- **Raise `TIMEOUT_S`** to ~12–14h.
- **Handle preemption (the real risk for long jobs)** — SfM has no intra-stage checkpoint, so a preempt restarts matching from scratch (observed once already). Long jobs need non-preemptible/reserved capacity or intra-SfM checkpointing.
- **Profile the SfM-vs-training split** before investing (TBD — not yet measured).
- **Reprice** — a 12h job ≈ $25–30 COGS, so the current 60-token tier is a test price; high-quality would be a premium tier.

### Notes on the frame-count knob (gotcha)
`target_framecount` is a **total-image budget ÷ 9 = frame pairs**, *not* a frame count, and it's floored at 800 (`vid2scene.py:622`). So runtime `--frames` can't shrink a job below ~85 pairs / ~1,530 images — the floor is the lever, and it needs a cog rebuild to change.

### Housekeeping
- Hetzner build box deleted. ✅
- Test clip still world-readable at `gs://dev-3dstreet.appspot.com/vid2scene-smoke-test/intersection.insv` — lock down/delete when done.
- Already on this branch: Max-plan gate, 5 GB cap, failed-source retention + Discord alerts. Pipeline repos (`vid2scene` / `vid2scene-cog`): X4/X5 trailer calibration parsing + fail-fast on unparseable calibration.

PR #1723, opened 2026-06-20
feat(perf): Low Power Mode toggle for demo/screenshare on fanless machines

kfarr:
## Why

On a MacBook Air M2, running a heavy 3DStreet scene (Google 3D Tiles + Gaussian splats + managed-street geometry) while simultaneously screensharing or recording and driving an external display pushes the machine past its GPU and thermal limit. The Air is fanless, so it thermally throttles within a couple minutes of a sustained demo, causing stutter and dropped frames.

The root cause is that A-Frame defaults to `devicePixelRatio` (2 on Retina), meaning the GPU renders **4x the fragments** vs. a non-Retina display. Since tiles, splats, and geometry all go through the same renderer, this is the highest-leverage single lever: capping pixel ratio to 1 roughly quarters per-frame fragment work and benefits everything at once.

This PR adds a **Low Power Mode** toggle in the View menu that applies that cap, wired to a persisted store preference so it survives reload.

## What's in this PR

- `src/store.js` — `lowPowerMode` boolean preference + setter, persisted to `localStorage`
- `src/aframe-components/low-power-mode.js` — new scene-level A-Frame component that reads the store, subscribes to changes, and applies/restores the pixel ratio cap via the Three.js renderer
- `src/index.js` — registers the component
- `index.html` — adds `low-power-mode` to `<a-scene>`
- `src/editor/components/scenegraph/AppMenu.jsx` — "Low Power Mode" checkbox under View menu (next to Show Grid)

## How to test

1. `npm start`, load a scene with 3D Tiles + a splat + street geometry
2. View → Low Power Mode (check it)
3. Canvas should look visibly softer/less sharp, FPS should jump noticeably in a heavy scene
4. Toggle off, quality returns
5. Reload — preference persists (localStorage)

## Known caveat to verify

The component attaches on `renderstart` if the renderer isn't ready at init time. Worth confirming the cap applies on first load (not just after toggling) in the master A-Frame build we use.

## What's next (not in this PR)

This PR is the minimal slice: one lever, globally applied. The next layer of tuning to add once this is validated:

### 1. Google 3D Tiles `errorTarget`
The `TilesRenderer` instance in `google-maps-aerial.js` has an `errorTarget` property (default: 16px screen-space error). Raising it to ~40 in low-power mode tells the tiles renderer to use coarser, cheaper LODs. This is the second-biggest lever for scenes with 3D Tiles coverage. The hook is already commented in `low-power-mode.js`:

```js
// --- Extension hook: Google 3D Tiles error target ---
// const tiles = this.el.querySelector('[street-geo]')?.components?.['street-geo'];
// if (tiles?.googleTiles) tiles.googleTiles.errorTarget = this.enabled ? 40 : 16;
```

Needs the `TilesRenderer` instance to be exposed on the `street-geo` or `google-maps-aerial` component (currently it's local to `init()`).

### 2. SparkRenderer splat budget
The Gaussian splat renderer (`SparkRenderer`) has a `maxPagedSplats` property for capping the in-memory splat budget for paged RAD files, and `lodSplatScale` for scaling splat density. Accessible via `STREET.splatDebug.snapshot()` today. Wiring these into low-power mode would meaningfully reduce splat render cost for scenes with large splat assets.

### 3. Shadow map resolution
`street-environment` hardcodes `shadowMapHeight: 2048` / `shadowMapWidth: 2048`. Halving to 1024 in low-power mode is cheap to add and reduces texture memory bandwidth.

### 4. Antialias
The `<a-scene renderer="antialias: true">` attribute is set at scene init and can't be toggled at runtime without a reload. Longer-term option: offer a "quality preset" that reloads with `antialias: false` baked in for low-power mode. Low priority — pixel ratio already does most of the work.

## Hardware context

This feature was motivated by the M2 Air's limits. The fix for that machine is this toggle. For reference on why hardware upgrade alone isn't the only answer: even a MacBook Pro M5 Pro (16-core GPU, active cooling) will benefit from this toggle during a max-load demo + simultaneous 4K screenshare + external display — thermal headroom is finite on any machine.

### DIFF
--- a/docs/low-power-mode-wip.md
+++ b/docs/low-power-mode-wip.md
@@ -1,0 +1,96 @@
+# Low Power Mode — WIP notes
+
+Status as of 2026-06-29. Branch `feat/low-power-mode`, PR #1723.
+
+Continuation notes for the splat-performance work layered on top of the original
+pixel-ratio toggle. The component is `src/aframe-components/low-power-mode.js`.
+
+## Problem being solved
+
+Heavy scene (Google 3D Tiles + Gaussian splats + managed-street geometry) on a
+fanless M2 Air sits at ~9 FPS and **gets worse as the scene loads**, because
+more splat LOD pages stream in over time.
+
+## Key finding: Spark LOD is not FPS-reactive
+
+Spark's `SparkRenderer` LOD is distance / screen-size driven against a **fixed
+splat budget**, not a frame-rate governor. There is no built-in "FPS dropped,
+shed detail" loop. On desktop-class hardware (the M2 Air is detected as desktop)
+the target is **2.5M splats**, and Spark fills toward it as pages load — which is
+why FPS degrades the longer the scene is open.
+
+Relevant Spark source (`node_modules/@sparkjsdev/spark/dist/spark.module.js`):
+- `defaultSplatTarget()` ~line 10361 → desktop 2.5M, iOS 1.5M, Android 1M, VisionPro 750K, Oculus 500K
+- `driveLod()` ~line 10372: `const maxSplats = (this.lodSplatCount ?? defaultSplatTarget()) * this.lodSplatScale;`
+- `pixelScaleLimit *= this.lodRenderScale;` ~line 10384 (sub-pixel cull threshold)
+- `maxStdDev` field → uniform copy ~line 10093 (Gaussian quad / fragment discard radius)
+- `SplatPager` built once with `maxSplats: this.maxPagedSplats` ~line 10443 (so
+  `maxPagedSplats` only bites at pager-creation time, i.e. when low-power is on
+  *before* the splat loads)
+
+All of `lodSplatScale`, `lodRenderScale`, `maxStdDev` are read live each frame,
+so changing them takes effect next frame with no reload.
+
+## Measured so far
+
+- Pixel-ratio cap to 1: **9 → 15 FPS** (the bulk of the win; global, immediate)
+- Tiles alone (3D Tiles + mesh, no splat): already 30 FPS — so the splat is the
+  bottleneck, not tiles or geometry.
+- Target: get the splat-heavy case from 15 → 30.
+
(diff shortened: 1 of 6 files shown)


kfarr:
## WIP update — splat LOD perf levers (pushed in 63009629)

Extended the toggle beyond the pixel-ratio cap after testing on a heavy scene (3D Tiles + splat + geometry) on an M2 Air. Full notes in `docs/low-power-mode-wip.md`.

**Key finding:** Spark's LOD is **not FPS-reactive** — it targets a fixed splat budget (2.5M on desktop) and fills toward it as pages stream in, which is why FPS *degrades the longer the scene is open*. The only way to make it respect a fanless GPU is to lower that budget ourselves.

**Measured:** pixel-ratio cap alone = 9 → 15 FPS. Tiles + mesh without a splat already hit 30, so the splat is the bottleneck. Goal is 15 → 30 on the splat-heavy case.

**Toggle now also applies (all live, no reload):**
- `lodSplatScale` 1.0 → 0.5 — LOD count target 2.5M → 1.25M (the fix for "loads toward max LOD"). Safe: renders fewer.
- `maxStdDev` √8≈2.83 → 2.0 — shrinks each quad, ~halves fill-rate. Safe: smaller, not culled.

**Console handle for tuning** (`STREET.lowPower.splat('lodSplatScale', 0.3)`, `.status()`, `.reset()`) lets us dial in values by eye.

**Gotcha worth flagging:** `lodRenderScale` at 2.0 made entire splats vanish at distance (raises the sub-pixel cull threshold). It's intentionally **console-only**, not in the toggle.

**Next steps:**
1. Dial in `lodSplatScale` (~0.25–0.35) + `maxStdDev` on a real scene, then bake the chosen numbers into `SPLAT_LEVERS`.
2. Decide whether to promote a gentle `lodRenderScale` (1.1–1.5).
3. Optional follow-up PR: an FPS-reactive auto-scaler that nudges `lodSplatScale` to hold a target FPS — the adaptive behavior Spark doesn't do itself.

PR #1733, opened 2026-06-24
building import in managed street

kfarr:
default building lot widths need work


kfarr:
Superseded by #1792, which carries this branch's building-import work forward but reads the canonical schemaVersion 34 `boundary` object (variant/floors/elevation) instead of the deprecated flat `leftBuildingVariant`/`rightBuildingVariant` fields, and migrates segment elevation from integer levels to meters in the same pass.

---

PR #1746, opened 2026-06-27
Reject degenerate splat reconstructions before charging (#1745)

### LINKED ISSUE #1745: vid2scene: failed SfM reconstructions are billed as successes (no sanity check on generated .ply)
## Summary

A vid2scene (video→splat) run that fails structure-from-motion is **billed as a success**: it charges the user 30 tokens, is marked `status: succeeded`, and saves a public, **unrenderable** asset. There is no sanity check on the generated `.ply` before charging/saving.

First hit in the wild during the vid2scene launch (2026-06-26): one user's two runs included one degenerate reconstruction; they were charged 30 tokens for an asset that won't display. Manually refunded.

## Root cause

vid2scene always emits a fixed **500,000-gaussian `.ply`** (41 floats/vertex → exactly `82,001,006` bytes every time — identical size across runs is expected, not duplication). When SfM fails (too few matches), it still emits a full 500k-gaussian file, but the geometry is garbage:

| | Good run | Failed-SfM run |
|---|---|---|
| NaN positions (first 2000 verts) | 0 | **114** |
| Coordinate extent | x[-10, -2.7], y[-17.7, -8.6], z[-3, -0.09] (~15-unit box) | x[**-2306, +2327**], y[-1839, +1415], z[-410, +822] (exploded) |

The pipeline treats this as success:
1. Output `.ply` written (NaNs and all).
2. Job marked `status: succeeded`, asset saved with `visibility: public`.
3. User charged 30 tokens.

The raw `.ply` won't render (NaN geometry), and the downstream cloudrun RAD/LOD step **correctly** skips it — `skipReason: "build-lod produced no .rad output"`, `phaseMs` has only `download`, no `convert`. The RAD skip is correct behavior and is **not** the bug; it's just the first place the failure becomes visible.

## Impact

- Users charged 30 tokens for unrenderable output, requiring manual refunds.
- Broken assets saved as public in the user's gallery.
- Silent: nothing flags the failure; it looks "successful" end to end.

## Proposed fix

Add a sanity gate on the generated `.ply` **before charging and saving** in the splat path (`public/functions/replicate.js`, around `saveSplatToGallery` / the token-charge step). Reject if the reconstruction is degenerate, e.g.:

- NaN/Inf ratio in vertex positions above a small threshold, and/or
- bounding-box extent beyond a sane cap (a real scene is ~tens of units, not thousands).

On rejection: mark the job `failed` (with a reason), **auto-refund** the tokens (reuse the existing `refundSplatToken` path), and do not save a public asset. This would have caught George's run automatically.

## Notes

- Source videos are unrecoverable for debugging past runs: `tempFilePath` = `users/{uid}/assets/splat-sources/*.mov` is deleted on success by `cleanupSplatTempFile`. If we want to debug failures, we'd need to retain the source on failure (or copy before delete).
- A degenerate-output check could also live on the modal/vid2scene worker side (fail the job there before returning a `.ply`), which avoids transferring/saving an 82 MB garbage file at all. Either location is fine; the Cloud Function gate is the smaller change.

kfarr:
Fixes #1745.

## Problem

A vid2scene (video→splat) run that fails structure-from-motion is **billed as a success**: it charges 30 tokens, marks the job `status: succeeded`, and saves a public, **unrenderable** asset. vid2scene always emits a full 500k-gaussian `.ply` even when SfM fails — but the geometry is garbage (NaN positions, bounding box exploded to thousands of units). There was no sanity check before charging/saving, so these required manual refunds.

## Fix

Add a sanity gate on the generated `.ply` **before the charge is kept and the asset is saved**, in `processTerminalPrediction` — the single chokepoint shared by the webhook and the scheduled reconciler, so every code path is covered.

- **`public/functions/ply-sanity.js`** (new) — a pure, dependency-free parser. `inspectPlyGeometry(buffer)` reads the PLY header + first ~2000 vertices and rejects a reconstruction when:
  - non-finite (NaN/Inf) positions exceed **1%** of sampled vertices (a good run has ~0; the failed run in the wild had ~5.7%), or
  - any bounding-box axis spans more than **1000 units** (real scenes are ~tens of units; the failed run spanned thousands).
- **`public/functions/replicate.js`**:
  - `fetchPlyHead()` range-reads only **~512 KB** of the file (never the full ~82 MB) — handles both staging locations: `gs://` (vid2scene, our bucket) via a ranged Storage read, and an `https` CDN URL (SHARP fallback) via a `Range` fetch.
  - The gate runs after the save-claim but before `saveSplatToGallery`. On rejection it **auto-refunds** via the existing `refundSplatToken` path, finalizes the job as `failed` with a user-facing reason, writes a `failed` `generationLog` row (keeping `metrics`/`estCostUsd` so rejected-but-costly runs stay visible), and **deletes the rejected staged `.ply`**.
  - **Fails open**: any fetch/parse error returns `ok: true` so a legitimate save is never blocked by a parser quirk (the prior behavior saved everything regardless).

## Design notes

- Gate placed in the Cloud Function (the smaller change the issue suggested) rather than the Modal/vid2scene worker. A worker-side check could avoid transferring the 82 MB garbage file at all — a reasonable follow-up.
- Thresholds (`NAN_RATIO_LIMIT`, `MAX_EXTENT`, `SAMPLE_VERTS`) are named constants in `ply-sanity.js`, easy to tune.
- Not addressed here (called out in the issue as separate): retaining the source video on failure for debugging — sources are still deleted by `cleanupSplatTempFile`.

## Tests

`test/core/ply-sanity.test.js` covers healthy, NaN-heavy, exploded-bounds, and unparseable/empty inputs. All pass via `mocha`.

## Status

Draft — opening for review of the approach and thresholds. I have not been able to run the Cloud Functions' own eslint/integration suite in this environment (the `public/functions` lint config needs its own installed deps); the new files pass `node -c` and follow the surrounding style.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJT7SR6ZhutUEnrkSqbAgY)_

### DIFF
--- a/public/functions/ply-sanity.js
+++ b/public/functions/ply-sanity.js
@@ -2,11 +2,17 @@
 //
 // vid2scene (and other splat models) always emit a fixed-size .ply even when
 // structure-from-motion fails: a failed reconstruction still produces a full
-// 500k-gaussian file, but the geometry is garbage — NaN/Inf positions and a
-// bounding box that explodes to thousands of units instead of the ~tens of
-// units a real scene occupies. That file won't render, yet the pipeline used
-// to treat it as a success: charge the user, mark the job succeeded, and save
-// a public (unrenderable) asset. See issue #1745.
+// 500k-gaussian file, but the geometry is garbage — its vertex positions are
+// peppered with NaN/Inf. That file won't render, yet the pipeline used to
+// treat it as a success: charge the user, mark the job succeeded, and save a
+// public (unrenderable) asset. See issue #1745.
+//
+// The reject signal is the NaN/Inf ratio, which is scale-invariant: a real
+// reconstruction has ~0, a failed one has percent-level. Bounding-box extent
+// is computed for observability but is NOT a gate — a legitimately large scene
+// (e.g. a 22M-gaussian drone scan measured at ~3400 units, all-finite) is
+// indistinguishable by absolute extent from a garbage explosion, so gating on
+// it false-rejects real scenes.
 //
 // This module is a pure, dependency-free parser so it can be unit-tested in
 // isolation. It inspects only the header plus the first N vertices of the
@@ -19,10 +25,11 @@
 // without letting a degenerate file through.
 const NAN_RATIO_LIMIT = 0.01;
 
-// Reject if any axis of the bounding box (of finite vertices) spans more than
-// this many units. Real scenes are ~tens of units; a failed SfM run spanned
-// thousands (e.g. x[-2306, +2327]). 1000 is a generous "this is clearly
-// exploded" cap, not a tight fit.
+// Advisory only — flag (do NOT reject) when any axis of the bounding box spans
+// more than this. A failed SfM run explodes to thousands of units, but so do
+// legitimately large scenes (a real 22M-gaussian drone scan measured ~3400
+// units, all-finite). Absolute extent can't separate the two, so we surface
+// `extentExceedsAdvisory` in stats for monitoring and let the NaN check decide.
 const MAX_EXTENT = 1000;
 
 // How many vertices to sample from the start of the payload. Matches the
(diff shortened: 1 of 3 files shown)


kfarr:
@claude can you do a PR review on this please?

PR #1747, opened 2026-06-29
feat(i18n): react-intl localization (es, fr, pt-BR) (#656)

### LINKED ISSUE #656: internationalization
benefit: many south america users, spanish is the primary language of many users
vincent has experience with react internationalization

kfarr:
Closes #656.

Ships internationalization as a cheap, mostly-automated acquisition/activation experiment, measured in PostHog against the English baseline. **It activates on merge — there is no runtime feature flag; the branch is the gate.**

Ships four languages: **English (`en`)**, **Spanish (`es`)**, **Brazilian Portuguese (`pt-BR`)** and **French (`fr`)**.

## What's included

**Plumbing (react-intl, per @vincentfretin's suggestion)**
- `IntlProvider` (`LocaleProvider`) driven by a `locale` value in the Zustand store, mounted at all editor React roots.
- Browser-language auto-detection on first load (`es-*` → `es`, `pt-*` → `pt-BR`, `fr-*` → `fr`, otherwise `en`).
- **Language switcher**: a `Language ▸` submenu in the editor **View** menu (`AppMenu`), listing each locale by its endonym. Selecting one switches instantly.

**Locale persistence**
- The choice is saved to `localStorage` and — when signed in — to the user's Firestore profile (`socialProfile/{uid}.locale`), so it follows the account across devices (`useProfileLocaleSync` applies it on sign-in). Because the field is on the user doc, the backend can read it to localize transactional emails (a separate Cloud Functions task; not in this PR).

**Measurement**
- Registers a `locale` PostHog super-property and emits a `locale_changed` event, so locale cohorts can be compared against the English baseline for activation / retention / conversion.

**Strings & catalogs**
- ~346 user-facing strings wrapped across modals, menus, toolbars, panels and sidebars with `FormattedMessage` / `useIntl` + explicit ids.
- Shared `commonMessages` catalog so duplicated buttons/labels are extracted and translated once (no duplicate strings); icons/emoji kept out of translatable text.
- `en.json` generated by formatjs extraction; `es.json` / `pt-BR.json` / `fr.json` LLM-translated with urban-planning terminology — 0 missing keys, 0 placeholder mismatches.

**Number & price formatting**
- `src/shared/utils/format.js` (`formatCurrency` / `formatNumber`, `Intl.NumberFormat`) — context-free so it works in both the editor and the generator (the generator has no `IntlProvider`). Applied to every price display (UpgradeModal tier price, yearly totals, AssetsPanelBody pill). e.g. en `$9.99` → fr/es `9,99 $`.

**Repeatable re-translation path**
- `scripts/i18n/translate.mjs` only re-translates new/changed keys (manifest-tracked), prunes removed ones, and has a `--check` drift mode.
- npm scripts: `i18n:extract`, `i18n:translate`, `i18n:check`.
- `.github/workflows/i18n.yml`: PRs fail on drift; pushes to `main` re-translate (using the `ANTHROPIC_API_KEY` secret) and auto-commit refreshed catalogs.
- Documented in `docs/i18n.md`.

## Validation

- `npm run lint` clean, `npm run dist` (production build) green.
- `npm run test:modern` passes, including locale-detection / initial-locale and number-format unit tests.
- A real **headless Chromium** harness (`test/editor/i18n/browser-harness/`) loads the actual catalogs through the actual react-intl setup, switches `en` → `es` → `pt-BR` → `fr`, and asserts rendered text matches — passes for all four.

## Notes

- String wrapping targets `.jsx` components. A few legacy property-editor widgets put JSX in `.js` files (`src/editor/components/widgets/*.js`); the formatjs extractor skips those until renamed to `.jsx`. Noted in `docs/i18n.md`.
- The switcher lives in the editor View menu (editor-only); auto-detection covers users who never open the menu. Surfacing language earlier (landing/sign-in) is a possible follow-up.
- Localized transactional emails are a backend follow-up; this PR stores the `locale` field that enables them.


.github/workflows/i18n.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+  i18n:
+    name: Check / refresh translations
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
```
vincentfretin:
```suggestion
        uses: actions/setup-node@v6
```

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,72 @@
...
+permissions:
+  contents: write
+
+jobs:
+  i18n:
+    name: Check / refresh translations
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
```
vincentfretin:
```suggestion
        uses: actions/checkout@v7
```

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
+          cache: 'npm'
+          cache-dependency-path: 'package-lock.json'
+
+      - name: Install dependencies
+        run: npm install
```
vincentfretin:
```suggestion
        run: npm ci
```

docs/i18n.md
```diff
@@ -0,0 +1,108 @@
...
+> those, so they are out of scope until renamed to `.jsx`.
+
+## Catalogs & the translation pipeline
+
+Catalogs live in `src/editor/i18n/locales/`:
+
+- `en.json` — **generated** by extraction; the English source of truth.
+- `es.json`, `pt-BR.json` — **generated** by translation.
+- `.translation-manifest.json` — bookkeeping: the English source each key was
+  last translated from, used to detect changed strings.
```
vincentfretin, line 102: Interesting, that's one way to do it. If that works, great. In i18n libraries documentation, I always see they promote using a platform that does the sync job, but I never saw in a solution to sync the files locally without a third party tool.
In a react project I used combination of json2po (apt install translate-toolkit) to create a pot file and msgmerge (apt install gettext) to sync the po files, then convert the po files to json with po2json. That's the only way I found at the pre-AI era to sync the translations, and I already had experience with the po files ecosystem, where it marks the translation as fuzzy if the original message changes and the translation needs to be updated. I generally did the translation manually with poedit editor and some copy paste from google translate. Now with LLM, they produces far better translation.

src/editor/components/elements/AssetInfoPanel.jsx
```diff
@@ -101,7 +103,7 @@ const AssetInfoPanel = ({ entity }) => {
               cursor: 'pointer'
             }}
           >
-            Cancel
+            <FormattedMessage id="assetInfo.cancel" defaultMessage="Cancel" />
```
vincentfretin: I would suggest to have a generic prefix buttons.cancel, buttons.save, buttons.validate, buttons.retry and so on, so it extracts only once the string to translate. You can ask the LLM to check for duplicate messages and eventually use a generic prefix if the context is not ambiguous (for example "Try" in a title or button can be translated as a noun "Essai" or a verb "Essayer" in French)

src/editor/components/modals/ProfileModal/ProfileModal.component.jsx
```diff
@@ -277,15 +334,20 @@ const ProfileModal = () => {
...
-                          })}
+                          <FormattedMessage
+                            id="profileModal.monthlyProRefill"
+                            defaultMessage="Monthly Pro refill: {date}"
+                            values={{
+                              date: new Date(
+                                new Date().getFullYear(),
+                                new Date().getMonth() + 1,
+                                1
+                              ).toLocaleDateString('en-US', {
```
vincentfretin: You need to use the locale from zustand store here. French users will be confused by the mm/dd/yy format, we read it as dd/mm/yy.
But probably use the FormattedDate component here.

src/editor/components/modals/ShareModal/ShareModal.component.jsx
```diff
@@ -295,39 +405,59 @@ function ShareModal() {
...
                 >
                   <Button
                     onClick={handleEmailShare}
                     variant="toolbtn"
                     className={styles.shareButton}
                   >
-                    ✉️ Email
+                    <FormattedMessage
+                      id="shareModal.emailButton"
+                      defaultMessage="✉️ Email"
```
vincentfretin: put the icon out of the message

src/editor/components/modals/ShareModal/ShareModal.component.jsx
```diff
@@ -295,39 +405,59 @@ function ShareModal() {
...
                 >
                   <Button
                     onClick={handleOSShare}
                     variant="toolbtn"
                     className={styles.shareButton}
                   >
-                    📤 Share
+                    <FormattedMessage
+                      id="shareModal.osShareButton"
+                      defaultMessage="📤 Share"
```
vincentfretin: put the icon out of the message

src/editor/components/modals/ShareModal/ShareModal.component.jsx
```diff
@@ -295,39 +405,59 @@ function ShareModal() {
...
                 >
                   <Button
                     onClick={handleDiscordShare}
                     variant="toolbtn"
                     className={styles.shareButton}
                   >
-                    💬 Discord
+                    <FormattedMessage
+                      id="shareModal.discordButton"
+                      defaultMessage="💬 Discord"
```
vincentfretin: put the icon out of the message

.github/workflows/i18n.yml
vincentfretin: You should update the versions in ci.yml as well.

vincentfretin (commented):
That's a big one, very nice to see the app internationalized!

.github/workflows/i18n.yml
kfarr: Done — bumped `actions/checkout` to v7 and `actions/setup-node` to v6 in both `i18n.yml` and `ci.yml` (13dbc17).

---

.github/workflows/i18n.yml
kfarr: Switched the i18n workflow to `npm ci` (13dbc17).

---

src/editor/components/elements/AssetInfoPanel.jsx
kfarr: Good call. Added a shared catalog `src/editor/i18n/commonMessages.js` (`defineMessages`) and pointed all duplicated strings at it — `Cancel`, `Retry`, `Delete`, `Duplicate`, `Focus`, `Rename`, `Surface`, `Documentation`, `Keyboard Shortcuts`, `Reset Camera View`, `Upgrade to Pro`, `Add layer`, `Sign in to 3DStreet Cloud`, `Share`, `Discord`, etc. That dropped the catalog from 365 → 346 keys with zero duplicate strings, so each is extracted/translated once.

I kept it to strings that are unambiguous in every context (your noun/verb point — e.g. "Try"), so context-sensitive words still keep per-component ids. Done in 13dbc17.

---

src/editor/components/modals/ProfileModal/ProfileModal.component.jsx
kfarr: Fixed — now uses `intl.formatDate(...)`, which formats in the active IntlProvider locale instead of hardcoded `en-US`, so the ordering is correct per locale (13dbc17).

---

src/editor/components/modals/ShareModal/ShareModal.component.jsx
kfarr: Done for all three buttons (Email / Share / Discord) — the emoji now lives outside the `FormattedMessage`, so translators only get the word (13dbc17).

---

docs/i18n.md
kfarr: Thanks for the context — the gettext/`msgmerge` fuzzy-marking workflow is exactly the prior art I was modeling. The manifest here (`.translation-manifest.json`) plays the role of the `.pot`/fuzzy tracking: it stores the English source each key was last translated from, so `i18n:translate` only re-translates keys that are new or whose source changed, and prunes removed ones. No third-party TMS needed, and as you say the LLM output quality makes manual poedit passes largely unnecessary now. Happy to swap in a TMS later if we outgrow it.

---

kfarr: Real human reply here -- @vincentfretin yes I purposefully was prompting for a solution that is local only to not use a third-party service (except claude :/)

kfarr:
@vincentfretin thanks for all your feedback here. I was surprised that pt-BR was the #1 locale for 3dstreet users after English, very closely followed by ES, combined about 10% of our traffic as far as I can tell.

vincentfretin:
You probably want to store the language preference in the user profile in firebase, not in localStorage, that way you can send localized emails as well.

vincentfretin:
If you show any price anywhere, be sure to use the proper format function
https://formatjs.github.io/docs/polyfills/intl-numberformat
$9.99 -> 9,99 $ in French
and any long number formatting:
1,200.50 -> 1 200,50 in French

vincentfretin:
for pricing, you need to translate /mo and /year as well, the Intl api should have that

src/editor/components/elements/AssetInfoPanel.jsx
```diff
@@ -147,7 +150,10 @@ const AssetInfoPanel = ({ entity }) => {
             fontSize: 11
           }}
         >
-          Upgrade for 5 GB storage
+          <FormattedMessage
+            id="assetInfo.upgradeStorage"
+            defaultMessage="Upgrade for 5 GB storage"
```
vincentfretin: You should use a variable for the "5 GB" here so you don't need to update all translations if you change the storage size later.

src/editor/components/elements/GeoSidebar.jsx
```diff
@@ -620,14 +747,21 @@ const GeoSidebar = ({ entity }) => {
                               verticalAlign: 'middle'
                             }}
                           />
-                          {tokenProfile.geoToken} free
+                          <FormattedMessage
+                            id="geoSidebar.tokensFree"
+                            defaultMessage="{geoToken} free"
+                            values={{ geoToken: tokenProfile.geoToken }}
+                          />
```
vincentfretin, line 754: you need to use plural rules here, in French
0 gratuit
1 gratuit
2 gratuits

vincentfretin (commented):
French translation is great!

src/editor/components/elements/AssetInfoPanel.jsx
kfarr: Done (59a0feb). The size is now a constant `PRO_STORAGE_GB = 5`, formatted via `Intl` unit and interpolated as `{storage}` — so the message is `"Upgrade for {storage} storage"` and changing the allowance never touches a translation. Bonus: `Intl`'s `gigabyte` unit also localizes the unit itself — en "5 GB", fr "5 Go".

---

src/editor/components/elements/GeoSidebar.jsx
kfarr: Done (59a0feb) — now an ICU plural: `"{geoToken, plural, one {# free} other {# free}}"`, with the translations agreeing by number. Verified rendering per locale:

```
fr     0 gratuit | 1 gratuit | 2 gratuits
es     0 gratis  | 1 gratis  | 2 gratis
pt-BR  0 grátis  | 1 grátis  | 2 grátis
```

---

kfarr:
@claude please do a code review

vincentfretin:
just saw in last commit "billed yearly, " that wasn't internationalized

vincentfretin:
found some others #1772

PR #1755, opened 2026-06-29
Runtime batching of repeated gltf-model and gltf-part entities

vincentfretin:
## Summary

Automatic runtime batching of repeated `gltf-model` and `gltf-part` entities into `THREE.BatchedMesh` draw calls. After the scene's DOM settles, `batchModels` groups batchable entities by source, builds one `BatchedMesh` per material for each group of ≥2 compatible members, and strips the per-instance meshes — collapsing many identical props/vehicles/humans into a single draw call.

## What's included

- **`batch-models.js`** — groups `[gltf-model]` / `[gltf-part]` under the street container, defers duplicate `gltf-model` loads, and builds a `BatchedMesh` per material. Groups that can't batch (skinned/morph/multi-material/single instance) load individually. A scene-level listener keeps batched slots in sync with transform/visibility changes and detaches, and `popMember` un-batches an entity if an incompatible component is added at runtime.
- **`gltf-model.js`** — a `gltf-model` component (replacing the stock one) that holds its load until grouping is decided, parses each src once and clones per instance (`SkeletonUtils` for skinned meshes), and shares decoded texture sources across instances.
- **`gltf-part.js`** — folded into batching (keyed by `src` + `part`); also dropped the obsolete `buffer` schema prop and the dead `THREE.Geometry` branch (now clones the geometry per instance) and emits `model-loaded` so batching can wait on it.
- **`sharedTextureSources.js`** — refcounted registry sharing one decoded `THREE.Source` (single GPU upload) across textures with the same image hash.
- **`disposeUtils.js`** — `disposeNode` helper honoring batch keep-alive flags.

A per-entity provider adapter (`key` / `ownsResources` / `strip` / `reload`) localizes the differences between `gltf-model` and `gltf-part` so the two paths share one batching core.

## Manual test

Import https://streetmix.net/kfarr/3/example-street and duplicate the street three times to have four instances.

With `sceneElement._batchingEnabled = false;` in `createEntities`:

```
console.table(AFRAME.scenes[0].renderer.info.memory); console.table(AFRAME.scenes[0].renderer.info.render);
{geometries: 929, textures: 217}
{calls: 1149, triangles: 1180522}
```

With `sceneElement._batchingEnabled = true;` in `createEntities`:

```
{geometries: 481, textures: 263}
{calls: 669, triangles: 1180522}
```

Reduces draw calls by 42%.


vincentfretin:
That is a port of my code that took me one or two weeks to iterate on it on and off, with the hubs extension stripped, and adding support for batching glt-part components.
The GLTFSharedTextureSourceExtension that allow sharing texture Source between glb files is included but not used because you don't store imageHash on your glb optimization pipeline, but you don't really have different glb with a similar materials currently anyway.

I see Claude found some things to fix, I'll look into it later.

vincentfretin:
1-6 fixed, 7, 12, 13, 14, 15, 18 needs to be addressed

vincentfretin:
Re 12 (BatchedMesh slot holes):

The premise that holes make GPU memory "slowly grow" isn't quite right for this design. Each group's `BatchedMesh` buffers (`maxInstanceCount` / vertex / index) are allocated once at build in `batchGroup`, sized exactly `members × sub-meshes`. `removeMember` only calls `setVisibleAt(false)`, and we never `addInstance` into a group after build — so a group's GPU footprint is fixed from construction and fully freed by `teardownBuiltGroup` once its last member leaves (`activeMemberCount` → 0). A hole is just an invisible slot inside an already-committed, soon-to-be-freed buffer: bounded by the group's build size, not unbounded. And `popMember` — the only runtime hole source — only fires when an unsafe (non-`BATCH_SAFE_COMPONENTS`) component is added to a batched entity, which is rare in a session.

The genuine growth vector over a long editor session is the opposite one: entities added *after* the batch pass (e.g. clone / layer-reorder, which detach + recreate entities) go through `onLateModelLoaded` and stay unbatched — each its own draw call and GPU upload. So I'm addressing it from that side:

- `removeMember` now uses `deleteInstance` instead of `setVisibleAt(false)`, so freed slots return to `_availableInstanceIds` and are genuinely reusable (hole-free) rather than dead space.
- a per-key tally of unbatched post-load entities, folded back into the group above a threshold via `addInstance` (reusing freed slots, growing capacity with `setInstanceCount` when needed), with a warning if the count climbs without a repack.

r181's `BatchedMesh` supports all of these primitives (`deleteInstance`, `addInstance` slot reuse, `setInstanceCount`, `optimize`).

vincentfretin:
I addressed the main review comments, I can add more tests though.
One remaining issue is removing a street leaves the trees and pedestrians and crash raycast on hover
Capture d’écran du 2026-07-01 18-10-37
probably some conflicts between batch implementation and those street-generated- components, need to investigate. I had a similar issue before in my project.

vincentfretin:
## Benchmark: batching on vs off

Measured the same scene (~524 `gltf-model`/`gltf-part` entities) with batching toggled at runtime (`STREET.batchingEnabled`) — real Chrome, GPU, **vsync uncapped** so FPS reflects actual render cost. 5 load trials each; FPS sampled over 8s via rAF frame-time (avg/median/p5). Script: `TRIALS=5 FPS_SECONDS=8 node scripts/measure-load.mjs`.

| metric | batching off | batching on |
|---|--:|--:|
| navigation → all models loaded (ms) | 2691 | 2579 |
| newScene → all models loaded (ms) | 1228 | 1354 |
| scene meshes | 1140 | 686 |
| batchedMeshes | 0 | 26 |
| unique geometries | 977 | 523 |
| GPU-resident geometries | 929 | 480 |
| **draw calls** (frame) | 1151 | **671** |
| triangles (frame) | 1,180,534 | 1,180,534 |
| textures (GPU) | 217 | 263 |
| **FPS avg** (uncapped) | 61 | **96** |
| FPS median | 62 | 98 |
| FPS p5 (worst) | 48 | 73 |
| frame time (ms) | 16.69 | 10.61 |

**Rendering — the win:** draw calls **−42%** (1151 → 671), meshes and geometries roughly halved → **FPS avg +57% (61 → 96)**, frame time 16.7 → 10.6 ms, and worst-frame (p5) FPS 48 → 73. Triangle count is unchanged — batching merges draw calls, not geometry.

**Load time:** effectively neutral (~2.6–2.7 s to all-models-loaded either way).

**Textures** go 217 → 263, but not color maps: each `BatchedMesh` allocates 2 internal `DataTexture`s (26 matrices + 26 indirect = 52), while actual material textures drop (726 → 472 unique refs). Those 52 data textures are a few KB each — negligible VRAM.

_Single machine; absolute numbers vary, but the direction is consistent across runs._

kfarr:
KF note -- I have changed the below to check boxes and I am checking them off as I invalidate or fix these issues:

# Code review

Overview: this PR replaces the stock `gltf-model` component with a defer-and-clone variant, rewrites `gltf-part`, and adds `batch-models.js`, which folds groups of 2+ identical models into `THREE.BatchedMesh` draw calls after scene load, plus a refcounted shared-texture registry and editor raycaster/viewport integration. The measured win is real (draw calls -42%, FPS +57% on the benchmark scene) and the architecture is solid. The issues below are mostly lifecycle edges of the new resource-sharing machinery.

*Note: these findings are from static code review (tracing the code paths), not from in-browser reproduction. Each was double-checked against the source before listing; line references are to the PR branch.*

## Findings (most severe first)
- [x] **0. test are failing ** - kf added this one
- [x] **1. `src/disposeUtils.js:15` + `src/aframe-components/gltf-model.js:394`: shared-Source refcount over-release, two mechanisms.** (a) Acquire is deduped per Texture instance, but `disposeTextures` releases once per material property. GLTFLoader assigns the *same* Texture instance to `metalnessMap` and `roughnessMap` (and `aoMap` for packed ORM textures), so disposal decrements the refcount 2-3x per single acquire; materials shared across multiple nodes multiply it further (one `disposeTextures` pass per traversed node, no dedup).
(b) `remove()` calls `removeMember()` first, which on last-member teardown clears the `_batchKeepAlive` tags and releases the ref resources, and *then* calls `removeMesh()`, whose `disposeNode` guards no longer fire because the tags were just cleared, releasing the same shared sources a second time.
Either path can drive a shared entry's refcount to 0 while the cached template, sibling clones, or cross-GLB `imageHash` sharers still reference it, closing an ImageBitmap under a live model (black/broken textures on unrelated models). Suggested fixes: dedup releases per dispose pass (a `Set` of seen textures), and swap the order to `removeMesh()` before `removeMember()` or make release idempotent per texture.

- [x] **2. `src/batch-models.js:571`: BatchedMesh never gets `castShadow`/`receiveShadow`.** Both default to `false` on `Object3D`, and the per-instance meshes that had shadow flags are stripped. Nearly every mixin in `assets.js` carries `shadow`, `street-environment`'s directional light casts, and segments receive, so any default street with 2+ clones of the same model loses those shadows the moment batching completes. Fix: copy the flags from the reference mesh onto the BatchedMesh.

- [x] **3. `src/aframe-components/gltf-part.js:80-91`: loader error path never settles.**
The GLTFLoader error handler is a bare `console.error`; the callback is never invoked, so `_loadSettled` stays false, no `model-error` fires, and the cached `LOADING_MODELS[src]` promise never resolves (later `gltf-part` updates for the same src chain onto it forever). A 404'd part stalls `batchModels`' `Promise.all` for the full 30s `LOAD_TIMEOUT_MS` on *every* load of that scene, with all deferred duplicates invisible in the meantime. The early return `if (!this.data.part && this.data.src) return;` at line 22 causes the same stall for a src-only `gltf-part` (still matches `BATCHABLE_SELECTOR`). Fix: pass an error callback that sets `_loadSettled`, emits `model-error`, and evicts the cache entry.

- [x] **4. `src/batch-models.js:648`: re-emitted `model-loaded` with `model: null` bubbles.**
The emit omits the bubbles arg (A-Frame defaults to bubbling) and fires for *all* members, including the reference member that already emitted a real one. `animation-mixer` (`src/lib/animation-mixer.js:40`) has no target or null check: `load(null)` clobbers a working mixer, and `playAction()` then throws reading `model.animations`. Stock characters escape because skinned meshes don't batch, but user-uploaded animated GLBs and ancestor listeners are exposed. Fix: `el.emit('model-loaded', {...}, false)` plus a target guard in animation-mixer.

- [x] **5. `src/batch-models.js:110`: `batch-models-root` leaks into every saved scene JSON.**
The root entity is created without the `autocreated` class, and the serializer's only exclusions are `autocreated` / `data-temporary-file` (`json-utils_1.1.js:96-98`). It is created even for single-model scenes (before the `members.length < 2` check) and round-trips forever; it also leaks into AI-chat context and MCP scene dumps. One-line fix: `el.classList.add('autocreated')` in `getOrCreateBatchRoot`, which also makes the ad-hoc `element.id === 'batch-models-root'` filter in `SceneGraph.jsx:156` unnecessary.

- [x] **6. `src/aframe-components/gltf-part.js:19`: `update()` never releases batch membership (edge case, not traced to a live repro).**
`gltf-model` releases via `loadModel() -> remove() -> removeMember()`; `gltf-part` only releases in `remove()`. A batched entity whose part/src data changes without component removal keeps a ghost instance in the old BatchedMesh, and `onLateModelLoaded` skips re-classification because `_batchStatus` is still set. The common single-mixin swap is protected because `EntityUpdateCommand` clears the mixin first (removing the component), but multi-mixin values skip that guard (`EntityUpdateCommand.js:192`) and direct attribute edits would hit it. Cheap insurance: call `removeMember` at the top of `update()` when data changed.

- [x] **7. `src/editor/index.jsx:234`: the batching gate is armed in core but only opened by the editor (latent, not a live bug today).**
`createEntities` arms `sceneEl._batchingEnabled` and every `gltf-model` parks on `batch-grouping-done`, but the sole `batchModels()` call lives in the editor's `onNewScene`. Today `src/index.js:50` unconditionally requires the editor, so this can't fire, but any future editor-free core bundle would hang all model loads forever, with no timeout on the park. Moving the gate-open (or a fallback timer) into `batch-models.js` itself removes the trap.

- [x] **8. `src/aframe-components/gltf-model.js:272-279`: parked load survives component removal (narrow window).**
`finish()` checks `deferLoad` and `el.isConnected` but not that the component instance is still attached. `removeAttribute('gltf-model')` during the park window (roughly 5 ticks during a batching scene load) leaves `finish()` attaching an orphan mesh no remove path will ever dispose. One-line ownership check (e.g. `this.el.components['gltf-model'] === this`) fixes it.

## Lower-stakes cleanup, worth doing while in here

- **Hot-path efficiency:** the scene-level `componentchanged` listener (`batch-models.js:707`) runs `querySelectorAll` plus a full `updateWorldMatrix(true, true)` subtree refresh on every transform change anywhere in the scene, and `viewport.js:326` repeats that per pointer-move frame during gizmo drags. Iterating the already-tracked group members with `root.contains()` (and snapshotting the list on drag start) bounds it by batched-member count instead of subtree size.
- [x] **Flag sprawl:** `STREET.batchingEnabled` is read at module load (`gltf-model.js:406`), at `createEntities`, and in the editor, so it is only half-toggleable at runtime and depends on import order. One owner (batch-models) would simplify it.
- **`BATCH_SAFE_COMPONENTS` allowlist** (`batch-models.js:67`) silently disables batching for any entity carrying a harmless new component; nobody will notice the perf regression when that happens. Consider a block-list of known-unsafe components instead, or a loud log.
- **Two parallel GLB parse caches** now exist (`gltf-model`'s `loadParsedGltf` and `gltf-part`'s `MODELS`/`LOADING_MODELS`): a src referenced both ways is downloaded and parsed twice, and cache fixes must land in both.
- **Naming per CLAUDE.md:** `src/disposeUtils.js` and `src/sharedTextureSources.js` break the stated kebab-case convention for `/src` (`dispose-utils.js`, `shared-texture-sources.js`), and CLAUDE.md should get a paragraph on the batching subsystem.
- **Hover re-enter staleness** (`raycaster.js:78`): the new hover diff never resets `lastHoveredEl` on canvas mouseleave, so re-entering over the same entity shows no hover box. Pre-PR behavior was identical (parity, not a regression), but since this PR now owns that lifecycle it is a one-line reset.

## Manual test plan

1. **Baseline batching win:** load `#https://streetmix.net/kfarr/3/example-street`, duplicate the street 3x, then `console.table(AFRAME.scenes[0].renderer.info.render)`. Expect ~669 calls batched vs ~1149 unbatched. For A/B, flip `STREET.batchingEnabled` at `json-utils_1.1.js:13` and reload, or run `TRIALS=5 FPS_SECONDS=8 BATCHING=off|on node scripts/measure-load.mjs`.
2. **Editor interactions on batched entities:** hover (box + HUD label), click-select, gizmo drag (watch for jank dragging a whole street layer), visibility toggle, undo/redo, delete one duplicate, duplicate post-load (late repack), and swap a model via the properties dropdown (watch for a ghost of the old model, finding 6).
3. **Save/load round trip:** save, reload via `#/scenes/<uuid>`, confirm identical rendering; inspect the saved JSON for a spurious `batch-models-root` entity (finding 5).
4. **Shadows:** default street, 2+ identical trees on a sidewalk, directional light on. Watch the ground at load: shadows appear then vanish when batching completes (finding 2).
5. **Broken asset:** point a repeated model's src at a 404 and load: duplicates currently stay invisible ~30s (finding 3).
6. **Heavy real-world scene:** load a known model-heavy user scene and compare `renderer.info.render.calls` + FPS (Chrome DevTools > Rendering > Frame Rendering Stats) with batching on vs off; check `[batch-models]` console logs for how many groups classify as "blocking" to gauge `BATCH_SAFE_COMPONENTS` coverage.

kfarr:
Here is the result of my manual testing:
- I was able to load and get a sustained ~9fps on a complicated scene using this branch. The same scene on main eventually degrades to 0fps and then out of memory error kills the webgl context. That test file is attached [1]. This is a big win -- the customer could no longer user their scene before this branch
- although claude says that shadows don't work in point # 2, I saw shadows render as expected on this branch [2]

[1] test file of super big scene with many clones [East Ave Latest.3dstreet (1).json.zip](https://github.com/user-attachments/files/29620264/East.Ave.Latest.3dstreet.1.json.zip)
[2] image:
image

vincentfretin:
Re point 4 (re-emitted `model-loaded` with `model: null`): I don't think this path is reachable, so I'm leaving it as-is.

- The `model: null` re-emit only fires on **batched** entities.
- `animation-mixer` isn't in the `BATCH_SAFE_COMPONENTS` allowlist, so it's a blocking component — any entity carrying it is excluded from batching and never emits that null itself.
- The only way the bubbled null could reach a mixer's `load()` is an ancestor carrying `animation-mixer` that has a *batched descendant*. In practice `animation-mixer` lives on standalone leaf character models, which are themselves blocked from batching and have no batched `gltf-model`/`gltf-part` children — so `load(null)` is never called.

I'm also keeping the re-emit bubbling on purpose: A-Frame's own `gltf-model` emits `model-loaded` with `bubbles=true`, so matching that keeps ancestor listeners consistent.

Happy to add a defensive null guard in `animation-mixer` if we ever expect user scenes to nest a batched model under an animated ancestor, but that configuration doesn't occur today.

vincentfretin:
Now batching stencils and skinned meshes without animation-mixer.
With your big scene
```
metric                      batching=off   batching=on
------------------------------------------------------
navigation → models (ms)           18174          5690
newScene → models (ms)             16936          4475
scene meshes                        8414          2237
batchedMeshes                          0            56
unique geometries                   8080          1903
GPU-resident geometries             5887          1459
draw calls (frame)                  6609           586
triangles (frame)               15902727      15902727
textures                            2819           192
FPS avg (uncapped)                     3            32
FPS median                             3            34
FPS p5 (worst)                         2            21
frame time (ms)                   365.25         33.08
```

PR #1761, opened 2026-06-30
Skip redundant street-generated regeneration on segment-changed (fix #1759)

### LINKED ISSUE #1759: Street-generated clones created twice on initial scene load (street-segment initializes twice on some segments)
## Summary

On the initial load of a saved scene, some street segments fire `segment-changed` **twice**, which makes their `street-generated-*` components (`-clones`, `-stencil`, `-striping`, `-rail`, `-pedestrians`) `clearEntities()` and regenerate. The net effect is that the generated child entities (vehicles, trees, lamps, stencils, pedestrians…) are **created, immediately destroyed, and re-created** during load — wasted work, and a brief flash of geometry that's torn down.

This was surfaced while adding runtime GLB batching (#1755): the per-src "how many entities will load this GLB" tally came out at roughly **2× the number of live entities** for the affected sources.

## Evidence

Loading a saved scene based on `streetmix.net/kfarr/3/example-street` and counting `segment-changed` per segment during load:

```
Dense Sidewalk:        1
Tree Planting Strip:   2   ← regenerates
Modern Street Lamp:    2   ← regenerates
Inbound Parking:       1
Inbound Drive Lane:    1
Outbound Drive Lane:   1
Outbound Parking:      1
Normal Sidewalk:       1
```

Only some segments emit twice. Capturing the `segment-changed` **payloads** shows the two emits on those segments are **both first-inits**, not a value change:

```
Tree Planting Strip: { widthChanged:true, lengthChanged:true, oldW:undefined, oldL:undefined, newW:0.914, newL:100 }
Tree Planting Strip: { widthChanged:true, lengthChanged:true, oldW:undefined, oldL:undefined, newW:0.914, newL:100 }
(every other segment: exactly one such emit)
```

`oldWidth`/`oldLength` are `undefined` **both** times, with identical `newW`/`newL` — i.e. `street-segment.update()` ran twice each with `oldData = {}`. So the `street-segment` component is being **initialized twice** on those specific segments (the component, or the segment entity, is re-created), which re-inits their `street-generated-*` siblings and regenerates the clones. It is *not* a later length change. Why only those two segments get a second init is unconfirmed — that's the thing to chase.

## Root cause (partially identified)

The affected segments have their `street-segment` component **initialized twice** (both `segment-changed` emits carry `oldData = {}`, see above). Each init re-runs the segment's `street-generated-*` siblings, which `clearEntities()` + regenerate — so the clones are created, torn down, recreated.

What's not yet pinned: *why* only those segments get a second `street-segment` init. Candidates to check — `managed-street.refreshManagedEntities()` (it `appendChild`s segment entities and `setAttribute('street-segment', …)`; if it re-creates/re-appends an existing segment the component re-inits) and the segment-regeneration branches in `street-segment.update()` (`type`/`variant`/`side` changes call `this.remove()` + `generateComponentsFromSegmentObject()`).

(An earlier theory that `managed-street.update()`'s initial `length` diff → `applyLength()` re-set the length and caused a second `segment-changed` does **not** fit the payloads — that would show `oldLength` defined on the second emit, but it's `undefined`. So the second emit is a fresh init, not a length change.)

Relevant code:
- `src/aframe-components/managed-street.js` — `update()` (`length` diff → `applyLength()`), `applyLength()` (`setAttribute('street-segment', 'length', …)`).
- `src/aframe-components/street-segment.js` — `update()` emits `segment-changed` when `width`/`length` changed.
- `src/aframe-components/street-generated-clones.js` (and `-stencil`/`-striping`/`-rail`/`-pedestrians`) — `init()` listens for `segment-changed`; the handler does `clearEntities()` then regenerates.

## Impact

- Wasted entity creation + GLB/component init then teardown for the discarded first generation (CPU, GC churn, a load-time flash).
- Inflated the runtime-batching per-src load tally (~2×). It's benign there (over-counting only leans the decision toward cloning), but it's a clear signal of the redundant work.

## How this was noticed (cleanup is fine — verified)

This surfaced because the runtime-batching per-src tally counts at `update()`-time (it deliberately does **not** decrement on `remove()`), so the double creation showed up as ≈2× the live count for the affected sources.

An earlier hypothesis that the discarded first-generation clones don't run their component `remove()` was **checked and is false**. Per the DOM spec, removing a subtree enqueues `disconnectedCallback` for the removed node **and all descendant custom elements**, and A-Frame's `disconnectedCallback` calls each component's `remove()`. Verified in-app: detaching only a grandparent `a-entity` calls `remove()` on the grandparent, parent, **and** grandchild components (`['c','gp','p']`), and a plain custom-element grandchild's `disconnectedCallback` fires too. So per-entity cleanup (e.g. `gltf-model` disposing its mesh) runs correctly on the regeneration teardown.

The only thing A-Frame skips on detach is `component.destroy()` (which returns pooled objects for reuse) — not a leak; those objects are garbage-collected, you just miss pool reuse on the next init.

So this is purely about the **wasted create-then-teardown work** (and the brief flash), not a cleanup/leak bug.

## Suggested directions (for later)

First **pin why those segments' `street-segment` initializes twice** (the candidates above). The actual fix likely falls out of that — e.g. avoid re-creating/re-appending segment entities that are already mounted, or avoid the redundant component re-init.

Failing that, a localized mitigation: **coalesce `segment-changed` → regenerate** in the five `street-generated-*` components (drop a redundant regen within the same frame). Small, but it makes generation async and affects every edit, not just load — and generators currently regenerate synchronously, which consumers may rely on; the segment lifecycle is order-sensitive (see the README), so this wants its own change + testing rather than riding along in an unrelated PR.

Fixing the double creation removes the wasted entity create/teardown and the brief load-time flash.

vincentfretin:
## What

The five `street-generated-*` components (`-clones`, `-stencil`, `-striping`, `-rail`, `-pedestrians`) regenerate their entities in **two** places: their own component `update()`, and the segment's `segment-changed` event (`onSegmentChanged → update()`).

During scene load, a segment's first `street-segment.update()` emits `segment-changed` carrying the **same** width/length the generator already used on its own init `update()`. The handler then `clearEntities()` + regenerates an **identical** set — pure create-then-teardown waste.

## Investigation (issue #1759)

The original report's headline — "`street-segment` initializes twice on some segments" — turned out to be a **measurement artifact**. The example street is symmetric and Streetmix gives both sides the same `data-layer-name`, so counting `segment-changed` by name double-counts the left/right Tree Planting Strip / Modern Street Lamp segments. Verified live: 10 segments, 10 inits (one each), `street-segment.remove()` never called — no segment re-init, no segment regeneration. Details in [this comment](https://github.com/3DStreet/3dstreet/issues/1759#issuecomment-4842895909).

The real (smaller) issue is the redundant generator regeneration above.

## Fix

`segment-changed` only ever fires for width/length changes, so guard the handler: bail when the dimensions the generator depends on are unchanged since its last run — length+width for clones/pedestrians/striping, length only for stencil/rail. It reuses the `this.length`/`this.width` that `update()` already stores. Legitimate dimension edits still regenerate (the component's own `update()` path is untouched).

## Verification

Instrumented a saved scene load in the browser:

| | before | after |
|---|---|---|
| redundant teardowns | 5 | **0** |
| clone creations | 61 | **38** |
| final generated entities | 112 | **112** (identical) |

And changing a segment's length still regenerates correctly (fixed-mode lamp count: 100→3, 40→1, 200→6, back→3).


vincentfretin:
I added aframe components unit testing as part of this PR, I had to convert some commonjs files to ESM for this.
The required test dependencies are in devDependencies, playwright doesn't auto download chromium with a postinstall script, it's not like puppeteer.
`npm run test:components:setup`  step fetches the ~114 MB chromium-headless-shell (full chromium in comparison is ~290MB)

kfarr:
@vincentfretin thanks for the fix and for the playwright example; i ported the managed street test over to playwright to match and reduce extra dependencies

Issue #1794, opened 2026-07-05
GLB/glTF export crashing on prod ("reading 'x'") — likely #1755 BatchedMesh regression + error-toast masking it


kfarr:
## Summary

GLB/glTF export is throwing on production for real users on `3dstreet.app`. There are **two stacked bugs** — a critical export regression and a secondary error-toast rendering bug that masks it.

**Sentry:** [Issue 7592680261](https://3dstreet.sentry.io/issues/7592680261/) — 9 events, all on `https://3dstreet.app/`, release `2026.6.0+2b653c58`, first seen 2026-07-03.

```
Error: Objects are not valid as a React child
(found: TypeError: Cannot read properties of undefined (reading 'x')).
```
- `handled: no`, mechanism `onunhandledrejection`
- Stack: `AppMenu.jsx:270` → `AFRAME.INSPECTOR.exporters.gltf.parse.binary`

## Bug 1 — 🔴 CRITICAL: GLB export fails with `Cannot read properties of undefined (reading 'x')`

The real underlying error is a `TypeError: Cannot read properties of undefined (reading 'x')` thrown inside the glTF **binary** exporter (`exporters.gltf.parse`, `{ binary: true }`).

**Strong suspicion: regression from #1755** ("Runtime batching of repeated gltf-model and gltf-part entities"), which is the commit in the affected release (`2b653c58`) and the newest thing on prod. That change introduces `THREE.BatchedMesh` draw calls. Hypothesis: `GLTFExporter` does not handle `BatchedMesh` (or the shared/instanced geometry the batching path produces), and dies reading `.x` off an undefined position/attribute while serializing.

**Impact:** GLB/glTF export is a paid, watermark-free feature (Pro value). If scenes containing repeated `gltf-model` / `gltf-part` entities (i.e. most real scenes — props, humans, vehicles via mixins) now fail to export, this is a direct hit to paying-customer value. Treat as urgent.

**Likely fix direction:** un-batch / use the original (pre-batch) meshes for export, or skip `BatchedMesh` nodes in the exporter, or temporarily disable batching during the export pass. Needs a repro with a scene that has ≥2 of the same gltf-model.

## Bug 2 — 🟡 Easy win: error toast crashes instead of showing the real error

`src/editor/components/scenegraph/AppMenu.jsx` lines **276** and **296** pass the raw `error` object into `intl.formatMessage`:

```js
STREET.notify.errorMessage(
  intl.formatMessage(
[...]

PR #1796, opened 2026-07-05
Fix GLB/glTF export crash and BatchedMesh export handling (#1794 bug 1)

kfarr:
The prod export crash ("Cannot read properties of undefined (reading 'x')") is a three-copies mismatch, not the BatchedMesh merge itself: the editor's GLTFExporter comes from the npm three package (r184), whose exporter branches on `object.pivot !== null` into a pivot-container path that immediately reads `pivot.x`. Scene objects are built by the super-three bundled in the A-Frame script, and builds predating the Object3D.pivot feature leave that property undefined — so every node takes the pivot branch and export dies on the first one. prepareSceneForGltfExport now normalizes the missing property to null before parsing (test reproduces the exact crash against the real exporter).

Batched scenes were additionally unexportable even without the crash: GLTFExporter has no BatchedMesh support and serializes the merged capacity buffer as one untransformed blob. expandBatchedMeshesForExport re-expands every batched member into ordinary meshes for the duration of the export — one temporary mesh per batch slot, built from the slot's source geometry and the BatchedMesh material with the slot's entity-local matrix, and kept invisible to the renderer/raycasters via layers.mask = 0 (GLTFExporter ignores layers). BatchedMeshes are hidden so the exporter's onlyVisible pass skips them, and an idempotent restore() runs from both exporter callbacks.

Wired into both export paths (AppMenu scene/AR export, CommonComponents single-entity export); the AppMenu error callback now also restores the helper/rigged-entity visibility filters so a failed export no longer leaves the scene hidden.

Claude-Session: https://claude.ai/code/session_01LDoHMa17c1Py9uRrknZ8Fj

### DIFF
--- a/src/batch-models.js
+++ b/src/batch-models.js
@@ -1741,6 +1741,55 @@ export function isBatched(el) {
   return !!el?.object3D?.userData?._batchSlots?.length;
 }
 
+// Make a batched scene exportable by GLTFExporter, which has no BatchedMesh support: it
+// serializes one as a plain Mesh — the merged capacity buffer, parked at the batch root,
+// with no per-instance transforms — so batched entities come out as a garbage blob (or
+// vanish once the BatchedMesh is filtered out). Instead, re-expand every batched member
+// into ordinary Meshes for the duration of the export:
+// - Each member object3D found under `root` gets one temporary Mesh per batch slot, built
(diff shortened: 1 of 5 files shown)


kfarr:
@claude code review pls

PR #1802, opened 2026-07-05
Fix GLB export post-processing failing with EXT_texture_webp on webp-textured scenes

vincentfretin:
## What

GLB export post-processing (`transformUVs` in `ScreenshotModal/gltfTransforms.js`) failed with:

```
Error: Missing required extension, "EXT_texture_webp".
```

on any scene containing assets with webp textures (most 3DStreet assets).

## Why

three r184's GLTFExporter keeps textures loaded from webp images as webp — GLTFLoader stamps `texture.userData.mimeType`, and the exporter then declares `EXT_texture_webp` in `extensionsRequired`. glTF-Transform's `readBinary` throws for any *required* extension not registered on the IO, and `transformUVs` only registered `KHRTextureTransform`. Fix: register `ALL_EXTENSIONS`, as `addGLBMetadata` in the same file already does.

Also inlines the `export` statements with the function declarations in `gltfTransforms.js`.

## Testing

New browser-mode regression test (`test/components/gltf-transforms.test.js`) exports a scene with a webp-stamped texture, asserts the GLB declares `EXT_texture_webp`, then runs `transformUVs` + `addGLBMetadata` on it. Verified the test fails with exactly the production error when the fix is reverted. `npm run test:components` (34 passed) and `npm run lint` are clean.

### DIFF
--- a/src/editor/components/modals/ScreenshotModal/gltfTransforms.js
+++ b/src/editor/components/modals/ScreenshotModal/gltfTransforms.js
@@ -1,15 +1,13 @@
 import { NodeIO } from '@gltf-transform/core';
-import {
-  KHRXMP,
-  ALL_EXTENSIONS,
-  KHRTextureTransform
-} from '@gltf-transform/extensions';
+import { KHRXMP, ALL_EXTENSIONS } from '@gltf-transform/extensions';
 import { vec3, mat3 } from 'gl-matrix';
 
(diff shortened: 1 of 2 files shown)


vincentfretin:
Integrated into #1801 (cherry-picked as part of the A-Frame 1.8.0 upgrade PR).

PR #1809, opened 2026-07-06
Add .managed-street.json export and round-trip import support

kfarr:
## Summary

Adds the ability to export a managed street's live DOM state as a Format-2 JSON file (`.managed-street.json`) and re-import it losslessly. This enables users to download street configurations and share them, while ensuring that properties previously lost during import (seed, cycleOffset, facing, positionX, stencilHeight, direction) are now preserved across round trips.

## Key Changes

- **New export function `getManagedStreetJSON()`** in `managed-street.js`: Walks the live DOM and rebuilds a Format-2 segment list, capturing all generated component properties and handling auto-striping logic to prevent unwanted recreation on re-import.

- **Enhanced import round-trip support** in `street-segment.js`: 
  - Added `buildGeneratedAttributes()` helper to pass through all schema-known properties from generated entries, preserving previously-dropped fields (seed, cycleOffset, facing, positionX/Y, stencilHeight, direction).
  - Handles both string and array forms of `modelsArray` for compatibility with re-imported data.

- **Striping auto-generation logic extracted** to module scope in `managed-street.js`: Moved `getStripingFromSegments()` out of the component so the export can reuse it to decide when to pin an explicit empty striping list (preventing auto-striping from recreating deleted stripes on re-import).

- **UI integration** in `ManagedStreetSidebar.jsx`: Added "Download Street JSON" button that serializes the live street state and triggers a browser download with PostHog analytics.

- **Store-free module structure**: Split entity-category query helpers into new `street-entity-utils.js` and made Streetmix parser dependencies lazy-loaded in `managed-street.js`. This allows the json-blob import/export path to work without the Zustand store's Firebase/PostHog dependency chain, enabling browser-mode component tests to run cleanly.

- **Module conversion**: Converted several CommonJS requires to ES6 imports in `street-align.js`, `street-ground.js`, `street-label.js`, and `street-layout-utils.js` for consistency.

- **Comprehensive test suite** (`managed-street.test.js`): 264-line test file covering:
  - Export captures all live state including previously-dropped properties
  - Round-trip losslessness (export → re-import → export yields identical object)
  - Deleted striping does not grow back across a round trip

- **i18n**: Added translation key for the new download button.

## Notable Implementation Details

- The export serializes `modelsArray` back to the comma-joined authoring convention used in json-blobs and presets, even though A-Frame internally stores it as an array.
- Auto-striping between adjacent lanes is now deterministic and reusable: the same logic applies during import and export, ensuring consistency.
- The round-trip contract is formally documented in the test file and enforced by the test suite.
- Lazy loading of Streetmix parser dependencies keeps the json-blob path (and its tests) free of transitive Firebase/PostHog imports, improving test isolation and load times.

https://claude.ai/code/session_01FWCzy5WJGZQt4ofRUpTBDz


kfarr:
Thanks for the review — pushed `dc5dccc` addressing 1–5:

1. **Filename prefix** — the download name now derives from the exported JSON's `name` (already prefix-stripped by `getManagedStreetJSON`), so the on-disk file matches the street name instead of carrying `Managed Street • `.
2. **Download helper** — switched to the shared `saveString` from `utils.js` (append-to-body + revoke handled there) rather than a hand-rolled anchor.
3. **Analytics ordering** — `export_initiated` now fires after the download is triggered.
4. **Test sleep** — replaced the 50 ms delay with a positive `vi.waitFor`: every fixture segment carries generated content, so the wait resolves once each segment has a `street-generated-*` component attached (i.e. its `loaded` handler has run). Green across repeated runs locally.
5. **Parallel imports** — `loadStreetmixParserDeps` now uses `Promise.all`.

**6. Clone `direction` no longer force-overridden** — confirmed intentional. Previously generated content had `direction: this.data.direction` unconditionally; the passthrough now honors an entry's explicit `direction` (e.g. `'none'` for angle parking / side-oriented objects) and only falls back to the segment direction when unset. We're treating this new rendering scheme as the source of truth going forward rather than preserving the legacy forced-direction behavior, so no change here.

The two nits (mid-upload `blob:` capture, and the silent interop fallback) I've left as-is for now — happy to follow up if you'd like either hardened.

---

Issue #1289, opened 2025-07-27
Show public username when viewing another user's scene


kfarr:
As a user viewing another user's scene, I'd like to see the scene author's username in the header near the title.

Future: this has a hover state with information about the user. Clicking will lead to a user's profile page.

Issue #1288, opened 2025-07-27
Prompt for public username confirmation for new users


kfarr:
As a new user to 3DStreet I need to know that my creations are public and remixable, and I would like to be presented an opportunity to choose a public-facing username. This is likely provided in the form of a new user modal when logging in. The definition of a new user is simply whether or not they have an existing socialProfile object in the database.

A default public username will be shown in the modal and the user is given the opportunity to change it.

Currently, default public usernames are generated when a user first visits their profile page. Instead, this should be generated when the new user modal is displayed.

PR #1812, opened 2026-07-06
Unified Viewer + play lifecycle, with driving sim and traffic-sensor replay

kfarr:
## What this is

One core concept — **a single non-editing Viewer presentation, a single Start/Pause simulation lifecycle, and content-derived capabilities that plug into it** — plus the first two capabilities living on it: the **driving sim** (ported from #1627) and the **traffic-sensor replay** (ported from `claude/magical-faraday-on70it`). Design rationale + implementation status in [`docs/unified-viewer-play-modes.md`](https://github.com/3DStreet/3dstreet/blob/claude/3dstreet-view-play-drive-cly4qp/docs/unified-viewer-play-modes.md).

## The model

| Piece | What it owns |
| --- | --- |
| `mode-manager` (system) | Scene control modes via registered enter/exit hooks (`editor`, `locomotion`, `drive`) + the **playable-capability registry** that gates all Play UI |
| `play-mode` (system) | `start/stop/pause/resume/reset`, `play-mode-*` scene events, race/collision outcomes, gamepad; advances the deterministic `scene-timer.simulationTime` (physics takes ownership when active) |
| Viewer top bar | Title · author byline (#1289) · VIEW ONLY chip · sim clock pill · Reset/Stop · **Start** (rename, see below) · Edit (author) / Remix (non-author, #1325) · Sign In / profile avatar |
| Entry rules | Editor Start button / `P` · non-authors on `#/scenes/UUID` land in the Viewer · `?viewer=true` · saved vantage = default snapshot → `memory.cameraState` → `?camera=` |

Capabilities are subscribers: `drive-mode` switches to its registered mode on `play-mode-start` when a `[drive-controls]` entity exists; `street-traffic` animates `managed-street[playable]` lanes; `street-traffic-replay` replays a sensor manifest. Static scenes (everything today) never see play controls.

## Decisions taken in this pass

- **Button is "Start", not "Play".** "Play" only reads well in English and localized into the media-player sense (es *Reproducir*, pt-BR *Reproduzir*, fr *Lecture*), which misdescribes the action. "Start" pairs with the Stop/Reset/Pause shuttle, reads for a lay (non-gamer) audience, and translates cleanly (Iniciar / Démarrer). Internal system naming (`play-mode`, `isPlaying`) is unchanged. A future per-capability label ("Drive") can specialize it.
- **Non-owners get Start too.** Previously the shuttle was gated to owners, leaving a visitor on a shared playable scene with a frozen render and no controls. Playing mutates nothing persistent (and the `P` shortcut already gated on `hasPlayable()` alone), so the shuttle now shows to any viewer of a playable scene; drive still requires an explicit action. **Deferred:** per-capability *ambient auto-play* (a traffic/replay link starting to move on entry, no click) — needs an `ambient` flag on the registry; noted as a Phase-3 follow-up.

## Code-review fixes (high-effort review of the branch diff)

- Sim clock rendered an illegal **`1:60.00`** at every minute boundary — extracted one shared `format-sim-time` util that rounds to centiseconds first, replacing 4 copy-pasted formatters. Regression test added.
- `street-traffic-replay` never handled `play-mode-reset` → pressing Reset mid-replay froze it (negative `tRel`) for as long as the prior run lasted. Now re-arms to the top of the manifest.
- Gamepad reset/camera/driving inputs fired **while paused** despite the comment; `pollGamepad` now has a pausedOnly mode.
- Chase-cam wheel listener `preventDefault`ed every wheel event document-wide (froze page/UI scroll while driving); now scoped to the scene canvas.
- `createEntityFromObj` could append a duplicate `<a-entity id="cameraRig">`, breaking the id lookups mode-manager uses; now reuses the existing rig.
- `useHasPlayable` ran a full-scene `querySelectorAll` sweep per mutation during load; coalesced into one rAF-deferred recheck.
- `bestTimes.js` → `best-times.js` (kebab-case convention).

## Ship checklist

- [x] **Legacy-scene regression** — 2 real dev scenes loaded in a headless browser: exactly one `#cameraRig` (dedup fix holds), viewer-first entry for signed-out visitors, locomotion/WASD works, legacy control stripping intact.
- [x] **Editor regressions** — panels toggle, managed-street sidebar (Reload/Download from main + the new "Animate in Play" playable toggle), inline title rename, undo/redo after a viewer round-trip — all verified in-browser.
- [x] **i18n** — the ported drive/replay panels (PlayModeControls, DriveControlsSidebar→`.jsx`, RaceFinishBanner, StreetTrafficReplaySidebar) converted to react-intl (43 keys, ICU plurals); es/pt-BR/fr + manifest hand-filled; `i18n:check` clean at 679 keys.
- [x] **Bundle / perf budget** — `rapier` is a **separate 2.0 MiB code-split async chunk** (lazy-loaded on first Start), excluded from the perf budget. The hard budget (3,984,589 B, `hints: 'error'`) was **not** raised by this PR; the main entry bundle passes at 3,909,351 B — **~73 KiB (1.9%) headroom** (thin; worth watching for future main-bundle additions).
- [x] **Non-owner-on-playable decision** — implemented (Start for visitors); ambient auto-play explicitly deferred (above).
- [x] **Merged latest `main`** — resolved (SceneEditTitle `canEdit = editable && !readOnly`; i18n manifest union); full suite + i18n check green.
- [x] **Docs + PR** — design-doc status section + naming resolution updated; this body reflects final state.

## Verified

- Suites: `lint`, `prettier`, `i18n:check` (679 keys), 88 (mocha) + 372 (vitest) + 39 (component) tests, plus a new `format-sim-time` regression test. `npm run dist` passes the hard perf budget.
- Real-browser (headless Chromium, real dev scenes): Start rename, viewer round-trip + WASD locomotion, legacy-scene load with single cameraRig, non-owner Start shuttle on a drive-controls scene, managed-street sidebar, rename + undo/redo.

## Known gaps / notes

- Vehicle meshes and character models load from the assets CDN; heavy scenes starve GLB loading under swiftshader, so hands-on driving physics is best confirmed on staging (the physics/lifecycle paths are what's verified here).
- Byline requires sign-in (Firestore `socialProfile` rules); anonymous visitors see title + VIEW ONLY only (#1288/#1289).
- Per-capability **ambient auto-play** for non-owners is deferred (Phase 3).
- XR/AR viewer presets remain removed (as in #1627); they return later as a registered capability.
- This PR supersedes #1627; the magical-faraday branch is fully absorbed.


kfarr:
# Code review — 10 confirmed findings, ranked by importance

High-effort review of the full branch diff (8 finder passes, every candidate independently verified against the code at f2a3a22). Ranked by risk to existing users first, then play-feature correctness.

## P0 — regressions that can hit existing users (fix before merge)

**1. Small camera drags now change the selection in the plain editor** — `src/editor/lib/raycaster.js:127`
The click-vs-drag guard was relaxed from `=== 0` to `< 0.01` normalized (~10px on a 1000px viewport) because layout shifts between Viewer and editor perturb the container rect, making the same physical click read as a tiny "drag". But the relaxation overshoots: a real ~10px orbit/pan nudge ending over an entity now switches (or clears) the selection. Every editor user hits this path.
**Fix:** compare raw `clientX/clientY` pixel deltas (viewport-absolute, immune to container-rect drift) with a tight ~2-3px threshold, instead of comparing rect-normalized coords loosely.

**2. Escape pressed to dismiss a modal also exits the Viewer** — `src/editor/components/scenegraph/Toolbar.jsx:203`
The viewer's window `keydown` handler only guards INPUT/TEXTAREA focus. Shared modals close on `keyup`, so their `stopPropagation` can't block it. A visitor who opens the sign-in modal and presses Escape gets bounced into the editor (or has a running sim stopped) at the same moment the modal closes.
**Fix:** early-return when a modal is open in the store; respect `e.defaultPrevented`.

**3. Serializer strips look/movement/hand/blink-controls from every entity** — `src/json-utils_1.1.js:538`
`LEGACY_STRIPPED_COMPONENTS` runs on all entity data in `createEntityFromObj` — which is also the recreate path for reparent and paste — so any scene legitimately carrying one of these components (legacy saves, hand-authored JSON hash scheme, AI-chat componentadd) loses it silently on load/reparent/paste and permanently on next save. The legacy junk only ever lived on the saved `cameraRig` entry.
**Fix:** scope the strip to `entityData.id === 'cameraRig'` (that branch already exists ~10 lines below).

## P1 — play-lifecycle bugs (visible in the partner demos)

**4. Stop during cold Rapier load resurrects the physics world** — `src/aframe-components/play/play-mode-vehicle.js:674` and `activate()` at :95
`activate()` sets `active = true` and recreates the World unconditionally after `await loadRapier()`, and the vehicle init has no liveness check after its await. Stop before the 2 MB rapier chunk resolves (first Play of a session) → cleanup runs, then the pending continuation recreates the world and builds a ghost chassis on the removed component. The world then steps every frame while nothing is playing, and the next Start spawns inside the ghost body (phantom collision at spawn). The existing `_activationToken` only guards drive-mode's collider seeding.
**Fix:** generation counter across the awaits in `activate()`/`deactivate()` + `if (!this.el.isConnected) return` before `buildVehicle()`.

**5. Replay + synthetic traffic double-hide leaves static clones permanently hidden after Stop** — `src/aframe-components/play/street-traffic-replay.js:304`
`hideStaticClones` records `wasVisible` from `object3D.visible` *after* street-traffic's own `play-mode-start` listener may have already hidden the same clones (setAttribute updates it synchronously). The second hider records `wasVisible=false`; on Stop the first restores, the second re-hides — static vehicles/pedestrians stay invisible until reload. Reachable with defaults: two replay layers on the same street, or one replay layer with `suppressSyntheticTraffic: false` on a playable street.
**Fix:** single owner (shared helper with refcount) for clone hide/restore.

**6. Collision markers survive Stop as orphaned, undeletable scene-root entities** — `src/aframe-components/play/play-mode.js:139`
Markers append to the a-scene root, but save walks only `#street-container`/`#environment`/`#reference-layers` and the SceneGraph is rooted at `#street-container` — so the comment "json-utils picks it up at save time" is false: after Stop → Edit, red spheres float in the viewport, unselectable, undeletable, silently dropped on save. Only `reset()` removes them.
**Fix:** strip `[collision-marker]` in `stop()` too; fix the comment.

**7. Replay agents can spawn riding the wrong direction** — `src/aframe-components/play/street-traffic-replay.js:340`
`pickLane`'s loop returns unconditionally once a candidate lane type has *any* lanes, so the documented fall-through to the next type never happens when the first type has no direction match. An outbound cyclist on a street with only an inbound bike-lane spawns riding against the lane instead of using an outbound drive-lane — misrepresents the recorded sensor data in exactly the partner-facing replay demos.
**Fix:** `continue` when `byDir` is empty; apply the any-direction fallback only after all types are exhausted.

## P2 — smaller play-mode correctness

**8. Resuming after a race finish leaves the clock pinned and the banner mounted** — `src/aframe-components/play/play-mode.js:196`
`resume()` never clears `playOutcome`/`playFinish` (contradicting its own comment; only start/reset/stop clear them, and 'finish' has no auto-clear timeout like 'crash'). **Fix:** clear them in `resume()`.

**9. Post-Reset crashes near the old crash site are silently swallowed** — `src/aframe-components/play/play-mode-vehicle.js:955`
The collision de-dupe uses a signed sim-time delta and `_lastCollisionAt` survives Reset; after Reset the delta goes negative, which always passes `dtMs < 1000`, so any crash within 1.5 m of the pre-reset site is dropped (no toast/marker/penalty) — unbounded in time. **Fix:** clear `_lastCollisionAt` on reset (and/or `Math.abs`).

**10. Painted lane markings vanish during play** — `src/aframe-components/play/street-traffic.js:254`
`hideFilter = () => true` on drive/bus/bike lanes hides every `[data-parent-component]` child, including `street-generated-stencil` clones (bike symbols, sharrows, turn arrows, BUS/TAXI words). The adjacent TODO acknowledges it; the same filter is copy-pasted into replay's `hideStaticClones`, so fix it in the shared helper both should use.

---

## Below the cut (confirmed, not blocking)

- `?replay=sample` demo bootstrap (`replay-demo.js`, self-described "throwaway") ships to prod unconditionally via `src/index.js`; gate it behind `NODE_ENV !== 'production'`.
- Non-passive `wheel` listener stays on `window` for the whole drive session (the canvas scoping landed as an in-handler check only) — page scroll loses the fast path while driving.
- `addOtherVehicleColliders` re-seeds an overlapping static body on every bubbling `model-loaded`.
- Per-frame waste in hot loops: triple ancestor-chain matrix recompose per traffic record (`street-traffic.js:367`), per-wheel Rapier constant fetches allocating across the WASM boundary each frame, delivery-bot antenna rebuilding its cylinder geometry (+GPU upload) every tick, race-target inverting a static Matrix4 per frame.
- Duplication worth consolidating: hide/restore-clones logic (traffic vs replay, already diverging), batch-aware bbox fallback (3rd copy), byline/author resolution (4th copy), replay sidebar re-implements `summarizeManifest` + file-picker from `createLayerFunctions` (and loses the counts fallback), dead `playStartedAt` state.
- Altitude: gamepad drive input + race scoring live in the capability-agnostic `play-mode` core; traffic↔drive know each other via selector sniffing; `showSimClock = controlMode === 'drive'` string-matches a mode name in the shared toolbar.
- CLAUDE.md not updated with the new mode-manager/play subsystem.

kfarr:
P0s 1-3 from the review above are fixed in f1cbbfd8: raycaster click-vs-drag now compares raw pixel deltas (≤2px), the Viewer Escape handler ignores keydown while a modal is open, and the legacy component strip is scoped to the saved cameraRig entry. Click behavior verified in a real browser (stationary + 1px-jitter clicks select, 15px drags don't). P1/P2 items remain open.

kfarr:
deploy checklist

- [ ] - manual testing of before / after click-vs-drag pixel delta changes; it's unclear why this change was made and if it breaks other things.
- [ ] - Fix P1 and P2
- [ ] - review demo bootstraps such as "?replay=sample" -- are there any others?
- [ ] - for wheel handler, this quote "page scroll loses the fast path while driving." confused me -- what does that mean?
- [ ] - Per-frame waste in hot loops -- that'd be important to fix actually
- [ ] - duplicate "(traffic vs replay, already diverging" -- in what way? what does that mean?
- [ ] - "Altitude..." what does that mean?
- [ ] - CLAUDE.md not updated - yes update

kfarr:
Deploy checklist resolution:

- [x] **click-vs-drag** — A/B tested against main verbatim: the 2-clicks-per-selection bug is caused by this PR's play→editor restoration (mouseup ordering), not latent on prod. Final fix is the order-independent up position + raw client pixels at **0px tolerance** (exact main strictness); the 2px loosening was unnecessary and dropped. Human-verified.
- [x] **P1 + P2 fixed** (3970d9ab): Rapier resurrection race on Stop-during-load; shared refcounted clone hide/restore (`play/clone-visibility.js`) fixing the traffic+replay double-hide and keeping painted stencils visible during play; collision markers stripped on Stop; replay pickLane direction fall-through; resume clears the finish banner; post-reset crash de-dupe.
- [x] **Demo bootstraps reviewed** — `?replay=` was the only ungated one, now dev-only; `#fixture:` was already gated. Prod demo path: saved cloud scene with the Traffic Replay layer (manifest inlines into the scene doc) — round-trip tested.
- [x] **Wheel handler** — non-passive listener moved from window to the scene canvas so page scroll keeps the browser fast path while driving.
- [x] **Per-frame hot loops fixed**: single matrix compose per traffic record; wheel constants cached instead of per-frame WASM fetches; delivery-bot antenna scaled instead of rebuilt; race-target inverse cached.
- [x] **Duplication** — hide/restore consolidated into the shared registry (also the P1 fix).
- [x] **CLAUDE.md updated** with the Play Mode & Viewer subsystem.
- Bonus: static scenes no longer show the confusing "View" button — the toolbar button only renders (as "Start") when a playable capability exists.

Deferred (below-the-cut): addOtherVehicleColliders re-seed on bubbling model-loaded, bbox/byline/summarizeManifest consolidation, dead playStartedAt, altitude refactors.

Issue #1780, opened 2026-07-01
Safari bug -- error using generator to create video from image source in Safari


kfarr:
need to repro

kfarr:
## Investigation summary (Safari video-from-image error)

Reproduced in macOS Safari. The user-facing symptom is a red **`internal`** error toast. Relevant console lines:

```
Failed to load resource: The network connection was lost. (generateReplicateVideo)
Video generation error: FirebaseError: internal
```

### Root cause (hypothesis)
The failure is the **`generateReplicateVideo` callable request being dropped mid-flight by Safari**, not the gallery save we initially suspected.

- `generateReplicateVideo` is a **synchronous, long-running callable**: server-side it does `await replicate.run(model_name, …)` (`public/functions/replicate.js:680`), which blocks until the *entire* video render finishes (often minutes). Client-side the callable is configured with a 9-minute timeout (`src/generator/video.js:593`).
- Safari does not hold a single HTTPS request open that long with no streamed data. It drops the connection with `NSURLErrorNetworkConnectionLost` → `"The network connection was lost"`. The Firebase SDK then surfaces the generic `FirebaseError: internal`, which is what renders as the red `internal` toast.
- Chrome tolerates the long-lived connection, which is why this is Safari-specific.

This is a different code path from the `saveToGallery` cross-origin `fetch()` concern raised earlier — the request dies **before** any video URL comes back, so `displayVideo()`/`saveToGallery()` never run.

### Noise ruled out
The `Blocked a frame with origin "https://3dstreet.app" from accessing a frame with origin "https://www.google.com" / "https://js.stripe.com"` errors are unrelated cross-origin iframe warnings from Google auth + Stripe embeds. They are not part of the video flow.

### ⚠️ Token / data-loss implication (needs log confirmation)
Tokens are deducted **after** `replicate.run` resolves (`replicate.js:687`+). Firebase `onCall` functions keep executing after a client disconnects, so the likely sequence in Safari is:

1. Safari drops the connection → client shows `internal`, no video.
2. Server finishes `replicate.run`, **deducts tokens**, writes `generationLog`/`tokenLog`, returns a `video_url` that no client is listening for.

Net result: **user is charged and a video is produced, but they never see or save it.** Worth confirming against Cloud Functions logs — check whether `generateReplicateVideo` for the affected `userId` logged `Video generation completed` + a `tokenLog` deduction despite the client-side `internal`.

### Recommended direction
[...]

Issue #1605, opened 2026-05-05
Snapshot gallery: focus button to return camera to capture position


kfarr:
## Problem

When iterating on AI renders for a single vantage point (e.g. *looking west on Lombard from ~Van Ness & Greenwich*) we typically need to capture multiple snapshots from the **same exact camera pose** — for example a "before" frame without building envelopes and a "massing" frame with envelopes visible, so the two can be paired in a before/after.

Today there's no way to return the camera to the pose used when a prior snapshot was taken. You have to eyeball it, which makes paired captures imprecise and rerolls of an AI render expensive.

## Proposal

Mirror the existing **focus-on-entity** pattern, but applied to snapshots in the gallery:

- When a snapshot is captured, persist the camera pose (position, rotation, fov) alongside the snapshot record.
- In the snapshot preview modal, add a discrete **focus** button to the right of the title. Clicking it restores the camera to the stored pose.
- For older snapshots without a stored pose, hide or disable the button.

## Why this matters

Unblocks a clean before/massing/after capture workflow used in our visualization pipeline (city scenarios → 3DStreet → AI render → social video). Without it, every variant render forces a manual camera realignment that's slow and rarely pixel-accurate.

PR #1836, opened 2026-07-09
Generation job queue: image kinds go async, fal webhook, email-race fix, generator pending cards (#1832–#1835)

### LINKED ISSUE #1832: Generation job queue: wire up fal webhook for mesh jobs (faster closed-tab completion + email)
## Context

PR #1779 added fal as the queue's first poll-provider (image → 3D mesh via Hunyuan3D/TRELLIS): no webhook, so a closed-tab job is only finalized by the 10-minute reconciler, and the completion email goes out on the sweep after that (3-min open-tab grace). Observed worst case on dev: email lands ~15-20 min after submit. Replicate kinds (splat/video) get real-time finalize + email via their webhook.

fal's queue API supports webhooks; we just did not hook it up for v1 (jobs run 30-90s, so poll+reconciler was acceptable).

## Ask

- Add a `falJobWebhook` (or extend the shared `handleJobWebhook`) and pass a webhook URL on submit in `generateFalMesh` (`public/functions/fal-3d.js`), with the same per-job secret pattern as `replicateJobWebhook`.
- Webhook body untrusted: re-fetch authoritatively via the existing `fetchFalPrediction` adapter, then run the shared `processTerminalPrediction` + `sendGenerationReadyEmail` (both already idempotent; poll + reconciler stay as backstops unchanged).

## Outcome

Closed-tab mesh jobs finalize + email in real time instead of on the reconciler cadence.

Related: open-tab email suppression race (separate issue) applies to this webhook too once it exists.

### LINKED ISSUE #1833: Completion email: open-tab suppression loses the race against webhook-finalized jobs (video/splat, future fal)
## Observed (dev, 2026-07-09, PR #1779 testing)

A video generation completed while its tab was open and polling, yet the completion email still sent (job `577b33f3`, `notify.sentAt` 21:47:46Z). Intended behavior: an open tab acks the result (`ackClientSeen` in `getGenerationJobStatus`) and the email is suppressed.

## Why

For webhook providers (Replicate: splat + video), the webhook finalizes the job and calls `sendGenerationReadyEmail` immediately, before the open tab's next 3s poll can ack. The webhook always wins the race, so open-tab suppression effectively never applies to webhook kinds. It only works for poll-only providers today (fal mesh: the open tab's poll IS the finalizer, confirmed working, job `58d932a3` acked with no email).

Once the fal webhook lands (#1832), fal inherits the same race.

## Possible approaches

- Grace delay before the webhook's email send (e.g. wait 2x client poll interval, then send only if still un-acked); the reconciler notify sweep already implements exactly this pattern (`NOTIFY_GRACE_MS`), so the cheapest fix may be: webhook never emails, sweep always does, at the cost of up to ~10 min email latency.
- Or a short scheduled task / delayed retry from the webhook path keeping real-time-ish sends.

## Severity

Low: worst case is a redundant email for a result the user already saw. But it undermines the "suppressed if tab open" contract stated on all three tabs' checkbox comments.

### LINKED ISSUE #1834: Generator: pending job card not visible in assets sidebar during video/mesh generation
## Observed (dev, 2026-07-09, PR #1779 manual testing)

While video and 3D mesh jobs were in flight, no "Generating…" pending card appeared in the generator app's assets sidebar. The finished assets DID land in the gallery, and the same pending-card mechanism is believed to work in the editor's Assets panel.

## Static review found no smoking gun

- Generator sidebar reuses the shared `AssetsPanelBody` → `AssetsContent`, same as the editor (`src/shared/assets/components/AssetsSidebar.jsx`).
- `useAssets` subscribes to `users/{uid}/generationJobs` where `status in ['queued','running','saving']`; only `splat-rad` is filtered out of the cards.
- `PendingJobCard` has `video` and `mesh` nouns (added in #1779).
- `useAssets` falls back to `window.authState` in the generator, and regular gallery items load fine, so auth/userId wiring appears OK.

## Repro / debug next steps

1. Start a video or 3D Model generation on dev.
2. Expand the assets sidebar (note: it starts COLLAPSED in the generator; if `AssetsPanelBody` only mounts on expand, the listener starts then) while the job is non-terminal.
3. Check the browser console for Firestore `onSnapshot` errors (rules/index) on the generationJobs query.
4. Compare against the editor Assets panel with the same job in flight.

Flagged as follow-up from #1779; not a merge blocker.

### LINKED ISSUE #1835: Migrate image generation to the async job queue (generator Image tab + editor AI render)
## Context

PR #1779 (+#1783) moved video and 3D mesh generation onto the `generationJobs` async queue: submit-and-return, charge-at-submit + refund-once, server-side gallery save, pending card, opt-in completion email, reconciler backstop. **Image generation is the last synchronous kind**: `generateReplicateImage` and `generateFalImage` hold the callable connection open for the whole render and the client saves the result to the gallery itself.

Two client surfaces call these:
1. **Generator Image tab** (`src/generator/generator-tab-base.js`)
2. **Editor AI render** (`src/editor/components/modals/ScreenshotModal/ScreenshotModal.component.jsx`, `source: 'editor'`) — client-side gallery save dies if the modal/tab closes mid-render

## Why

- Same failure class as #1780: slower image models (multi-image, hi-res, LoRA) can outlive Safari's idle connection tolerance; result lost, tokens charged.
- Consistency: images are the only kind without charge-at-submit/refund-once semantics, server-side save, pending card, or completion email. The queue already anticipates it — `PendingJobCard` has an `image` noun and the `generationReady` email template has image copy.

## Sketch

- `kind: 'image'` jobs; Replicate models via the existing webhook+poll+reconciler convergence, fal image models via the fal poll adapter (or webhook once #1832 lands).
- `saveImageToGallery` in the shared terminal processor (mirror `saveVideoToGallery`/`saveMeshToGallery`).
- Clients mirror `video.js`/`model3d.js`: submit → jobId → poll loop; editor keeps its modal UX but survives closure.
- Most image renders are fast (seconds), so completion email default OFF may be the right call for this kind.

Can split backend+generator vs editor surface into separate PRs.

Related: #1832, #1833, #1834.

kfarr:
Closes the four related generation-job-queue follow-ups from #1779.

Closes #1832, closes #1833, closes #1834, closes #1835.

## #1835 — Image generation migrates to the async job queue (the big one)

Images were the last synchronous kind: `generateReplicateImage` / `generateFalImage` held the callable open for the whole render and the **client** saved the result to the gallery — so a closed tab/modal (or a Safari idle-connection drop on slower models) lost the image while tokens stayed charged.

**Backend**
- Both callables are now submit-and-return `kind: 'image'` jobs: pending doc → charge-at-submit + refund-once → provider submit **with webhook** → `{ jobId }`. Timeouts drop 300s → 120s. Token semantics flip from charge-after-success to charge-at-submit + refund-once, matching every other kind.
- `saveImageToGallery` in the shared terminal processor mirrors `saveVideoToGallery`: streamed provider-CDN → Storage at `users/{uid}/assets/images/{assetId}.{ext}` (MIME/extension from the downloaded bytes), deterministic assetId so webhook/poll/reconciler retries converge on one asset, same asset-doc contract the client save wrote (`type: 'image'` / `category: 'ai-render'`, quota trigger unchanged).
- The Discord post moves from the submit callables into the terminal processor's claimed success branch — fires exactly once, on real completion, with the durable Storage URL.
- New sanitized `gallery_metadata` passthrough on the job doc: the editor's `sceneId`/`sceneTitle`/`cameraState`/`renderMode` still land in the saved asset's `generationMetadata`, so the gallery's scene link and focus-camera button (#1605) keep working with the server-side save.
- `fetchFalPrediction` is now kind-aware (`images[0].url` for image jobs, mesh aliases otherwise). Reconciler needs no changes — provider dispatch already covers both.

**Clients**
- Generator Image tab (`generator-tab-base.js`) mirrors `video.js`: submit → `jobId` → self-rescheduling 3s poll; client-side `saveToGallery` retired; "Email me when my image is ready" checkbox added, **default OFF** (image renders usually finish in seconds).
- Editor ScreenshotModal keeps its UX (incl. 4x renders) but polls the job — the render now survives closing the modal or tab; no email opt-in surfaced there.
- Skipped the design doc's optional ~50s inline-wait wrapper: with a 3s poll, the fast-image regression is one poll tick — not worth a second code path.

## #1832 — fal completion webhook

- Submits (mesh **and** image) attach `fal_webhook` to the `queue.fal.run` URL (`falQueueSubmitUrl`), pointing at the new **`falJobWebhook`** — the shared `handleJobWebhook` with a `provider: 'fal'` branch. Same trust model as Replicate/Modal: per-job `webhookSecret` in the query string, body never trusted, authoritative re-fetch via `fetchFalPrediction`, shared idempotent processor + real-time completion email.
- Closed-tab fal jobs now finalize + email in real time instead of the observed ~15–20 min reconciler cadence; client poll + reconciler stay as backstops.

## #1833 — Open-tab email suppression race

The webhook emailed the instant its save committed — always beating the open tab's next ~3s poll ack, so the "suppressed if tab open" contract never held for webhook kinds (observed: video job `577b33f3` emailed while watched). The webhook now waits `WEBHOOK_NOTIFY_ACK_GRACE_MS` (10s ≈ 3 poll cycles) before calling the transactional `sendGenerationReadyEmail`, **only** when the job is opted-in and still pending. An ack during the wait suppresses cleanly; a closed tab's email arrives ~10s later than before; a webhook killed mid-wait loses nothing (`notify.pending` stays set, the sweep delivers). Applies to Replicate, Modal, and the new fal webhook alike.

## #1834 — Pending job card missing in the generator assets sidebar (root cause found)

`useAssets` synced its `currentUser` state from `AuthContext` even when **no provider exists** — which is exactly the generator's situation (React islands, no `AuthProvider` around the Assets island). `useContext` returns the default context value (`currentUser: null`), and the effect `if (contextUser !== currentUser) setCurrentUser(contextUser)` clobbered the real user (set by the `authStateChanged` window event) back to `null` on the next render. So `userId` stayed null and the `generationJobs` listener **never subscribed** — no pending cards, ever, in the generator. Gallery items still loaded because `reloadItems()` reads `auth.currentUser` from the SDK directly via the `assets:refresh` window event, which masked the bug.

Fix: derive the user as `contextUser || firebaseUser`, where `firebaseUser` comes from `onAuthStateChanged(auth, …)` — no cross-source clobbering, works in both apps, sign-out still clears. The editor path is unchanged (provider user wins).

## Docs

`docs/generation-job-queue.md` (provider table, Phase 3 marked done, fal webhook + email-race sections) and the CLAUDE.md queue paragraph updated.

## Testing / deploy notes

- `npm run lint`, prettier, functions eslint: clean. `npm test`: 88 mocha + 377 vitest + 39 component tests pass. `npm run dist:staging` compiles.
- **Deploy:** new function `falJobWebhook`; changed: `generateReplicateImage`, `generateFalImage`, `generateFalMesh`, `getGenerationJobStatus`, `replicateJobWebhook` (module redeploy). The image callables change response shape (`{ jobId }` instead of `{ image_url }`) — same in-place migration precedent as `generateReplicateVideo` (#1780), so deploy functions and hosting together to keep the stale-bundle window short.
- No Firestore rules/index changes needed (owner-read on `generationJobs` and the notify sweep index already exist).

https://claude.ai/code/session_01NQhCyB8t3rHMHN3hQD1R2M

---
_Generated by [Claude Code](https://claude.ai/code/session_01NQhCyB8t3rHMHN3hQD1R2M)_


kfarr:
Testing note -- when clicking on an email indicating job completion, handle case where user is not authed (prompt for auth and then show asset)

PR #1839, opened 2026-07-10
Nav refactor + other tweaks

diarmidmackenzie:
### Latest status

- This branch contains

  - The ExperimentalNavControls that Diarmid built
  - Constrained by 2 flags set up by Kieran: `wasd` & `streeetview` - both defaulting to off.

- With no new flags set, this delivers the following:

  - Cursor-focussed zoom
  - The compass pointer and controls
  - Circle marker for rotation center
  - Rotate about camera center, with nearby fallback for cases where camera points above horizon / out of range.
  - *Improvements to double-click controls (cardinal viewpoints; building height changes; navigate to a point on a lane)
    - (I added this in: not sure why it was excluded as it doesn't depend on other functions and is IMO a pure improvement on previous function - I can revert if you don't agree with this decision)

- With the `wasd` flag set, this also delivers

  - WASD controls

- With the `streetview` flag set, this also delivers

  - LB drag becomes pure pedestal, not perpendicular movement to camera direction
  - Cursor-based rotation center
  - The "street view" letterbox view below a 25 degree tilt, and a distinct control system:
    - Rotate around camera center
    - Cursor-based FOV-zoom
    - LB drag becomes truck, rather than pedestal
  - Viewpoint display/control button (drone / streetview  + sunlight when beneath an object), with Space as a keyboard shortcut for this toggle.
  - 3-phase swoop, with Cursor-based FOV-zoom

  

### Code status

- Substantial refactoring activity undertaken (plus some small, contained bug fixes)
  - Major refactor of 4000+ line Experimental Controls monolith into a structured set of modules (PR#16).  302-item test suite constructed to ensure behaviour preserved (PR#15)
  - Address undesirable quirks identified in refactor (PR#16 was intentionally zero-behaviour changed and preserved the quirks) to simplify code and improve behaviour (PR#17)
  - Address GC performance issues with dynamically allocated THREE objects (PR#18)
  - PR#19 addressed some issues with PR#18 and PR#17 after they had merged.
  - Added double-click improvements to the base version (no flags enabled) (PR#20)
  - Code readability and code commenting improvements (PR#21 & PR#22)

- There may be a few further minor tweaks - running a final automated review, and will update later.

Note PR numbers above refer to PRs in https://github.com/diarmidmackenzie/3dstreet/pulls

### Testing with Default controls

Issues identified:

- (addressed) New double-click behaviour was not included in the default controls, even though it is purely additive.  Made 1 line change to bring it into this scope PR#20.

- Minor inconsistency between rotation centers:

  - Compass L/R rotations rotate around camera
  - Above-horizon fallback rotation center is a constructed point a few yards in front of the camera

- When camera is just above a surface (e.g. at street level), there is no way to look upwards

  - Rotation center is a point on the ground just in front of the camera, so rotating to look upwards pulls the camera underground.  On gesture release, the  camera view reverts to the previous view.
  - This is arguably a regression vs previous function, where you could look upwards, albeit from below the ground plane
  - Options:
    1. Do nothing: accept this.
    2. Change fallback rotation center to be the camera, consistent with Compass L/R rotations, and with future streetview mode.  Allows camera to look directly upwards
    3. Change recovery path so the user is elevated to surface, but camera rotation is retained, rather than being reversed.

  

### Enabling wasd=on

No new issues identified.  Known issues:

- We need a discovery mechanism for the previous key bindings that have been replaced.

### Enabling streetview=on

No new issues identified.  Known issues:

- Swoop zoom phase 2/3 boundary doesn't feel right - not easy to avoid crossing the boundary and going into FOV zoom unintentionally when zooming in, or into pedestal-up when zooming out in phase 3.

- Overall UX testing of 3-phase zoom is still required.  This is the least intuitive / most novel mechanic, and may not land well for users.

- Drone / Street view icon: This shows the intended state that will be switched to when clicked on; it does not show the current state.  This may be confusing: needs UI design attention.

- Open question whether you want to be able to test/demo the tilt/streetview mode independently of the 3-phase swoop zoom.  We could put these behind distinct flags, if that would be useful.

  

### Open Questions

- Are you happy with the call to enable the new Double Click behaviour by default?  I think it is a strict improvement on the previous behaviour.

- PR#20 (promoting Double Click behaviour) flagged an issue with analytics events on this potentially being missing.  Is that a gap you want addressed?

  > Adversarial review (SHIP-WITH-FIXES) found the un-gating drops canvas double-click out of the focus PostHog discovery funnel, and the new nav emits no nav_control_used (a pre-existing gap — the ?streetview=on path was already silent). Whether double-click should feed focus, get its own funnel step, or stay silent is a product-analytics call — I flagged it rather than bake in a guess. Details in reports/039-orchestration-log.md + verbatim review in reports/039-review-appendix.md.

- Do you have what you need for UX testing for `wasd` and `streetview`?


kfarr:
Heads up: this work now continues in #1851, which carries this branch plus default-tier bug fixes (context-button gating, double-click focus reverted to the legacy deterministic framing) and the merge with main's unified Viewer/Play construct (#1812/#1845/#1848). Review is happening over there — leaving this PR open for now for reference and discussion continuity.

kfarr:
closed by https://github.com/3DStreet/3dstreet/pull/1851

Issue #1, opened 2017-07-07
hide the header on mobile


kfarr:

kfarr:
https://twitter.com/mertonium/status/883278103446822912

PR #1862, opened 2026-07-19
Geospatial map opacity + 3d-tiles-renderer 0.5.0 performance upgrade

### LINKED ISSUE #1122: use load-model event for 3d tiles blending options
Blending mode changes exhibit "popping" and issues sometimes where newly loaded tiles don't have blending applied or there is a delay before it is applied. This could probably be optimized to more closely follow the [custom material demo here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/f90b36cdd471a9f47cf118ed5fb1176de4bc5689/example/customMaterial.js#L123)

`	tiles.addEventListener( 'load-model', onLoadModel );`

https://github.com/NASA-AMMOS/3DTilesRendererJS?tab=readme-ov-file#custom-material

### LINKED ISSUE #1235: allow custom opacity for 3d tile map blending

### LINKED ISSUE #1236: remove non-opacity options for 3d map blending

### LINKED ISSUE #1738: geospatial opacity for all map types
- as a user creating a 3dstreet scene i'd like to be able to specify a custom opacity amount from 0 to 100% and not have to choose from present blending modes; the existing blending modes do nothing so remove those; i'd like that opacity amount to apply for all geospatial map types
- as a user using a blending mode, i don't want to have a performance impact for using a custom material to the extent that it's practical
- i don't want to see "flashes" of new tiles popping in before their opacity is set

kfarr:
## Summary

Unified PR for #1738, #1236, #1235, and the tile-blending performance problem tracked in #1122. The theme: one simple opacity control for geospatial maps, implemented the way the NASA-AMMOS `3DTilesRendererJS` docs recommend, on the latest version of that library, with the per-frame overhead removed.

### Performance (the big one)

- **Deleted the `blending-opacity` component.** Its `tick()` traversed the entire tile scene graph and rewrote every material **every frame** — this was the main FPS drain whenever blending was enabled. Opacity is now applied per tile exactly once, from the `TilesRenderer` `'load-model'` event, before the tile's first render (the approach suggested in #1122 and the library's custom-material example). Changing opacity later does a single `forEachLoadedModel` pass.
- **No more tileset teardown.** Toggling blending used to destroy and recreate the entire Google 3D Tiles element (full re-download/re-parse). Opacity changes are now just material property writes.
- **Fixed stale camera accumulation.** `google-maps-aerial`'s tick called `tiles.setCamera(sceneEl.camera)` every frame, but `setCamera` *registers* cameras — it never replaces them. Every editor ↔ viewer ↔ drive-mode camera swap left the previous camera registered, so the tileset kept frustum-testing and downloading tiles for cameras no longer in use. The component now tracks the active camera and `deleteCamera`s the old one on swap.
- **Upgraded `3d-tiles-renderer` 0.4.21 → 0.5.0** (latest), picking up upstream scheduling/traversal fixes (e.g. tiles getting stuck visible during traversal) and shared processing queues.
- Tiles keep their stock materials — no custom shader — so opacity 1.0 has zero extra draw cost and <1.0 costs only standard alpha blending.

### Features / UX (#1738, #1236, #1235)

- New `street-geo` property `opacity` (0–100 %), replacing `blendingEnabled` + `blendMode` presets. The Geo sidebar's "Blending" toggle + preset dropdown becomes a single **Map Opacity (%)** row (section renamed "Opacity & Flattening").
- Opacity applies to **Google 3D Tiles** and the **Mapbox 2D** satellite plane. OSM 2.5D tiles/buildings are loaded by the external osm4vr lib with no per-tile load hook, so they're deferred as a follow-up (per issue guidance that focusing on the 3DTilesRenderer implementation first is acceptable).
- **No tile pop-in flashes**: because opacity is set in `'load-model'`, a tile is never rendered opaque before its opacity is applied.
- The non-functional blend modes (Darker/Lighter/Normal/etc.) are removed.

### Migration / compatibility

- **Legacy scenes**: `blendMode`/`blendingEnabled` stay in the schema (deprecated, hidden from UI) so old saved JSON parses without warnings. On load, `blendingEnabled: true` scenes are converted once: `30% Opacity` → 30, `60% Opacity` → 60, Darker/Lighter/Normal → 100 (those modes were already broken in practice).
- **Scene orientation preserved across the 0.5.0 upgrade**: 0.5.0 removed `setLatLonToYUp()`. The replacement `ReorientationPlugin` defaults to a different frame (+X west / +Z north vs the legacy +X north / +Z east). Using `azimuth: π/2` reproduces the legacy matrix exactly — verified numerically (max element diff ~2e-9 against the old implementation) — so every previously saved geo scene stays aligned.
- i18n: en/es/fr/pt-BR locales updated; `npm run i18n:check` passes.

### Testing

- `npm run lint` ✅
- `npm test` (mocha core + 742 vitest + 50 browser component tests) ✅
- `npm run dist` production build ✅ (0.5.0 is ESM-only with `three/addons` imports; bundles cleanly against the existing `three`-external webpack setup)
- Not yet manually verified in a live browser against the Google 3D Tiles API — that's why this is a draft; would appreciate a smoke test of: tile load at a saved geo scene (orientation!), opacity slider, flattening toggle, and drive-mode entry/exit with tiles active.

Closes #1738, closes #1236, closes #1235, closes #1122

https://claude.ai/code/session_01XtNLzXY4Kt1uVj97jk9L6y

---
_Generated by [Claude Code](https://claude.ai/code/session_01XtNLzXY4Kt1uVj97jk9L6y)_


kfarr:
## Code review findings — follow-up checklist

Automated multi-angle review of this branch (8 finder passes + adversarial verification against the installed 0.5.0 sources). Nothing here blocks the smoke test; these are queued as follow-ups with suggested fixes. Good news first: the `ReorientationPlugin` azimuth-π/2 math was independently re-verified to reproduce the legacy `setLatLonToYUp` frame exactly, and every 0.5.0 API used (`deleteCamera`, `forEachLoadedModel`, `load-model {scene}`, `getAttributions`, `transformLatLonHeightToOrigin(lat, lon, height, azimuth)`) exists as called. The core upgrade looks correct.

### Correctness

- [x] **1. Legacy migration can double-create the osm3d layer** — `src/aframe-components/street-geo.js:122`
  The migration's 0 ms `setTimeout` → `setAttribute` re-enters `update()` while `osm3dCreate()`'s script is still async-loading (`loadScript` in `src/utils.js` has no dedupe, and `this.osm3d` is only assigned in the load callback). The re-entered update sees `!this.osm3d` (guard at line 132) and creates again → duplicate osm3d + osm3dBuilding entities; `self['osm3d']` tracks only the last pair, so a later map-type switch orphans the first pair permanently.
  **Fix:** set an in-flight flag in `osm3dCreate()` (`this.osm3dPending = true`, cleared in the load callback) and include it in the creation guard. (Longer term, see #4 — moving the migration to load time removes the trigger entirely.)

- [x] **2. Migration applies legacy presets to map types blending never affected** — `src/aframe-components/street-geo.js:121`
  On main, blending was gated on `if (this.google3d)` and the toggle UI was google3d-only, but switching map type never reset `blendingEnabled` — so scenes saved with `maps='mapbox2d'` + stale `blendingEnabled: true` rendered fully opaque. The new migration runs on any truthy `blendingEnabled` and writes `opacity: 30` (default blendMode), and mapbox2d now honors opacity → those scenes' satellite plane suddenly renders 30 % transparent with no user action, persisted on next save (same silent rewrite for `maps='none'`, surfacing later on a switch to google3d).
  **Fix:** in `migrateLegacyBlendMode()`, only carry the preset over when `data.maps === 'google3d'`; otherwise migrate to `opacity: 100`.

- [x] **3. Opacity 0 promises "invisible" but keeps full render + network cost** — `src/aframe-components/street-geo.js:25`
  At opacity 0 the google3d tileset still frustum-tests, downloads metered Google 3D Tiles API data, and draws every tile at alpha 0; the mapbox2d plane likewise keeps fetching/rendering. Notable given the PR is performance-themed.
  **Fix:** at `opacity === 0`, set `visible: false` on the layer entity via `setAttribute` (per the CLAUDE.md visibility rule) and skip `tiles.update()`; restore on opacity > 0.

- [x] **4. Migration timeout never re-checks `blendingEnabled`** — `src/aframe-components/street-geo.js:123`
  The deferred callback guards only on component identity. If two street-geo writes ever land in the same task queue while `blendingEnabled` is still true, both timeouts run the migration — and since `migrateLegacyBlendMode()` reads `this.data.blendMode` (untouched by the first pass), the second pass re-clobbers `opacity` back to the legacy 30/60, wiping any intervening edit. No current call site stacks writes like this, but nothing prevents it.
  **Fix (one line):** re-check `this.data.blendingEnabled` inside the timeout callback. **Deeper fix:** move the migration into the existing load-time migration layer (`json-utils_1.1.js` already does street-geo `visible→maps:'none'` migration around line 437) as a pure data transform — that removes the timer, the re-entrancy (#1), and the stale-UI window (#5) in one move.

- [x] **5. Migration bypasses `entityupdate`, leaving an open Geo sidebar stale** — `src/editor/components/elements/GeoSidebar.jsx:483`
  Raw `el.setAttribute` emits no editor `entityupdate`, and that event is GeoSidebar's only refresh trigger. If the geo tab is already open when a legacy scene loads in place, the map drops to 30 % while the slider still shows 100; the first drag then jumps the map. Narrow window (geo tab defaults closed), but confirmed mechanism.
  **Fix:** emit `entityupdate` (or the equivalent Events emit) after migrating — or land the deeper fix in #4, which makes the migrated value present before the sidebar ever reads it.

- [x] **6. Arrow keys broken on the new Map Opacity number input** — `src/editor/components/elements/GeoSidebar.jsx:232`
  `NumberWidget.onKeyDown` steps by hardcoded ±0.01 and ignores the `step` prop (`NumberWidget.js:143-152`). With `precision={0}`: ArrowUp → `parseInt(30.01)` → 30 (no-op, but still fires a no-op entityupdate per keypress); ArrowDown → `parseInt(29.99)` → 29 (steps −1). Asymmetric keyboard behavior. Pre-existing widget bug (also affects precision-0 PropertyRow fields), newly surfaced by this UI.
  **Fix:** make `onKeyDown` use `this.props.step` (default 1) instead of 0.01 — fixes it everywhere at once.

- [ ] **7. "Opacity & Flattening" header shown for mapbox2d, which has no flattening** — `src/editor/components/elements/GeoSidebar.jsx:932`
  The collapsible renders for both map types but the flattening rows are gated on google3d (line 946), so mapbox2d users see a header promising a control that never appears.
  **Fix:** conditional header — "Opacity" for mapbox2d, "Opacity & Flattening" for google3d (needs one extra i18n key).

### Performance

- [x] **8. Slider drag traverses every cached tile scene per input event** — `src/editor/components/elements/GeoSidebar.jsx:225`
  Each `input` event (60–120/s while dragging) dispatches entityupdate → street-geo → google-maps-aerial → `forEachLoadedModel`, which walks the **entire in-memory tile hierarchy** — the 0.5.0 LRU cache retains up to ~6000–8000 loaded tiles — plus a `scene.traverse` per loaded model. Thousands of traversals per drag pixel on a well-browsed scene. (The `needsUpdate` storm at the 1.0↔0.99 boundary is one-time by design and fine; undo is already coalesced via `updatable`.)
  **Fix:** coalesce through rAF — store the latest value and issue one `execute`/`applyOpacityToLoadedTiles` per animation frame. Small change, preserves live preview.

### Cleanup / robustness

- [x] **9. Two hand-rolled copies of the opacity slider row** — `src/editor/components/elements/GeoSidebar.jsx:188`
  `MapOpacityRow` repeats `MaterialControls.jsx:138-160`'s markup and dispatch almost exactly — the diff even hoisted the SCSS out of `.material-controls` into a shared rule, but left two JSX copies to drift (one has a read-only % span, the other a NumberWidget).
  **Fix:** extract a shared `OpacitySliderRow` (natural home: `src/editor/components/widgets/`, next to NumberWidget) parameterized by min/max/step and value readout; use from both.

- [x] **10. Location-update path pokes ReorientationPlugin's undocumented fields** — `src/aframe-components/google-maps-aerial.js:294`
  `plugin.lat`/`plugin.lon` are incidentally public today (the pending `load-root-tileset` callback re-reads them) but the shipped `.d.ts`/API.md document them only as constructor options — a 0.5.x refactor to private fields would silently break pre-root-load location updates. `height`/`azimuth` are also duplicated between init and update.
  **Fix:** on location change, `tiles.unregisterPlugin(this.reorientationPlugin)` (public API; its `dispose()` resets the group transform) and register a fresh plugin built by one shared options factory, so the plugin's own supported init path handles both root-loaded and root-pending cases and `LEGACY_AZIMUTH` lives in one place.

- [x] **11. `data.opacity / 100` repeated at 4 sites, two mechanisms** — `src/aframe-components/street-geo.js:194, 229, 265, 273`
  The percent→fraction boundary is re-derived in `mapbox2dCreate` (inside a material template string), `google3dCreate`, `google3dUpdate`, and `mapbox2dUpdate` (three-arg setAttribute). Miss one on a future change and a map type renders at 100× off.
  **Fix:** one `opacityFraction()` helper on the component used at all four sites; have `mapbox2dCreate` set opacity via the same `setAttribute('material', 'opacity', …)` call as `mapbox2dUpdate` instead of baking it into the string.

- [x] **12. `blendMode`'s `oneOf` list is dead metadata** — `src/aframe-components/street-geo.js:34`
  No consumer remains: the dropdown UI is gone, A-Frame core never reads `oneOf`, the generic inspector panel can't reach street-geo (`reference-layers` routes around `ComponentsContainer`), and the migration reads only the two opacity keys. The list invites someone to "re-support Darker/Lighter" against a property the adjacent comment declares deprecated.
  **Fix:** reduce to `blendMode: { type: 'string', default: '30% Opacity' }` — the default is the only part that still matters for legacy parse parity.

### Noted, no action

- The **Darker/Lighter drop is destructive on next save** for any legacy scene that actually used those modes — the code comments and PR body document this as intentional per #1738, so it's flagged here only as a heads-up.
- Refuted during verification (not issues): user-scene exposure of the deleted `blending-opacity` component (never reachable via any product UI path); the `setCamera(undefined)`-at-init edge (byte-for-byte pre-existing on main); per-drag GeoProvider re-render cost (zero `useGeoContext` consumers; React bails out on the stable `children` element).

**Suggested triage order:** #2 and #1 first (user-visible regressions on legacy scene load), then #8 and #3 (they land in the same code paths as the smoke test), then the rest opportunistically. Fixing #4's deeper variant (load-time migration in `json-utils_1.1.js`) resolves #1, #4, and #5 together.

https://claude.ai/code/session_01XtNLzXY4Kt1uVj97jk9L6y

---

PR #1868, opened 2026-07-19
fix(nav): wheel zoom-out escape floor (#1865) + cursor-ray pan fallback (#1867)

### LINKED ISSUE #1865: regression: cannot zoom out after focusing on 0 0 0 entity like geojson import
http://localhost:3333/#/scenes/953f87f0-3458-46ce-bcc2-a4fea64100eb
or dev server
double click on tax parcels zoningviz data layer
try to zoom out using scroll wheel

workarounds:
- press the zoom out button in the action bar
- press the plan mode button (compass icon)

### LINKED ISSUE #1867: regression: cannot pan up while looking out at the horizon and grabbing top half of screen
reproduction here:

https://github.com/user-attachments/assets/97475fcc-00f8-492c-b8d5-5d19d529d330

scene:
https://3dstreet.app/#/scenes/a6e3fe2f-48ba-473c-aa97-e4edd0f5de57

kfarr:
Fixes #1865, fixes #1867 — two navigation regressions vs legacy EditorControls.

## #1865 — cannot zoom out after focusing a 0,0,0 entity (geojson layer)

**Root cause:** `focus()` on an entity whose bounding box is empty (e.g. a geojson data layer whose meshes aren't measurable under its object3D) falls back to `distance = 0.1` and parks the camera **~0.25 m** from the entity origin. The new wheel dolly is purely multiplicative in the camera→anchor distance (`ZOOM_PER_WHEEL_TICK` = 5%/tick), so from 0.25 m each zoom-out tick moved **~1 cm** — the wheel read as dead. Legacy EditorControls always stepped at least `minSpeedFactor(8) × zoomSpeed(0.1) = 0.8 m/detent`, which is why this never happened before. (The ActionBar zoom-out button and Plan View use different math — hence the workarounds in the issue.)

**Fix:** new `WHEEL_ZOOM_OUT_MIN_ANCHOR_DIST_METRES = 16` (TH-80). When zooming **out** closer than 16 m to the anchor, `cappedDollyStep` sizes the step as if the anchor were 16 m away → `16 × 0.05 = 0.8 m/tick`, exactly legacy parity. Zoom-**in** is untouched (stays asymptotic, never shoots through the anchor); exact in/out reversibility inside the floor radius is deliberately traded for never-stuck. The lateral lurch cap still bounds a floored step.

## #1867 — cannot pan up looking at the horizon, grabbing the top half of the screen

**Root cause:** when the pan grab-point raycast misses everything (grab over sky above the horizon), `worldPointAt`'s Step-3 fallback returned 30 m along the camera's **centre-forward**, not the **cursor ray**. The LB-pan gestures anchor their drag plane on that point and compute per-move deltas as `anchor − pointUnderCursorNow`, so the first move applied the cursor↔screen-centre offset as a spurious camera lurch (≈ half a screen's worth, downward for a top-half grab). Dragging from the top half toward the centre to pan up therefore netted ~zero — exactly the repro video.

**Fix:** the Step-3 fallback now lies on the cursor ray (the raycaster is already set from the cursor NDC). A real mesh/ground anchor is already on the cursor ray, so those paths are byte-identical; every other fallback consumer (wheel dolly, orbit pivot, Phase-3 re-aim, swoop classifier) branches on `source === 'fallback'` and never uses the fallback point itself, so this only changes the pan anchor — covers all three pan sub-modes (screen / truck / pedestal) in one place.

## Tests

- `cursorAnchor.test.js` — fallback lies on the cursor ray, not centre-forward
- `navMath.test.js` — `cappedDollyStep` floor: zoom-out floored / zoom-in untouched / far interior unchanged / camera-on-anchor safe (no NaN) / lateral cap still binds
- `ExperimentalControls.lbPan.test.js` — sky-grab at the horizon pans up ~4 m for a 90 px drag with no lurch (pre-fix this nets downward)
- `ExperimentalControls.wheelZoomEscape.test.js` — one out-tick from a near-anchor pose moves ≥ 0.5 m; sustained zoom-out escapes past 16 m in 25 frames (pre-fix ~6 m); zoom-in keeps the 5% step

`npm run test:core`, `test:modern` (752), and `test:components` (50, real Chromium) all pass; lint + prettier clean.

## Verification notes for review

- #1865 repro: dev server → open the linked scene → double-click the tax-parcels zoningviz layer → scroll-wheel zoom out should now pull away immediately.
- #1867 repro: look out at the horizon → left-drag starting in the top half (sky) → the camera should pan up smoothly with no initial downward jump.

https://claude.ai/code/session_01VMhNiU7cU2N8eBzkRQTy7X

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMhNiU7cU2N8eBzkRQTy7X)_


kfarr:
After manual testing, this is not ready:
- for 1867 -- this is a slight improvement but it also causes another issue -- then as a user if I am panning upward by click-and-drag from the horizon and dragging down, the first pan up works as expected, however by the second or third, it is easy to click the distant ground at that point since the elevation is higher, and then the rate of ascension is very high and "catapults" the viewer out of the frame; this is worth more discussion -- this relates to some of the core issues that Diarmid identified and attempted to provide a workaround for with the horizon swoop available in the experimental controls; it's worth double checking how this is handled in legacy (screen space panning?) and the attempt that I made to include that and why Diarmid recommended removing that again -- although I could be remembering that wrong; i'm wondering if below a certain angle we're doing screen space pan instead of ground pan, but i think this needs more research
- for 1865 -- i notice that it zooms out further but there is still a case such as :
- http://localhost:3333/#/scenes/953f87f0-3458-46ce-bcc2-a4fea64100eb
- where when I focus on the tax parcels layer it is effectively 0 0 0 as it does not have an accessible bounding box. When the user attempts to zoom out and is placing the cursor at the center of the screen, it still does not do any zoom out motion. sometimes if i am doing the zoom out scroll motion and I move the cursor around some other places in the vicinity of that 0 0 0 focus will allow me to zoom out; i'm wondering why the cursor location will determine it, one example I can think of is that the screenspace cursor position is intersecting with different object; oh yes i wonder if 1865 the root cause is that the 0 bounding box element is at 0 0 0 and that the zoomout cursor center is intersecting with the origin marker (not a scene element might be added just as a three.js object3d) or another similar overlap there.

kfarr:
Researched both threads, reproduced the failure mechanisms, and pushed two follow-up commits (`ec3ed53`). Findings:

## 1867 — the "catapult" on the second/third pan

**History check (you're remembering it right, with one nuance):** legacy `EditorControls.pan()` moved the camera at `max(minSpeedFactor=8, distanceToCenter) × panSpeed` per pixel — **completely independent of what sits under the cursor** (no raycast at all). Your Stage-1 restore (`583676b`, per `07-phased-rollout-plan.md`) brought back the legacy **direction** semantics — one continuous ⊥-camera-plane screen-space pan at every tilt, replacing the ground-plane pan that Diarmid flagged for lurching at low tilt and stranding users. So with street-level off we *are* already doing screen-space pan below (and above) every angle — the pan basis isn't the problem.

**What regressed vs legacy is the *rate*:** the new pan anchors its drag plane on the raw cursor raycast hit, and world-per-pixel scales with that plane's depth. The raycast is unbounded for mesh hits and 2000 m for the ground plane. After the first pan-up raises you, the top-half cursor at the horizon now lands on distant ground a few hundred–2000 m out → each pixel is worth ~0.5–3 m of world → catapult. My cursor-ray fallback fix removed the sky-grab lurch but did nothing about far *real* hits.

**Fix pushed (TH-81):** cap the anchor's camera-distance at `max(30 m, 2 × camera→center distance)` — essentially legacy's distance-to-center speed bound with ×2 slack. A beyond-cap anchor is pulled **in along the cursor ray**, so the first-move delta stays exactly 0 (no lurch); normal grabs (anchor near the working distance) keep exact point-under-cursor tracking; only far-outlier grabs pan slower than the cursor — the same trade legacy made for *every* point. Tunable: gain 1 = strict legacy rate; larger keeps exact tracking farther out but re-admits the catapult.

## 1865 — the remaining dead zoom-out on the tax-parcels scene

**Origin-marker hypothesis — ruled out by code:** the RGB-axes origin indicator (and grid, selection/hover boxes) live in `inspector.sceneHelpers`, which is a *separate* `THREE.Scene` (`src/editor/index.jsx:104`), not a child of `sceneEl.object3D` — the nav wheel's cursor raycast never traverses it.

**Reproduced the real mechanism in the test harness:** with the post-focus pose (camera 0.25 m from origin) I ran every scene variant. All plain-dolly regimes now move fine with the escape floor (0.84 m/tick). But with **street-level nav on and a collision floor under the camera** — Google 3D Tiles qualify as a floor, and that scene has them — AGL ≈ 0 puts the wheel in the swoop's **Phase-3 FOV regime**: the first ~5–8 zoom-out detents only widen FOV 50→75 with **zero camera movement**, which is nearly invisible over uniform content like the parcels layer, then the rate-capped Phase-2 ascent starts. That matches "it zooms out further but this scene still does nothing at first."

**Fix pushed (root cause):** `focus()` on an empty/degenerate-bbox target (the geojson layer, lights, empty wrappers) now stands off **10 m** instead of legacy's 0.1 — the camera frames the entity origin from `(0, 5, 25)` where every nav regime is immediately usable. This removes the pathological pose class entirely rather than patching each regime it breaks. (Legacy `EditorControls.js` still has the 0.1 fallback — happy to mirror it there too if you want scheme parity.)

**Two things I'd like to confirm on your side:**
1. Which control scheme were you testing (standard vs experimental/`?streetview=on`)? If the center-cursor deadness happens with street-level **off** even after this branch, the Phase-3 explanation doesn't cover it and I'd want to know what the center ray hits in that scene.
2. Design question, out of scope here: should Phase-3's FOV-first zoom-out be shortened (it's ~8 silent detents from the default 50° fov before anything moves)? That's swoop feel territory (KD-08/KD-12), so I left it alone.

All suites green (mocha core, 754 vitest, 50 browser-mode component tests), lint + prettier clean.

---

kfarr:
Pushed `e4b5032` — sky-grab rate harmonization per the feel-test feedback: `_capAnchorReach` now re-depths outlier anchors along the cursor ray in **both** directions. A beyond-reach real hit is still pulled in (anti-catapult), and a sky-grab fallback anchor (previously fixed at 30 m — the "very slow above the horizon" you observed) is pushed **out** to the same `max(30 m, 2 × camera→center)` reach. Above- and below-horizon grabs now pan at the same working-distance rate; real hits within reach keep exact point-under-cursor tracking, and the no-first-move-lurch property holds everywhere since the re-depthed anchor stays on the cursor ray. Added a rate-parity regression test (same 100 px drag just above vs just below the horizon lands within ~25%). All 352 nav tests green.

---

kfarr:
This does indeed solve the 2 issues that I originally raised, and I can't find other new regressions this may have caused, so I'm ok to merge this.

Issue #1677, opened 2026-06-15
[plg] Redesign GeoModal and sidepanel activated "Location Set" state


kfarr:
## Context

The PLG geo upsell work (#1654) reworked the **empty** and **located-but-not-activated** states of the Geospatial panel into a single `GeoHero` CTA (spine model: geo-specific controls gated behind an activated location, source-aware provenance copy, geo-token reinforcement).

Two surfaces were intentionally left out of that scope and now look outdated next to the new design:

1. **The GeoModal itself** — the map-picker / set-change-location modal still uses the older layout and copy (tips list, token messaging, button states). Bring it in line with the new hero/spine visual language and token framing.

2. **The activated "Location Set" panel state** — once a scene is activated, the panel reverts to the legacy layout (`✅ Location Set` badge + Map Type grid + Location/Intersection/Elevation rows + Blending & Flattening). This activated view was not redesigned and looks dated against the new not-activated hero.

## Scope
- Redesign GeoModal layout/copy consistent with the `GeoHero` spine.
- Modernize the activated "Location Set" panel (status, map-type selector, location details, blending).
- **3D-canvas callout** for the located-not-activated state — deferred from #1654 (the in-panel hero shipped; the lower-right viewport callout did not).

## Already done (out of scope here)
- Empty + located-not-activated hero, control gating, and source provenance: #1654.
- Asset storage upsell: #1644.

Implemented on branch `feat/plg-upsell-part2`.

PR #1899, opened 2026-08-01
Fix GeoModal ignoring typed address when no autocomplete suggestion is selected

### LINKED ISSUE #1897: Geo: Set Location ignores typed address — lat/long not updated, scene loads default location
## What happened

While setting a scene location, a user typed a full intersection/address into the location search field and clicked the green **Set Location** button, but the scene's longitude/latitude did not update — the scene loaded at the default San Francisco location instead. Repeated attempts with the correctly-typed address kept committing wrong coordinates. Observed live on an onboarding call (2026-07-31).

## Repro (as observed; needs confirmation)

1. Open Geospatial panel → Change Location
2. Type a full address/intersection into the search field — possibly *without* explicitly selecting an autocomplete suggestion
3. Click **Set Location**
4. Scene loads the default (or stale) location rather than the typed one

## Suspected cause (unconfirmed — needs exploration)

Free-typed text may only update the display string, with the underlying lat/long state changing only when an autocomplete suggestion is explicitly selected. If so, **Set Location** commits stale/default coordinates while the field shows the correct address.

## Workaround

Clicking the red marker directly on the map to set the center point works correctly.

## Notes

Distinct from the GeoModal redesign (#1677) — this is a functional bug in the current modal, and whatever the fix is should carry over to the redesigned modal.

kfarr:
Fixes #1897

## Root cause

In `GeoModal.component.jsx`, typing into the location search field never updated the lat/long state:

- The search `Input`'s `onChange` handler was a no-op (`onChange={(value) => {}}`), so typed text was discarded.
- Coordinates were only set inside `onPlaceChanged`, and only when `autocomplete.getPlace()` returned a place **with geometry** — which Google Places Autocomplete only provides when the user explicitly clicks/selects a suggestion from the dropdown. Pressing Enter without selecting a suggestion returns a place with just the raw text and no geometry, so the handler silently did nothing.
- "Set Location" then committed whatever stale `markerPosition` was in state: the default San Francisco coordinates on first use, or the last-resolved location on later attempts. There was no error or feedback, so the mismatch was only discovered after the ~5–10 s tile load.

This also explains the confusing "goes to a different street than I typed" retry behavior: the committed coordinates came from a previous resolution, not the current text. Note the typed text also *appears* remembered between opens — the modal stays mounted and the uncontrolled input keeps its DOM value, and after a successful commit the reverse-geocoded `locationString` of the committed point shows as the input's placeholder. Neither ever fed back into lat/lng.

## Fix

- Track typed search text in a `pendingSearchQueryRef` (non-empty means the marker does **not** yet reflect the typed text). Selecting a suggestion, successfully geocoding the text, or explicitly picking a location another way (map click, manual coordinates) clears it — the last explicit location choice always wins.
- **Enter without selecting a suggestion:** `onPlaceChanged` now geocodes the typed text via `google.maps.Geocoder` and moves the marker, so the user gets immediate map feedback.
- **Set Location with unresolved typed text:** `onSaveHandler` geocodes the pending query first and commits those coordinates (stamping `manual` provenance). If geocoding finds nothing, it shows an error notification and keeps the modal open instead of silently committing stale coordinates. This also closes the in-flight-geocode race — the save path always resolves the current text before committing.
- The map-click and manual lat/long entry paths are unchanged (they already worked).
- New i18n message `geoModal.locationSearchNotFound` extracted to `en.json`, with hand-written es / pt-BR / fr translations and manifest entries (`npm run i18n:check` passes).

## Manual test plan (~10 min)

Setup: `npm start` → http://localhost:3333, open any scene, sign in with an account that is Pro or has geo tokens. Open **Geospatial → Change/Set Location** to reach the GeoModal. Keep the browser devtools console open to spot errors.

1. **Type + Set Location, no suggestion click (the #1897 repro, ~3 min).** Type a full address into the search field (e.g. `Washington Street and Scott Street, Chualar, CA`). Do **not** click any autocomplete suggestion and do **not** press Enter. Click **Set Location**.
   - ✅ Expected: scene loads at the typed address; success overlay's Location string matches it.
   - ❌ Before this fix: scene loads default San Francisco / previous location.
2. **Enter key, no suggestion click (~1 min).** Reopen the modal, type a different address, press **Enter** without clicking a suggestion.
   - ✅ Expected: map preview pans and the red marker + lat/long field update to the typed address *before* you click Set Location.
3. **Gibberish query fails loudly (~1 min).** Reopen the modal, type something unresolvable (e.g. `zzzqqqxxx nowhere`), click **Set Location**.
   - ✅ Expected: error notification "Could not find …", modal stays open, no location commit, **no geo token consumed**.
4. **Regression — suggestion click still works (~1 min).** Type an address and click one of the dropdown suggestions, then **Set Location**.
   - ✅ Expected: marker jumps on selection; commit goes to the selected place.
5. **Regression — map click and manual coords still work (~2 min).** Click a spot on the map preview → marker and lat/long update → Set Location commits it. Then paste coordinates directly into the Lat, Long field (e.g. `36.570, -121.516`) and Set Location.
   - ✅ Expected: both commit the shown coordinates, as before.
6. **Typed text then map click — last choice wins (~1 min).** Type an address (no suggestion click), then click a *different* spot on the map, then **Set Location**.
   - ✅ Expected: the scene commits the **map-clicked** point, not the abandoned typed text. This mirrors the workaround flow from the original report (failed search, then map click) — the click must not be overridden.

Optional if time remains: repeat step 1 in a fresh reload with a scene that already has a saved location, confirming the placeholder text (previous location string) alone doesn't get committed as a search.

## Automated testing

- `npm run test:core` (Mocha) — pass
- `npm run test:modern` (Vitest, 756 tests) — pass
- `npm run test:components` (browser component tests, 50 tests) — pass
- ESLint + Prettier clean on changed files

## Notes

- The GeoModal redesign planned in #1677 should carry this behavior forward: never commit coordinates that don't reflect the current search text, and fail loudly when the text can't be resolved.
- Possible follow-up (out of scope here): show where the typed text resolved to *before* committing, since each failed attempt costs a long tile-load feedback loop.

https://claude.ai/code/session_01Xkao1vfgcZtbhEcqBQqbww

### DIFF
--- a/src/editor/components/modals/GeoModal/GeoModal.component.jsx
+++ b/src/editor/components/modals/GeoModal/GeoModal.component.jsx
@@ -177,7 +177,10 @@ const GeoModal = () => {
   const setMarkerPositionAndElevation = useCallback((lat, lng) => {
     if (!isNaN(lat) && !isNaN(lng)) {
       // A user-driven pick (map click, search, manual coords) — provenance
-      // becomes 'manual' on save.
+      // becomes 'manual' on save. It also supersedes any typed-but-unresolved
+      // search text: the last explicit location choice wins, so a map click
+      // after an abandoned search commits the click, not the stale text.
+      pendingSearchQueryRef.current = '';
       userChangedLocationRef.current = true;
       setMarkerPosition({
         lat: roundCoord(lat),
--- a/src/editor/components/modals/GeoModal/GeoModal.component.jsx
+++ b/src/editor/components/modals/GeoModal/GeoModal.component.jsx
@@ -206,10 +206,12 @@ const GeoModal = () => {
   }, []);
 
   // Resolve a free-typed search string to coordinates. Returns
-  // { lat, lng } or null when nothing matched / the API is unavailable.
+  // { status: 'ok', lat, lng }, { status: 'notFound' } when the text didn't
+  // match anything, or { status: 'unavailable' } when the Geocoder itself
+  // is missing or failing (API not loaded, quota, network).
   const geocodeSearchQuery = useCallback(async (query) => {
     if (!window.google?.maps?.Geocoder) {
-      return null;
+      return { status: 'unavailable' };
     }
     if (!geocoderRef.current) {
       geocoderRef.current = new window.google.maps.Geocoder();
@@ -217,10 +219,15 @@ const GeoModal = () => {
     try {
       const { results } = await geocoderRef.current.geocode({ address: query });
       const location = results?.[0]?.geometry?.location;
-      return location ? { lat: location.lat(), lng: location.lng() } : null;
+      return location
+        ? { status: 'ok', lat: location.lat(), lng: location.lng() }
+        : { status: 'notFound' };
     } catch (error) {
-      // The Geocoder rejects on ZERO_RESULTS and friends — treat as not found.
-      return null;
+      // The Geocoder rejects on any non-OK status; only ZERO_RESULTS means
+      // the query itself found nothing.
+      return error?.code === 'ZERO_RESULTS'
+        ? { status: 'notFound' }
+        : { status: 'unavailable' };
     }
   }, []);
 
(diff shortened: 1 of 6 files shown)


kfarr:
manual test steps all pass, including regression checks

PR #1921, opened 2026-08-17
fix(ai-tools): stop dropping valid segment fields in managedStreetCreate

kfarr:
## Summary

Discovered via MCP: a `managedStreetCreate` call with two `boundary` segments (`side: left` + `variant: brownstone`, `side: right` + `variant: arcade`) produced no buildings at all, with both boundaries stacked on the right edge.

**Root cause:** the handler's segment mapper whitelists fields when composing the json-blob and silently dropped `side` and `variant`, even though the tool schema documents both and the `managed-street` json-blob parser reads both (`managed-street.js:605-606`).

**Audit of the rest of the tool surface vs the street-* components** turned up more valid configurations blocked for MCP/AI callers:

| Gap | Effect |
|---|---|
| `slope`/`slopeStart`/`slopeEnd` dropped by mapper + undocumented | sloped (coastmix) segments impossible to create |
| `floors` dropped + undocumented | boundary building-height metadata lost |
| `generated.rail` missing from schema | rail/tram track configs blocked |
| `surface` advertises invalid `planting-strip`; omits valid `cracked-asphalt`, `parking-lot`, `water` | callers author untextured surfaces |
| `clones.mode` omits `fit` | edge-to-edge tiling mode undiscoverable |
| `pedestrians.density` omits `empty`/`sparse` | partial enum |
| `striping.striping` unenumerated | callers guess pattern names |

## Changes

- `nonCommandTools.js`: mapper passes `variant`, `side`, `floors`, `slope`/`slopeStart`/`slopeEnd` through to the json-blob (conditional spreads, matching the existing `generated` pattern); schema additions/corrections per the table above. `managedStreetUpdate` shares the segment schema so add/update-segment benefit too (its handler already passed segments through untouched).
- `defaultStreets.js`: the Planted Median sample used the same invalid `planting-strip` surface (no texture mapping exists); now `grass`.

## Test plan

- [ ] Pair an MCP session against this build and re-run the boundary street create: brownstone buildings on the left edge, arcade on the right, both facing the street
- [ ] Add Layer > default street with Planted Median renders grass texture instead of flat color
- [ ] eslint + prettier clean (verified locally)

### DIFF
--- a/src/editor/components/elements/AddLayerPanel/defaultStreets.js
+++ b/src/editor/components/elements/AddLayerPanel/defaultStreets.js
@@ -1092,7 +1092,7 @@ export const stroad150ftROW = {
     {
       name: 'Planted Median',
       type: 'divider',
-      surface: 'planting-strip',
+      surface: 'grass',
       color: '#338833',
       elevation: 0.15,
       width: 0.61, // Original 2ft
--- a/src/editor/lib/commands/nonCommandTools.js
+++ b/src/editor/lib/commands/nonCommandTools.js
@@ -35,6 +35,16 @@ async function managedStreetCreateHandler(args) {
       direction: segment.direction || 'none',
       color: segment.color || '#888888',
       surface: segment.surface || 'asphalt',
+      ...(segment.variant ? { variant: segment.variant } : {}),
+      ...(segment.side ? { side: segment.side } : {}),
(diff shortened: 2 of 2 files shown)


kfarr:
see #1925 instead

PR #1927, opened 2026-08-31
Claude/managed intersection prototype 0x7j4m

kfarr:


kfarr:
Closing: this branch was merged into #1974 (commit 576d5866), which carries the managed-intersection prototype forward together with the OSM street work.

PR #1985, opened 2026-09-11
Add meshopt simplify to GLB optimization, fix thumbnail capture transport

### LINKED ISSUE #1978: large meshy glb assets are not properly compressed
a user scene has meshy ai imported assets of buildings, each of which is 100mb+ and is not properly compressing leading to very long load times and challenges visualizing all the properties together

perhaps:
- consider destructive poly simplification / compression or even lod 
- consider server-side processing job for these assets

example scene: https://3dstreet.app/#/scenes/c4dbb034-e233-4faa-8c9f-ce9f1c28ec12

vincentfretin:
This fixes #1978

## Summary

Two independent changes to the user asset upload path, plus cleanup.

### Optimization pipeline

- **Add `simplify()`** (meshoptimizer, `ratio: 0`, `error: 0.001`) after `weld()`. It is the only lossy step in the pipeline: it decimates as far as the 0.1% error bound allows, so dense scans collapse hard while clean low-poly models come out untouched. Verified against the real pipeline: a 3M-triangle test sphere → 8.8k triangles, a 12-triangle cube → unchanged.
- Adds the `meshoptimizer` dependency, imported from its `/simplifier` subpath so the worker bundle skips the encoder, decoder, clusterizer and tangents. Its WASM is inlined as base64, so unlike `draco3dgltf` there is nothing to copy next to the worker bundle and no `locateFile` to fix up. It lands in lazy chunks (~47 KB), not the main bundle.
- **Raise the worker timeout 15s → 30s.** A real 113 MB photogrammetry GLB measured 9.4s on a fast machine, leaving too little headroom — a slower CPU or a cold dep/WASM load blew past 15s and silently fell back to the unoptimized original. The docblock now enumerates what the budget actually covers (dep chunks, WASM instantiation paid per upload since every call spawns a fresh Worker, parse, all 12 steps, writeBinary).

### Thumbnail capture

Fixes `[asset-upload] thumbnail capture failed Error: failed to fetch blob url: TypeError: Failed to fetch`, reported on a 113 MB GLB.

The screenshot iframe was handed a `blob:` URL to `fetch()`. That fetch fails under memory pressure — reproduced in Chrome at 11.6 MB, the exact size of the reporter's optimized blob, while XHR on the *same* URL succeeded. It is not the iframe boundary (same-document `fetch` fails identically) and not a blob-store quota.

The parent now **postMessages the Blob** instead. Structured-cloning a Blob passes a handle to the browser-process blob store rather than copying bytes, so it is also one full copy fewer than `fetch()` + `r.blob()`. Measured: 11.6 MB in 20ms, 113 MB in 144ms, where the URL form failed at both sizes. The iframe already supported this payload — it was sitting there as the legacy branch. The now-unreachable `{ url }` branch is removed.

Confirmed fixed by the reporter on the original 113 MB file.

### Cleanup

- Split the near-duplicate module JSDoc so the worker owns the pipeline and message protocol while the shim owns the timeout and fallback contract.
- Consolidate `GLB_MAGIC` into `glbMagic.js`. It was declared in three modules, each having grown its own slightly different short-buffer guard.
- Correct an out-of-date default in the thumbnail JSDoc (`timeout` documented as `30000`, actual `DEFAULT_TIMEOUT_MS` is `15000`). This is a doc-code sync fix in a different file, not a reduction of the worker timeout this PR raises.

## Test plan

- [x] `npm run dist` compiles (only pre-existing sass deprecation warnings); main bundle 4.12 MiB against the 4.25 MiB budget
- [x] `npm run test:modern` — 238 shared tests pass, including a new `captureThumbnail.test.js` pinning that we post the Blob and never call `createObjectURL`
- [x] `npm run lint` clean
- [x] Pipeline exercised end-to-end in Node against generated meshes (triangle counts and timings above)
- [x] Reporter confirmed a thumbnail is now produced for the 113 MB GLB that previously failed

## Notes for review

- `simplify()` is **lossy** and applies to every GLB upload — this is the first step in the pipeline that permanently discards geometry from the served asset. The original bytes are still uploaded and kept (`storageUrl`); only the served `optimizedSourceUrl` is decimated. There is no per-asset opt-out today.
- There is deliberately no size bail-out on thumbnail capture: if a model is too heavy to render, the existing capture timeout bails and tears down the iframe.

https://claude.ai/code/session_01AX7nmScX4dNpFbfh7b42TZ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "jspdf": "^4.2.1",
         "jszip": "^3.10.1",
         "lodash-es": "^4.17.23",
+        "meshoptimizer": "^1.2.0",
         "nanoid": "^5.1.6",
         "pbf": "^5.1.2",
         "posthog-js": "^1.356.1",
@@ -19660,6 +19661,12 @@
       "engines": {
         "node": ">= 8"
       }
+    },
+    "node_modules/meshoptimizer": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/meshoptimizer/-/meshoptimizer-1.2.0.tgz",
+      "integrity": "sha512-davRZeIJbxJrE24cwQle7ZDsxjdk/OphNOV83oX+efQinyoHY9Jcyz3MHbaoG0qySZajldGztNZ1RN/T19PZsg==",
+      "license": "MIT"
     },
     "node_modules/methods": {
       "version": "1.1.2",
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "jspdf": "^4.2.1",
     "jszip": "^3.10.1",
     "lodash-es": "^4.17.23",
+    "meshoptimizer": "^1.2.0",
     "nanoid": "^5.1.6",
     "pbf": "^5.1.2",
     "posthog-js": "^1.356.1",
--- a/public/model-viewer-screenshot.html
+++ b/public/model-viewer-screenshot.html
@@ -34,6 +34,7 @@
          resulting blob as the asset thumbnail. Same stabilisation pattern
          as @shopify/screenshot-glb: poster-dismissed → 3× rAF →
          modelViewer.toBlob(). -->
+
     <model-viewer
       id="mv"
       loading="eager"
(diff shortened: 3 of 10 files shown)


vincentfretin:
Thanks — pushed a514d536 addressing the test findings, plus the two clarifications.

**Fixed**

- **Unhandled rejections.** Correct, and the sharper version of the problem is that the pending promises kept the 5s ready timer and 15s capture timer alive past the end of the test. Rather than swallow with `.catch(() => {})` I gave the tests a `startCapture()` helper that stubs `contentWindow` and returns a `send()`, so every test now drives the promise to settlement — `cleanup()` clears both timers, so nothing outlives the test.
- **Happy path.** Added: dispatching `3dstreet:screenshot-blob` resolves with the JPEG and the iframe is removed from the DOM. Also added the error path (`screenshot-error` rejects), since it was one line once the helper existed.
- **`event.source` on the iframe handler.** Added the check plus a comment stating the trust assumption — worth being explicit that it is a sanity check rather than a security boundary: the page renders whatever GLB it is handed and posts the thumbnail back to that same window, so the worst a stray sender achieves is a screenshot of its own model.
- **JSDoc `timeout` 30000 → 15000.** Fair point that it reads oddly in this PR. Called out in the description now as a doc-code sync fix in a different file.

**One correction**

> all three call sites now consistently treat a <4-byte buffer as "not a GLB" (was "throws" in the DataView paths previously by way of `getUint32` on OOB)

No behavior change — all three were already guarded before this PR:

- `extractGlbAttribution.js`: `const magic = view.byteLength >= 4 ? view.getUint32(0, true) : 0;`
- `analyzeGltf.js`: `if (view.byteLength >= 4 && view.getUint32(0, true) === GLB_MAGIC)`
- `optimizeGlb.worker.js`: `if (bytes.byteLength < 4) return false;`

None could throw on a short buffer. The duplication being removed is three *equivalent* guards, which is exactly why consolidating them is low-risk. Nothing to note in the commit.

**Acknowledged, not changed here**

- **Lossy simplify with no opt-out.** Agreed on hard-edge normals and CAD-derived parts; `ratio: 0` with the 0.1% bound is deliberately aggressive. `storageUrl` keeps the original bytes, so nothing is unrecoverable, but a per-upload "don't decimate" toggle is the real answer if artifacts show up in practice. Worth a follow-up issue rather than widening this PR.
- **Per-upload WASM instantiation.** Real, and the 30s budget absorbs it today. A persistent optimize worker is the fix if batch upload ever lands; noted in the `DEFAULT_TIMEOUT_MS` docblock so the cost is visible to whoever tunes it next.

https://claude.ai/code/session_01AX7nmScX4dNpFbfh7b42TZ

kfarr:
Superseded: folded into #1986, which is now targeted at main and includes these commits unchanged.

PR #1986, opened 2026-09-11
Add a Reoptimize action to the asset details modal

vincentfretin:
> **Stacked on #1985** — base branch is `feat/glb-simplify-and-thumbnail-transport`, so this diff shows only the reoptimize commit. Merge #1985 first; this retargets to `main` automatically.

## Summary

Re-runs the current client optimization pipeline over an asset already in the gallery and repoints its doc at the result. Today an asset optimized by an older pipeline — or one whose optimization was skipped entirely because the worker timed out, leaving it serving the raw original — can only be improved by deleting and re-uploading it. With the lossy `simplify()` step landing in #1985, that gap is about to matter for every existing asset.

The control is an inline button in the Size row, next to the text that explains why you'd press it:

```
Size: 121.2 MB (Optimization skipped) [Retry optimization]
Size: 121.2 MB → 11.6 MB (−90%) [Reoptimize]
Size: 121.2 MB [Optimize]                  ← uploaded before the pipeline existed
```

Capture d’écran du 2026-09-11 17-04-01

It is disabled while running, with progress ("Optimizing…"), the outcome, and any failure on their own line directly beneath. Owner-only and GLB-only — splats are transcoded server-side by the RAD job instead. All strings go through `sharedMessages` (en/es/pt-BR/fr), since this modal also renders in the generator and bollardbuddy islands where no `IntlProvider` is mounted.

## How the file swap works

The original (`storagePath` / `storageUrl`) is never touched: it stays the quota-counted file and the escape hatch if an optimization ever makes a model worse. Only the optimized variant is replaced.

The new GLB goes to a **new** path and the doc is then repointed, rather than overwriting in place — so the doc never points at a half-written object, and a failed upload leaves the existing variant serving. That deliberately orphans the previous object, which is what `cleanupOrphanedStorage` already collects: it builds its referenced set from `storagePath` + `optimizedSourcePath` + `thumbnailPath` across every asset doc (soft-deleted included) and deletes anything under `users/*/assets/` that nothing references and that is older than its 24h grace window. Clients cannot delete Storage objects, so this is the only disposal path. Worst case the stale object lingers until the next monthly run; it costs storage but never quota, which counts `size` alone. I added that dependency to the job's docblock, since reoptimize turns orphans from "something went wrong" into a routine occurrence.

If the upload succeeds but the doc update fails, the new file is protected by the grace window and then collected — a partial failure leaks nothing permanently.

## Security rules

`optimizedSourcePath` and `optimizedSourceSize` are no longer frozen after create. Neither can spoof quota: the tally reads `size` alone, and `asset-quota.js:119` explicitly excludes `optimizedSourceSize` as a platform-derived artifact. `size`, `storagePath` and `userId` stay frozen.

`optimizedSourcePath` remains constrained to the owner's own folder, mirroring the create rule, so a client can't repoint a doc at another user's bytes and keep them alive. It is validated **only when actually changed**, so an untouched legacy value can never block an unrelated update such as a soft delete.

## ⚠️ Not fully tested — needs @kfarr

I don't have permission to deploy Firebase config, so two things block end-to-end verification:

1. **The rules change must be deployed** before the button can work at all — the doc update is rejected by the currently deployed rules.
2. **Bucket CORS blocks the download step.** `fetch()` on the asset's `storageUrl` fails with `net::ERR_FAILED` against a `200 (OK)` — the browser refusing the cross-origin read. `public/cors.json` already lists the right origins but has to be applied per bucket by hand (same manual step documented in `docs/rad-cloud-run-pipeline.md:138`):

   ```
   gsutil cors get gs://dev-3dstreet.appspot.com
   gsutil cors set public/cors.json gs://dev-3dstreet.appspot.com
   ```

   Worth noting this is not specific to this PR: any JS-side read of user GLBs from that bucket is blocked the same way, including `GLTFLoader` scene loads and the model-viewer preview. The Download button is unaffected because `window.open()` is a navigation, and CORS doesn't apply to navigations.

3. **The rules tests were written but not run** — `npm run test:rules` needs JDK 21+ and this machine has only 8 and 11. Please run it before deploying the rules.

## Test plan

- [x] `npm run test:modern` — 258 shared tests, 11 of them new for `reoptimizeAsset` (new-path upload, artifact tagging, the three repointed fields, original untouched, skip/no-win outcomes, stage reporting, download failure, abort)
- [x] `npm run test:components` — 84 pass
- [x] New `sharedMessages` test: every message id has a string in all four locales (that table is hand-maintained and outside the formatjs extraction pipeline, so nothing else caught a missing translation)
- [x] `npm run dist` clean, `npm run lint` clean
- [ ] **Rules tests** — written, not run (no JDK 21 locally)
- [ ] **End-to-end in the browser** — blocked on the rules deploy + CORS above

https://claude.ai/code/session_01AX7nmScX4dNpFbfh7b42TZ


vincentfretin:
Pushed 2ab9be34 — all four fixed, including the race.

**1. Cross-asset navigation race.** Confirmed and fixed. I took the AbortController version you preferred: `reoptimizeAsset` already accepts `signal` and plumbs it to both `fetch` and `uploadToStorage`, so navigating away now actually cancels the download/upload instead of letting it run to completion and discarding the result. On top of that the effect keyed on `[assetId, ownerUid]` clears the three status states, and stale stage reports and results are dropped — the abort rejects the promise, but the window between the last `await` and the state write isn't covered by that, so both guards are needed.

Covered by a new `MeshDetailsModal.reoptimize.test.jsx` that renders the real modal (the existing component tests define mock components inline, so this is the first one to mount it). Two cases: A's late completion must not touch B, and A's in-progress status must not appear under B. **I verified both fail without the fix** — `expected false to be true` on the abort assertion and `expected <div> to be null` on the leaked status — and pass with it.

**2. Abort as a "no change" outcome.** Good catch, and it stopped being latent the moment I wired the signal in (1) — so this would have shipped as a visible "No change (aborted)." `reoptimizeAsset` now re-raises `reason === 'aborted'` as a `DOMException('…', 'AbortError')`, which the modal already drops silently. Test added.

**3. `sharedMessages` substring false positive.** Right — and worth spelling out why it mattered: `chunk.includes('es:')` matches a body like `'notes: see …'`, so a missing `es` key would have been masked. Replaced with a line-start regex that handles both shapes (`en:` bare, `'pt-BR':` quoted), and I checked it flags a deliberately removed key rather than just passing.

**4. `..` in the optimizedSourcePath regex.** Tightened with `!path.matches('.*/[.][.](/.*)?')` rather than an extension allowlist — the pipeline emits `.glb` today, but legacy optimized paths carry the original upload's extension, and an allowlist would be a trap for whoever adds the next variant format. I ran the pattern under full-match semantics against both the paths we actually write and the traversal forms; a legitimate filename with a dot run (`my..model.glb`) is unaffected because the `..` isn't a path segment. Two rules tests added.

⚠️ The rules tests still **have not been run** — `npm run test:rules` needs JDK 21+ and this machine has 8 and 11 only. That now covers six cases in this file, including the two new ones. Whoever deploys the rules should run it first; the regex reasoning above is careful but unverified against the real evaluator.

Verified otherwise: 261 shared tests pass, `npm run lint` and `npm run dist` clean.

https://claude.ai/code/session_01AX7nmScX4dNpFbfh7b42TZ

kfarr:
Retargeted to `main` and consolidated into the single PR for manual testing and review. This branch now carries:

- **#1985** meshopt simplify + thumbnail transport fix (already stacked underneath)
- **#1986** Reoptimize action (this PR), plus follow-ups: 90s manual worker budget, **Remove optimized** (reversal of a lossy optimization; rules allow dropping `optimizedSourcePath`, null still denied, emulator tests added), background runs stay visible when the modal is closed and reopened (module-level run registry), Download Original / Download Optimized labels
- **#1991** unified upload stage copy across the pending card and properties panel, byte-weighted progress across original + optimized uploads, scene-graph upload dot as an icon corner badge for attention states only

Also merged `main` in. `npm run test:rules` (JDK 21) passes 46, `npm run test:modern` green, lint and `npm run dist` clean. Rules are deployed to dev-3dstreet.

Closing #1985 and #1991 as superseded by this PR.

PR #1991, opened 2026-09-12
Unify upload stage copy across pending card and properties panel

### LINKED ISSUE #1989: Upload status copy is out of sync between the pending card and the properties panel
Follow-up from #1986 review. Cosmetic, but it makes large GLB uploads look stuck.
  
  ### What happens today
  
  The upload pipeline (`src/shared/asset-upload/uploadAsset.js`) reports four stages: `validating`, `optimizing`, `uploading`, `thumbnailing`. Two surfaces render them differently:
  
  - **Pending card in the assets panel** (`PendingUploadCard.jsx`) shows "Uploading…" and then "Finishing…" for `thumbnailing`.
  - **Properties panel dot / AssetInfoPanel** (`useAssetUploadStatus.js`, `AssetUploadDot.jsx`) shows "Uploading NN%" and has no label for `thumbnailing`, so it keeps saying "Uploading 100%" while the card says
  "Finishing…".
  
  Three specific problems:
  
  1. **"Uploading 100%" stalls.** `assetsService.addAsset()` uploads the original, then the optimized GLB, then (for images) a thumbnail. The `uploadProgress` event is only emitted for the original
  (`assetsService.js` ~L220). So for a 100 MB GLB the properties panel reaches "Uploading 100%" and then sits there for many seconds while the optimized variant uploads with no feedback.
  2. **"Uploading" before anything uploads.** `validating` and `optimizing` both render under the generic "Uploading" affordance in the assets panel button, and the optimize step alone can take up to 30s on a
  large model. It should read "Preparing…" or similar until bytes are actually moving. 
  3. **Which file?** When the status does say Uploading, it doesn't say whether that's the original, the optimized variant, or the thumbnail. With #1985 raising the optimize budget and #1986 adding manual
  reoptimize, users are going to be watching these states more.
  
  ### Proposed
  
  - One shared stage→label table (probably in `sharedMessages` since the pending card renders in generator/bollardbuddy too) used by both the card and the properties panel, so they always agree.
  - Labels roughly: Validating… / Optimizing… / Uploading original NN% / Uploading optimized… / Finishing… (thumbnail).
  - Emit `uploadProgress` for the optimized upload too (or split progress into per-file segments), so 100% means done.
  - Keep "Finishing…" as the terminal state on both surfaces rather than "Uploading 100%".
  
  ### Out of scope
  
  Reoptimize progress in the details modal already has its own stage line and is covered by #1986.

kfarr:
Fixes #1989.

## Summary

Large GLB uploads looked stuck because the two upload surfaces kept their own status tables and drifted: the properties panel said "Uploading 100%" while the gallery's pending card said "Finishing…".

- **Byte-weighted progress.** `addAsset()` uploads the original and the optimized GLB in parallel behind one progress bar, but only the original reported progress. The bar now aggregates both by bytes (`createAggregateProgress`), so 100% means every byte of every file is up. The `uploadProgress` event contract is unchanged.
- **One stage→label table.** `shared/assets/uploadStageLabels.js` (strings in `sharedMessages`, en/es/pt-BR/fr, since the pending card also renders in the generator and bollardbuddy islands) drives the pending card, the assets panel Upload button, the properties panel and the scene-graph dot. The percentage rides inside the label ("Uploading 42%"), so both surfaces read identically.
- **Nothing says Uploading before bytes move.** The editor's per-entity slot starts as `validating` ("Preparing…") through quota preflight and the model-load gate. The Upload button shows the current stage (without a ticking percentage) instead of a blanket "Uploading…".
- **Finishing on both surfaces.** After the asset doc is written, the editor slot flips to `finishing`, matching the card, instead of lingering on "Uploading 100%" through the GLB preload and entity swap.

## Deviation from the issue

The issue proposed "Uploading original NN%" and "Uploading optimized…" as separate labels. Since the two files upload in parallel, a single byte-weighted percentage is the honest number; naming the file would imply a sequence that doesn't exist.

## Test plan

- [x] `npm run test:modern` — 1128 pass, incl. new `uploadProgress.test.js` and `uploadStageLabels.test.js`; the existing `sharedMessages` test enforces all four locales for the new ids
- [x] `npm run lint` clean, `npm run dist` clean
- [ ] Browser: drop a ~100 MB GLB in the editor and watch the properties panel and the pending card say the same stage throughout; confirm the bar doesn't hit 100% until the optimized upload finishes; confirm the Upload button reads Preparing…/Optimizing… before Uploading

https://claude.ai/code/session_01ANAecAtrr1TfnLEYNYmiYL

### DIFF
--- a/src/editor/components/elements/AssetInfoPanel.jsx
+++ b/src/editor/components/elements/AssetInfoPanel.jsx
@@ -4,8 +4,11 @@ import PropTypes from 'prop-types';
 import posthog from 'posthog-js';
 import useAssetUploadStatus, {
   STATUS_LABELS,
-  REASON_TEXT
+  REASON_TEXT,
+  CANCELLABLE_STATUSES,
+  getStatusText
 } from './useAssetUploadStatus';
+import { useSharedMessages } from '@shared/i18n/sharedMessages.js';
 import useAssetUploadStore from '@/editor/state/assetUploadStore.js';
 import useCurrentUploadStore from '@shared/assets/state/currentUploadStore.js';
 import { uploadAndPlaceAsset } from '@/editor/lib/asset-upload/uploadAndPlaceAsset.js';
@@ -29,6 +32,7 @@ const formatStorage = () =>
 
 const AssetInfoPanel = ({ entity }) => {
   const intl = useIntl();
+  const t = useSharedMessages();
   const state = useAssetUploadStatus(entity);
   const [detailsOpen, setDetailsOpen] = useState(false);
   if (!state) return null;
@@ -61,10 +65,10 @@ const AssetInfoPanel = ({ entity }) => {
     useStore.getState().startCheckout('storage');
   };
 
+  // The percentage rides inside the stage label ("Uploading 42%") so this
+  // panel and the gallery's pending card read identically.
   let detail = '';
-  if (state.status === 'uploading' && state.progress > 0) {
-    detail = `${state.progress}%`;
-  } else if (
+  if (
     (state.status === 'uploaded' ||
       state.status === 'local' ||
       state.status === 'local_error') &&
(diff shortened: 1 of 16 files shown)


kfarr:
Added a second commit: the scene-graph upload dot now pins to the upper right of the entity icon and only renders for attention states (amber in-flight, red failed/missing). Synced assets and plain local previews show no dot.

kfarr:
Superseded: merged into #1986 for combined manual testing and review.

Issue #2004, opened 2026-09-15
OSM cross-section fidelity: strassenraumkarte-grade tag coverage


kfarr:
## Context

We have a collaboration opportunity with the OSM Berlin group behind [osmberlin/strassenraumkarte](https://github.com/osmberlin/strassenraumkarte) — the Berlin street-space "micromap" that renders micromapped OSM at decimeter accuracy. Their August 2026 rewrite is a generalizable osm2pgsql + PostGIS pipeline (Apache-2.0, so portable into this AGPL codebase with attribution).

Key finding from the deep dive: their per-way **lane model** (`data/lua/lanes.lua` — ordered lanes with type, direction, width, signed offset, markings per edge) is structurally the same object as our managed-street segment array. They consume ~416 OSM tags; our Overpass hydrator (`src/tested/osm-way-tags.js`, shipped in #1974) reads about a dozen. This issue tracks closing that gap — the "flesh" track, parallel to #1930's architecture ("skeleton") track.

## Scope

All work lands primarily in the pure, unit-tested modules: `src/tested/osm-way-tags.js` (`crossSectionFromTags`), `src/tested/osm-street-import.js` (segment builders / class rules), with UI provenance via the existing `facts` record.

### Phase A — tag parity in the hydrator (no dependencies, start now)
- [x] `width` / `width:carriageway` → derive real drive-lane widths (carriageway minus parking/bike lanes, per strassenraumkarte's `width:effective` approach); record `mapped` vs `derived` provenance in `facts` *(PR #2006)*
- [x] `width:lanes(:forward/backward)` per-lane widths *(PR #2006)*
- [x] Street-parking schema (post-2022, authored by the Berlin group): `parking:left/right/both` + `:orientation` → parallel / diagonal / perpendicular parking segments at orientation widths (lanes.lua defaults: 2.0 / 4.5 / 5.0 m); `:restriction` / `no` suppresses the lane *(PR #2006)*
- [x] `cycleway:*:width`; `cycleway=track` or `cycleway:*:separation:*` → protected bike lane (buffer/divider segment) vs painted lane *(PR #2006)*
- [x] `turn:lanes(:forward/backward)` → per-lane stencil arrows (`left`, `right`, `straight`, `left-straight`, `right-straight`, `both`, `all` mixins already exist) *(PR #2006)*
[...]

Issue #1930, opened 2026-08-31
Street node graph: owned centerlines, non-destructive connections, OSM import + LOD hydration


kfarr:
## Why

The managed-intersection prototype (`claude/managed-intersection-prototype-0x7j4m`, `docs/managed-intersection.md`) proved that intersections can be derived automatically from street endpoint "nodes" — and surfaced the architectural ceiling of doing it without a real node model:

- Street nodes are **derived state** (`position` + `rotation.y` + `length` + `street-align`), re-derived independently by the endpoint gizmo and the intersection. There is no persistent node identity two streets can *share*.
- Keeping streets flush at a junction currently **mutates street data** (the snap pass rewrites `position`/`length`). It converges, but there's no undo step, no restore when an intersection is deleted, and edit→save→reload round-trips carry the mutated values as truth.
- Curved (path-following) streets store their centerline in a **separate referenced shape entity**, so they have no straight-node ends and were deferred from intersection support entirely.

Each of these is the same missing piece. This ticket is the umbrella for building it once, in a way that also unlocks OSM import and area-scale rendering.

## Architecture (three pillars)

**1. Street-owned centerlines with shared end nodes.**
A managed street owns its centerline inline: an ordered point list whose first/last points are **nodes** — stable identities that streets and intersections can share. A straight street is the degenerate 2-point case; today's curved-street `path` reference demotes from storage to authoring tool ("make street from drawn path" *copies* geometry in). This is exactly OSM's data model (way = node, geometry, node), so import becomes translation, not adaptation. The endpoint gizmo generalizes to "drag a centerline vertex", with end vertices being connection nodes — one interaction model for straight and curved.

**2. Non-destructive connections via render-time insets.**
Stored street geometry ("A to B, 60m") becomes immutable under connection. An intersection occupying a node writes a per-end **inset** that the layout cascade and every generator (segments, striping, clones, ground, labels) respect at render time. Deleting an intersection trivially restores full streets; undo history stays clean; saved scenes carry user intent, not derived trims. This replaces the prototype's snap-mutation (which stays as the shipped interim behavior until this lands).

**3. Intersections as node occupants.**
[...]

PR #2006, opened 2026-09-15
OSM streets: strassenraumkarte-parity tag coverage + generation continuity (#2004)

kfarr:
Two related improvements to OSM street generation, combined for one QA pass:

## Part 1 — tag coverage (first slice of #2004)

The Overpass hydrator (`crossSectionFromTags`) now consumes the highest-leverage tags from the [osmberlin/strassenraumkarte](https://github.com/osmberlin/strassenraumkarte) lane model (`data/lua/lanes.lua`, Apache-2.0 — width defaults and parked-car constants ported with attribution). Everything lands in the pure, unit-tested modules; the `facts` provenance chain is unchanged in shape.

- **Widths** — `width` / `width:carriageway`: drive-lane widths derived as carriageway minus on-carriageway extras (parking lanes, painted bike lanes), clamped 2–4.5 m (the strassenraumkarte `width:effective` derivation). Per-lane `width:lanes(:forward/backward)` values win where tagged.
- **Street parking, old and new schema** — orientation (parallel / diagonal / perpendicular → 2.2 / 4.5 / 5.0 m lanes) with orientation-specific parked-car spacing (5.2 / 3.1 / 2.5 m, the Neukölln-validated constants) and angled clone facing (0° / 55° / 90°); `parking:*:width` overrides; `parking:*:restriction` suppresses the lane.
- **Cycleways** — `cycleway:*:width`; `track` / `:separation` / `:buffer` → protected lane with a raised buffer divider (protected tracks also keep their own surface when `surface=*` repaints the carriageway).
- **Turn lanes** — `turn:lanes(:forward/backward)` → per-lane stencil arrows (existing atlas mixins), applied to bus lanes too; driver-leftmost-first entries mapped onto our curbside-first segment order; unknown movements draw nothing rather than a wrong arrow.
- **Bus/PSV lanes inside ordinary roads** — `bus:lanes*` / `psv:lanes*` designated positions, or count-only `lanes:bus*` placed curbside per direction.
- **Surface** (carriageway segments only) and per-side **`sidewalk:*:width`**.
- **Rail & transit class presets** — these fell through the unknown-class fallback to the *residential* preset (a railway generated as a 2-way road with parked cars). Now: `rail` → sloped gravel berms flanking a 1 ft ballast bed with a 1435 mm track (dimensions from the reference scene); `transit` (tram/light_rail) → single flush track on concrete; subclass `subway` excluded from ribbons/generate; honest ribbon widths and a non-road tint; hydration guard so a nearby road's `highway` tags can never repaint a railway.

Chip line example: `Karl-Marx-Straße · 2 lanes · width 11 m · 1 bus lane · sidewalks both sides · parking both sides (diagonal) · protected bike lane right · sett surface` — unmapped fields still flagged as assumed.

## Part 2 — generation continuity (the "disjointed streets" fixes)

Generated streets read disjointed: junction cuts ate 22+ m of street, sliver-side cuts minted no intersection, side-road Ts chopped the through street, and the one-generate-per-way gate permanently truncated long ways at the ±200 m window.

- **Carriageway-width junction cuts** — cuts clear the crossing street's carriageway (`importedCarriagewayMeters`, sidewalks excluded) with a 2 m pad, instead of half the sidewalk-to-sidewalk width + 4 m. A residential crossing's gap shrinks from ~22 m to ~14.4 m.
- **T junctions no longer cut the through street** — `junctionsAlongStretch` classifies each junction `crossing` vs `terminal` (does the other way continue past the stretch, or end there?); only crossings split. A side street whose end terminates on another road trims back to that road's carriageway edge (`trimStretchEndsAtWays`, half carriageway + 2 m) so it butts flush instead of poking to the centerline; end-to-end touches (tile fragments of the same road) are left alone.
- **Sliver-cut pads** — intersections mint at any cut bordered by ≥1 generated street end (was ≥2), so a cut whose far side dropped below the minimum piece length gets a pad instead of a bare hole; later generates of the crossing way still reuse it by proximity.
- **Partial-intersection pads** — a `managed-intersection` with exactly one connected arm no longer renders the floating 6 m disc: it draws that arm's roadway as an asphalt stub reaching half a roadway past the junction center, with the arm's crosswalk band at the mouth — an "unfinished intersection" waiting for its cross street. The disc remains for the zero-arm hand-authoring case; full polygon geometry, snapping and `lastGeometry` (plan exporter) still require ≥2 arms.
- **Extendable ways** — every generated street carries a `data-osm-stretch` coverage stamp (compact centerline encoding, survives save/reload). Clicking the same way again clips the new window against stamped coverage (`clipStretchToUncovered`: resample, longest uncovered run, boundary snapped flush to the covered piece's endpoint) and **extends** the street instead of refusing `already-generated`. Pre-stamp scenes keep the old whole-way refusal rather than double-generating.
- **Semitransparent ribbons** — the 2.5D ribbons render at 0.65 × the layer opacity so the basemap's baked street-name labels stay readable beneath them; a `LessDepth` depth test keeps the builder's same-height cap/joint overlaps from double-blending, and hover/selection highlights stay fully vivid.

## Changes

- `src/tested/osm-street-import.js`: `PARKING_BY_ORIENTATION`, orientation-aware `parking()` builder, `bikeBuffer()` divider, rail/transit presets (`railBerm`/`railTrack`), `importedCarriagewayMeters`, junction `kind` classification, `trimStretchEndsAtWays`, `clipStretchToUncovered`, coverage option on `stretchForWindow`, encode/decode stretch stamps, inset pad 4→2
- `src/tested/osm-way-tags.js`: width parsing, parking detail (both schemas), cycleway detail, `turn:lanes` mapping, bus-lane designation, surface map, per-side sidewalk widths; extended `facts` + `describeFacts`
- `src/tested/osm-street-style.js`: rail/transit widths + tints, `EXCLUDED_TRANSIT_SUBCLASSES`
- `src/aframe-components/osm-streets.js`: coverage stamps + coverage-aware gating, crossings-only splitting + end trimming in `planUpgrade`, ≥1 intersection minting, `upgradeNearFocus` extends partial ways, subway filter, rail hydration guard, semitransparent ribbon material
- `src/aframe-components/managed-intersection.js`: `buildPartialPlaceholder` for the one-arm case
- Tests: 37 new mocha cases across the pure suites; 4 new browser component cases (extend-beyond-window with flush butting; sliver-cut minting with partial-pad + crosswalk assertions; T-junction continuity + side-street trim; ribbon material transparency)
- Docs: `docs/osm-street-lod.md` rewritten for crossings-only splitting, coverage stamps, carriageway insets, partial pads, rail presets and ribbon transparency; `docs/managed-intersection.md` documents the one-arm partial placeholder

## Not in this PR (tracked in #2004 / #1930)

Phase B (`placement`, `dual_carriageway`, `area:highway`, street-running transit merge), Phase C (crossing/signal nodes, roundabouts), trees/street furniture, `shoulder`; flush pathed-arm snapping at intersections rides #1930 phase 5.

## Validation

- `npm run test:core`: 407 passing · `npm run test:modern`: 1190 passing · `npm run test:components`: 92 passing (real browser)
- `npm run lint`, prettier: clean

https://claude.ai/code/session_016GK9DHtEA8t2PZKtRWWaei


kfarr:
Review addressed in 3d74e70:

- **Bus lanes + turn arrows**: stencils now apply to designated bus lanes too (appended alongside the BUS clone config), locked in by a test.
- **Protected tracks vs `surface`**: `surface=*` no longer repaints a protected bike track — it sits off the carriageway behind its buffer and keeps the bike-lane default; painted lanes still inherit. Test added.
- **Naming**: builder renamed to `bikeBuffer` (still exported as `segmentBuilders.buffer`).
- **`describeFacts`**: singular "bike lane" for a single tagged side.
- **Comments** added for the permissive directional-separation read, the deliberate no-arrow fallback for `merge_to_*`/`reverse`, and why bus lanes skip the derived carriageway width.
- **Tests**: all four suggested gaps covered (left-side track ordering, `oneway=-1` turn:lanes mapping, `psv:lanes` alias, arrows on bus lanes). 388 core + 1190 vitest passing, lint clean.

The `sides()` note you self-retracted is indeed fine as-is (later per-side branches override the `both` value).

---

Issue #1993, opened 2026-09-12
consider rotation center axis to be currently selected object


kfarr:
as a user who is not free flying around a scene and instead who has an object selected, it may make more sense to have the rotation centerpoint be about the pivot point of the selected object;

as a user who wants to "get free" of this mode then i should press escape or click outside of the currently selected object

issues:
- how to handle cases where the selected object is out of the field of view (that's easy, just fall back to the current rotation scheme)
- how to handle cases where the selected object is far from the center of the field of view but not outside the field of view -- that is a tough one, i don't know

PR #2007, opened 2026-09-17
Add combined move and rotate gizmo

kfarr:
Upstream submission of Diarmid's fork PR https://github.com/diarmidmackenzie/3dstreet/pull/36 (branch `task-200-easy-gizmo-resume`), merged with current `origin/main` (64 commits behind at the time, merged cleanly with no conflicts). Lint, the full Mocha/Vitest suite (1,376 + 365 + 89 tests) and the browser component suite pass on the merged branch.

Original PR description follows.

---

Adds an opt-in combined move/rotate tool behind `?easygizmo=on`. Dragging follows continuous eligible surfaces, holds height across large drops and offers explicit landing targets. Near-level views flatten the controls and use horizontal movement with yaw rotation.

Placement uses a strict hierarchy: streets, buildings, imported meshes, then furniture. Items can rest only on item classes above them. Google 3D Tiles remain eligible terrain for every class, including embedded rooftops. Streets align their road-level reference with terrain; other items retain bounding-box-base alignment. The selected subtree's flattening is temporarily suspended while easy mode is active, with regeneration gating and restoration on detach, disposal or editor close.

Each catalog building contributes only its highest intersected surface per placement column, for following and both landing directions. Separate stacked models retain successive roofs. Lower terraces beneath higher geometry within one catalog model are excluded; maintainer review is requested if existing users rely on them. Imports and tiles retain their surface policy.

Batched-model intersections resolve the original per-instance entity, so furniture retains upper-floor and roof landings on duplicated catalog buildings. Splat selection bounds reuse caller-owned output and corner scratch while recomputing current streamed data. THREE/BVH raycast result allocations remain.

The toolbar retains existing controls until the lazy-loaded controller is ready. Gesture ownership, cancellation, undo boundaries, camera changes and managed-segment width-only routing are integrated with the current editor.

This PR targets the fork for live testing before upstream submission. The feature remains off by default. Native pointer capture, rendered affordances and streamed-scene performance still need live testing.

Validation:

- Modern suite: 1,304 tests across 84 files. Browser component suite: 85 tests across 13 files.

- Real raycast and gesture regressions cover the hierarchy, allowed landings, sub-step overlaps, scaled street road alignment and terrain following.

- The installed TileFlatteningPlugin fixture covers selected contribution removal, delayed regeneration and disposal release; browser tests cover suspension ownership and unchanged saved settings.

- The earlier complete-scene browser repro established batched floor/roof ownership; the new roof-only policy is covered by real raycast and gesture regressions, including successive stacked roofs and shared BatchedMesh instances.
- Six additional isolated mutations detect skipped roof filtering, global and mesh-host collapse, imported-mesh restriction, stale roof maps and above-only filtering. Each executes all 99 targeted tests with zero runner Errors.

- Four new isolated mutations detect recurring constructors, clones, omitted caller-owned bounds and omitted batched owner lookup; each executes all 107 targeted tests with zero runner Errors.

- Eleven earlier deliberate placement defects are caught by independent mutation checks in a disposable copy.

- ESLint, Prettier and unchanged upstream pre-commit checks pass. Translation catalogs are current in es, pt-BR and fr at 867 keys each.

- Production webpack build passes; the core bundle is 4,402,311 bytes, within the 4.25 MiB limit. Six existing Sass deprecation warnings remain.

Implementation rationale: `docs/easy-gizmo.md`.

-----------

# Easy transform gizmo: key decisions

## Core gizmo

- **C1:** Easy mode is limited to horizontal movement and yaw rotation; exclude pitch, roll, scale and free vertical dragging.
- **C2:** Small and large objects have the same handle size at the same viewing depth, superseding object-footprint-based sizing.
- **C3:** Size controls using the geometric mean of world-size and screen-size behavior, within minimum and maximum sizes, balancing usability at different zoom levels.
- **C4:** Use a yellow movement square with directional arrowheads, placed at the object’s base and turning with its heading.
- **C5:** Use a cyan open rotation ring whose gap marks heading, retaining its open shape and arrowheads during rotation.
- **C6:** Draw translucent controls over occluding geometry so they remain visible when the selected object is obscured.
- **C7:** Communicate rest, hover, dimming and pending action through opacity, keeping the same colors and remaining translucent.
- **C8:** In the looked-down-at view, rotate continuously by cursor bearing around the object, supporting turns beyond 180° without snapping back.

## Movement over irregular terrain

- **M1:** For grounded items, follow gradual changes in eligible support surfaces automatically.
- **M2:** Follow slopes up to 60°, independently of navigation’s floor/wall threshold; steeper slopes behave like walls.
- **M3:** Treat upward and downward steps up to 0.35 m as continuous, covering kerbs and shipped street elevation changes.
- **M4:** For elevated items with eligible support beneath them, preserve the object's initial clearance above that support while following continuous surfaces.
- **M5:** At larger steps, drops and steep faces, hold the object's current height and offer explicit landings instead of automatically changing levels; allow it to pass into buildings and steep faces as with the classic gizmo.
- **M6:** Decide support beneath the object’s origin, accepting partial overhang or intersection rather than introducing footprint fitting.
- **M7:** Sample movement paths at regular intervals with a bounded per-frame budget; hold height when that budget is exceeded, preventing a fast mouse move from treating a sudden roof-height jump as a gentle slope over a long distance.
- **M8:** Reuse navigation's positive surface categories, extended locally to admit imported meshes for placement; leave navigation behavior unchanged.

## Vertical movement

- **V1:** Offer explicit landing targets on eligible surfaces above and below, whether the object is floating or already supported; choose the nearest above, and below prefer eligible street/building/imported surfaces over Google 3D Tiles, choosing the nearest within those groups.
- **V2:** Show a landing target beyond a 10 cm gap and remove it below 5 cm; between 5 and 10 cm retain its previous visibility, avoiding flicker. Gaps below 5 cm are treated as placement tolerance and offer no explicit landing correction.
- **V3:** Update landing targets during movement.
- **V4:** Landing changes the object's height while preserving its horizontal position and yaw.
- **V5:** A landing square behaves like a button: the initiating pointer must press and release over the same target to trigger it, with highlight tracking departure and return.
- **V6:** Depict landing direction with noninteractive chevron stacks, animating toward the target on hover.
- **V7:** Terrain provides support for every item class; rank items as streets, buildings, imported meshes and other furniture, with items able to land only on earlier classes (e.g. buildings on streets, never streets on buildings).
- **V8:** Keep imported meshes in their own class even when they depict buildings; imported images and splats provide no support, and furniture provides no support for other items.
- **V9:** Keep Google 3D Tiles eligible as terrain for every class, accepting that embedded building rooftops cannot be distinguished from terrain.
- **V10:** Treat each catalog building's highest intersected surface in the placement column as its only available roof, excluding intermediate surfaces inside the model while retaining successive roofs for separately stacked models; **Kieran review:** lower terraces beneath higher geometry of the same model are excluded, so please check whether existing users rely on them. Exposed terraces with no higher geometry in the column remain available.
- **V11:** Keep flat-map/OSM/ground-box support and a synthetic zero-height fallback outside this change.
- **V12:** Defer the policy on using hidden buildings as support for Kieran's input.
- **V13:** Align streets by road level: managed streets use the transformed 0.15 m road offset, legacy streets and standalone segments use their origin, and other objects use their bounding-box base.

## Awkward camera angles

- **A1:** Near eye level, turn the movement square into a camera-facing horizontal strip with two directional arrowheads.
- **A2:** Near eye level, offer horizontal cursor movement only; repositioning the object in depth requires a camera tilt.
- **A3:** Trigger the near-level presentation using the elevation angle from camera to control anchor, rather than camera pitch or navigation mode; the movement/rotation pair and each landing target change presentation independently, allowing mixed states.
- **A4:** Measure the movement/rotation pair from the object's base and landing targets from their surface positions; enter near-level presentation below 10° from horizontal and leave above 12°, whether viewed from above or below.
- **A5:** Animate presentation changes over approximately 500 ms so they read as one control changing shape.
- **A6:** Make an animating subsystem inert while still swallowing its press so it cannot select objects behind; a press begun during animation requires release and a fresh press.
- **A7:** Initialize newly selected objects directly in the correct presentation, without an inert startup animation.
- **A8:** Freeze the chosen gesture regime and conversion frame at press time so movement cannot change its own interaction model midway through a drag.
- **A9:** Align the resting near-level rotation arms along the camera's horizontal view axis, rather than pointing each object's arms toward the camera; avoid compensating turns as the object translates.
- **A10:** Near eye level, use rack-and-pinion rotation: constant horizontal cursor travel produces constant yaw change, with vertical cursor movement ignored.
- **A11:** Calibrate near-level rotation at the join between the two arms, using a single rate independent of where along the control it is grabbed; accept the visible lag and differing arrowhead speeds during large near-level turns as the consequence of linear rotation in perspective.
- **A12:** Move or flip the near-level handle to clear nearby landing bars while keeping each landing bar at its true destination height; tight overhead landings require tilting the camera.
- **A13:** Resolve overlapping easy-mode controls by nearest camera hit, with shape vertex handles taking precedence where they share the movement square's plane.

## UI integration and existing features

- **U1:** Keep the existing transform modes available for operations that easy mode does not offer.
- **U2:** Collapse translate, rotate and easy into a click-operated flyout showing the active mode’s icon, closing on selection or an outside click.
- **U3:** Keep scale off the toolbar and flyout, retaining its existing keyboard route for now (for separate discussion whether scale can be retired altogether).
- **U4:** Give easy mode the `m` shortcut and correct the translate tooltip to advertise its actual `t` binding.
- **U5:** Give each gesture one initiating pointer and prevent gizmo presses or hover from changing scene selection or highlighting the underlying scene.
- **U6:** Record one undo step per completed change, with no entry for a click, a cancelled action or a gesture returning to its starting pose.
- **U7:** Make Escape restore the starting pose, end interrupted gestures safely and restore camera control, suppressing conflicting editor shortcuts during a gesture.
- **U8:** Lazy-load the controller and retain existing toolbar and shortcut behavior until easy mode is ready; gate the toolbar restructure with the same feature flag.
- **U9:** Easy mode does not apply to managed street segments (individual lanes); they retain their existing width-only controls, while easy mode moves or rotates the whole street.
- **U10:** Disallow transforms on protected/generated entities, such as generated street clones and street ground slabs, while keeping the dimmed transform-menu button clickable to choose a mode for a later eligible selection.

## Other considerations

- **O1:** Accept that touch interactions will not trigger hover emphasis, dimming or hover-driven chevron animation.
- **O2:** While a selected item is in easy mode, temporarily undo any Google 3D Tiles terrain flattening contributed by it or its children, so placement measures terrain without that item's modifications rather than a flat surface created by that same item; restore its flattening when deselected, leaving easy mode or closing the editor, without changing saved settings or other items' flattening.
- **O3:** After temporarily undoing that flattening, wait for the Google 3D Tiles terrain to finish restoring before offering tile support or landing targets; apply this wait after selection or flattening changes, not just scene load.

## Implementation

- **I1:** Apply movement before scene rendering and synchronize descendant and batched meshes with their selected parent, matching the classic gizmo’s visible behavior.

## PR management

- **P1:** Ship behind an opt-in feature flag initially, with becoming the approachable default as the eventual aim.


kfarr:
manual qa feedback for new gizmo

issue 1:
- snap back - in some cases, when dragging the rotate or yellow "start" box, after letting go of the drag button, then the dragged object returns to its original position and rotation
- See video section below:
	- [1] issue1-a-new-drag-and-drop-not-working.mov 
	- [2] issue1-b-new-drag-not-working.mov
- experienced when dragging and object from outside of a street to on top of a street and vice versa 
- scene: http://localhost:3333/#/scenes/8e975597-ce65-4b62-8e2a-53adc23007fb
- kf note: I think cascading street selection was merged while this new selection movement was in development -- is cascading selection interfering with the drop (button release) and selecting the street instead of completing the drop action for movement? on further testing i'm not sure this is related, it happens in other cases as well. we also shipped deselect, is that interfering?

issue 2:
- drag and drop of a taxi has incorrect bounding box location and also does not work upon click and drag (no movement of the dropped object)
- See video section below:
	- [3] issue2-error-drag-and-drop-error.mov
	- [4] issue2-current-drag-and-drop-no-error.mov
- scene: http://localhost:3333/#/scenes/cff0482d-a3d9-4229-a9bb-1eaf043e7558

issue 3:
- when a managed street is selected, and it is showing the new gizmo, flattening is temporarily disabled until the object is deselected
	- instead it should behalf in the same manner as the legacy translate gizmo

issue 4:
- when a user clicks and drags the new gizmo, and if the selected object's right-hand property panel is showing, the position / rotation properties do not update in realtime as the user drags / rotates as it does with the legacy gizmos

other misc issues:
* low fps -- compared to the other legacy gizmos, as a user dragging an object quickly around a scene that has geospatial on, the new gizmo is not as responsive / has a lower fps compared to the old gizmos
* sometimes there will be a yellow square around the object when it is close to the ground but not exactly on the bottom of the ground
	- if a user moves the object up higher it becomes clearer that the yellow square is meant to indicate there is distance between the object and the ground, but at low elevations it is challenging to distinguish the box outline as being a separate clickable area from the filled in box
* the yellow arrows moving ui for height adjustment is good but can be difficult to see in some contexts with bright backgrounds (bright surfaces, roadways, dirt, etc.)

style / ux issues:
- the icon is a good start, but: the icon doesn't actually match the design of the in canvas gizmo (ie the icon has 2 arrows, the actual canvas width has 1) and the icon shape inside looks like a rhombus, not a "star square" shape; the icon is also a bit smaller than the surrounding ones making it more difficult to discern
	- note: should the icon literally match the canvas gizmo icon? should the canvas gizmo be updated to include 2 arrows like the icon instead? 
- the new "move and rotate" control should be the default gizmo at the top of the gizmo action selector menu, not the bottom
- button bg styling for the new gizmo selector button does not match action bar buttons
- no indication that the new gizmo selector action bar button can be selected to choose other options, for example in photopea (photoshop clone) it uses a curved right-hand corner, perhaps to hint that is the direction the popup will appear; similarly there could be a simple arrow / curve in the upper-right hand corner of the 
- spacing not great between icon and row label in the popup menu for selector on action bar
- as a user i expect pressing the keyboard shortcut `m` will select the existing move gizmo if another mode is selected, or it will cycle through the other move gizmo options (advanced move, advanced rotate) instead of assigning separate keyboard shortcut to each of the 3 move gizmos

notes:
- unlike existing translate gizmo, the easy mode does not show a line of the "Delta" of the movement from starting to new location (the existing one also shows a subtle white axis line) -- we're assuming it's ok to not copy that, I think so
- unlike existing translate gizmo, the new gizmo no longer shows a centerpoint at the object origin -- is this ok? it's inferred from the geometry of the square, but no longer explicit; perhaps it's ok for now but would be revisited as part of https://github.com/3DStreet/3dstreet/issues/1993

videos:

[1]

https://github.com/user-attachments/assets/c756fb82-8625-429e-b806-cb9e11d3cc02

[2]

https://github.com/user-attachments/assets/ec3a009e-0210-42f3-8b78-81a7721b41ea

[3]

https://github.com/user-attachments/assets/452c46c4-76dc-4b4c-b952-2747bbc44f79

[4]

https://github.com/user-attachments/assets/e350755b-44a6-4d70-8a6f-7eb2688cd7d5

